PR #24, opened 2018-02-19
add travis-yml file

balajimanoharan:
add travis-yml file


.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+  - '6'
+
+cache:
+  directories:
+    - $HOME/.npm
+    - $HOME/packages/*/.npm
+    - $HOME/packages/*/.cache # includes bowers cache
+
+env:
+  - PACKAGE=navi-app
```
jkusa: Can we put this under `env` --> `matrix`

https://docs.travis-ci.com/user/environment-variables/#Global-Variables

balajimanoharan: Isn't it by default a `matrix` env var?

PR #29, opened 2018-02-22
App style updates

jkusa:
- Update buttons to match design language:
screen shot 2018-02-22 at 12 54 10 pm

- Nav bar style updates:

screen shot 2018-02-22 at 1 37 31 pm

### DIFF
--- a/packages/app/app/styles/common/defaults.less
+++ b/packages/app/app/styles/common/defaults.less
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2018, Yahoo Holdings Inc.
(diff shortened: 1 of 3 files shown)


packages/app/app/styles/common/defaults.less
```diff
@@ -23,8 +23,26 @@ a:link {
 //TODO make btn a component and move styles there
 .btn {
     border: none;
+    border-radius: 2px;
+    font-size: 14px;
 }
 
 .btn-primary {
     background: linear-gradient(to right,#3697F2,#3570F4);
+    min-width: 120px;
```
tacocat: Does this mess up any of the modal buttons?

jkusa: I checked the modals. We are good. 

screen shot 2018-02-23 at 1 51 04 pm

PR #57, opened 2018-03-08
Automate prerelease package publishing

jkusa:
Needs https://github.com/yahoo/navi/pull/58

- Automate prerelease package publishing
- Add `travis_wait` to longer running steps

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,38 @@
----
 language: node_js
 node_js:
-  # Using the same minimum supported node version as Ember CLI & Lerna
-  - '6'
-
-sudo: true  
-
+- '6'
+sudo: true
 addons:
-  chrome: stable  
-
+  chrome: stable
 cache:
   directories:
-    - $HOME/.npm
-    - $HOME/packages/*/.npm
-    - $HOME/packages/*/.cache # includes bowers cache
-
+  - "$HOME/.npm"
+  - "$HOME/packages/*/.npm"
+  - "$HOME/packages/*/.cache"
 env:
   matrix:
-    - PACKAGE=navi-app
-    - PACKAGE=navi-core
-    - PACKAGE=navi-dashboards
-    - PACKAGE=navi-data
-    - PACKAGE=navi-reports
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,42 +1,40 @@
...
-  - npm install -g bower
-  - bower --version
-  - npm install phantomjs-prebuilt
-  - node_modules/phantomjs-prebuilt/bin/phantomjs --version
-
+- npm config set spin false
+- npm --version
+- npm install -g bower
+- bower --version
+- npm install phantomjs-prebuilt
```
balajimanoharan: Unrelated to your PR, but anyways... since we have switched to chrome, do we still need these?

.travis.yml
```diff
@@ -1,42 +1,40 @@
...
+- node_modules/phantomjs-prebuilt/bin/phantomjs --version
 install:
-  - npm install
-
+- travis_wait npm install
 script:
-  - lerna run test --scope $PACKAGE
+- travis_wait lerna run test --scope $PACKAGE
```
balajimanoharan: Do we need the travis_wait?

scripts/ci-publish.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -e # Exit with nonzero exit code if anything fails
+
+###############################################################################
+# Config NPM
+###############################################################################
```
balajimanoharan: Do we need these banners? I don't really like them here.

.travis.yml
jkusa: https://github.com/yahoo/navi/pull/58

.travis.yml
jkusa: We do, I was seeing build timeouts without it

scripts/ci-publish.sh
jkusa: The script was initially bigger and made more sense. Still, bash scripts can get unwieldy. I like setting a standard that can carry forward if/when this scripts evolves.

PR #59, opened 2018-03-08
Update README.md

tacocat:
Adds the demo link


tacocat:
@balajimanoharan the failure is a build timeout issue that @jkusa fixed in a separate PR.

PR #67, opened 2018-04-12
Update: show close button after save schedule

Xuefeng-Zhu:
* enhancement: show close button after save schedule


balajimanoharan:
We are trying to use the eslint convention for commits. Can we update the commit message please?
Ref: https://github.com/conventional-changelog-archived-repos/conventional-changelog-eslint/blob/master/convention.md

Xuefeng-Zhu:
@balajimanoharan I think it can be done when we merge the PR?

PR #68, opened 2018-04-12
Fix: Shows error when report has invalid logical table

cythrawll:
* Would crash UI otherwise
* Likely to happen when users share reports to a table that
one user doesn't have access
* Also hides components that would crash without metadata
that user wouldn't be able to use anyway
* Allows users to select a valid table.
* Changes error message to reflect the actual problem

apr-13-2018 11-25-16

### DIFF
--- a/packages/reports/addon/consumers/request/logical-table.js
+++ b/packages/reports/addon/consumers/request/logical-table.js
@@ -22,7 +22,7 @@ export default ActionConsumer.extend({
      */
     [RequestActions.UPDATE_TABLE](route, table) {
       let currentModel = get(route, 'currentModel'),
-          oldTimeGrain = get(currentModel, 'request.logicalTable.timeGrain') || {name: ''}; // allow findBy to work when switching from an invalid table so switching to valid a table works
+          oldTimeGrain = get(currentModel, 'request.logicalTable.timeGrain') || {name: ''}; // allow findBy to work when switching from an invalid table so switching to a valid table works
 
       set(currentModel, 'request.logicalTable.table', table);
 
--- a/packages/reports/addon/consumers/request/logical-table.js
+++ b/packages/reports/addon/consumers/request/logical-table.js
@@ -22,7 +22,7 @@ export default ActionConsumer.extend({
      */
     [RequestActions.UPDATE_TABLE](route, table) {
       let currentModel = get(route, 'currentModel'),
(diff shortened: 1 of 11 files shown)


balajimanoharan:
Will choosing a table with the table-selector cause the dimensions and metrics to load? This screenshot still looks like it would crash the UI.

cythrawll:
Yes the user can recover from the error. We show the table selector here for that purpose. Choosing a valid table will cause the metrics and dimensions to load.

packages/reports/tests/unit/models/bard-request/fragments/logical-table-test.js
```diff
@@ -160,7 +160,7 @@ test('Validations', function(assert) {
         'There is one validation error');
 
       assert.equal(validations.get('messages').objectAt(0),
-        'Please select a table',
+        'Table is invalid or unavailable',
         'Table cannot be empty is a part of the error messages');
```
balajimanoharan: Can we update this test message please?

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1626,3 +1626,14 @@ test('Date Picker doesn`t change date when moving to time grain where dates are
...
 });
+
+test('Report with an unknown table doesn\'t crash', function(assert) {
+  assert.expect(1);
+  visit('/reports/9');
+
+  andThen(() => {
+    assert.equal(find('.navi-info-message__error-list-item').text().trim(),
+      'Table is invalid or unavailable',
+      'Should show aan error message when table cannot be found in metadata');
```
balajimanoharan: s/aan/an

packages/reports/addon/consumers/request/logical-table.js
```diff
@@ -22,7 +22,7 @@ export default ActionConsumer.extend({
      */
     [RequestActions.UPDATE_TABLE](route, table) {
       let currentModel = get(route, 'currentModel'),
-          oldTimeGrain = get(currentModel, 'request.logicalTable.timeGrain');
+          oldTimeGrain = get(currentModel, 'request.logicalTable.timeGrain') || {name: ''}; // allow findBy to work when switching from an invalid table so switching to valid a table works
```
balajimanoharan: One more string update
`to valid a table` => `to a valid table`

Issue #71, opened 2018-04-13
Refactor metrics into objects in visualization models' metadata


mattgraham1995:
- [x] Line Chart
- [x] Pie Chart
- [x] Table
- [x] Metric Label
- [x] Goal Gauge 
- [x] Tooltips (Not a model, but rather utilize the new metric objects to simplify tooltips)

PR #73, opened 2018-04-16
Refactor table visualization metric 

mattgraham1995:
Refers to #71 

The field property of each column is now always an object.

### DIFF
--- a/packages/visualizations/addon/models/table.js
+++ b/packages/visualizations/addon/models/table.js
@@ -5,6 +5,7 @@
 import Ember from 'ember';
 import DS from 'ember-data';
 import VisualizationBase from './visualization';
+import { canonicalizeMetric } from 'navi-data/utils/metric';
 import { validator, buildValidations } from 'ember-cp-validations';
 import { metricFormat } from 'navi-data/helpers/metric-format';
 
@@ -53,14 +54,14 @@ export default VisualizationBase.extend(Validations, {
         metrics = get(request, 'metrics') || [];
 
     let dateColumn = {
-      field: 'dateTime',
+      field: {dateTime: 'dateTime'},
       type: 'dateTime',
       displayName: 'Date'
     };
 
     let dimensionColumns = dimensions.map(dimension => {
       return {
-        field: get(dimension, 'dimension.name'),
+        field: {dimension: get(dimension, 'dimension.name')},
         type: 'dimension',
         displayName: get(dimension, 'dimension.longName')
       };
@@ -74,7 +75,7 @@ export default VisualizationBase.extend(Validations, {
           longName = get(metric, 'metric.longName');
 
       return {
-        field: get(metric, 'canonicalName'),
+        field: metric.toJSON(),
         type,
         displayName: metricFormat(metric, longName)
       };
@@ -103,7 +104,7 @@ export default VisualizationBase.extend(Validations, {
  */
 function hasAllColumns(request, columns) {
   //retrieve everything but dateTime from metadata.columns
-  let columnFields  = arr(arr(columns).rejectBy('field', 'dateTime')).mapBy('field'),
+  let columnFields  = arr(arr(columns).rejectBy('type','dateTime')).map(column => get(column, 'field.dimension') || canonicalizeMetric(get(column, 'field'))),
       dimensions    = arr(get(request, 'dimensions')).mapBy('dimension.name'),
       metrics       = arr(get(request, 'metrics')).mapBy('canonicalName'),
       requestFields = [ ...dimensions, ...metrics ];
--- a/packages/visualizations/addon/serializers/table.js
+++ b/packages/visualizations/addon/serializers/table.js
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2018, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import VisualizationSerializer from 'navi-visualizations/serializers/visualization';
+import { parseMetricName } from 'navi-data/utils/metric';
+import { get } from '@ember/object';
+
+const TRANSFORMS = {
+  metric:    (column) => Object.assign(column, { field: parseMetricName(column.field) }),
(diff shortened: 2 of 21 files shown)


packages/visualizations/addon/models/table.js
```diff
@@ -105,9 +106,9 @@ function hasAllColumns(request, columns) {
   //retrieve everything but dateTime from metadata.columns
   let columnFields  = arr(arr(columns).rejectBy('field', 'dateTime')).mapBy('field'),
       dimensions    = arr(get(request, 'dimensions')).mapBy('dimension.name'),
-      metrics       = arr(get(request, 'metrics')).mapBy('canonicalName'),
+      metrics       = get(request, 'metrics').map(metric => metric.toJSON()),
       requestFields = [ ...dimensions, ...metrics ];
 
   return requestFields.length === columnFields.length &&
-    requestFields.every(field => columnFields.includes(field));
+    requestFields.every(field => !!columnFields.find(column => isEqual(column, field)));
```
balajimanoharan: Can we compare by `canonicalName` as we did with line charts?

packages/visualizations/tests/unit/serializers/table-test.js
```diff
@@ -0,0 +1,94 @@
...
+});
+
+test('normalize', function(assert) {
+  assert.expect(2);
+
+  assert.deepEqual(Serializer.normalize(),
+    {data: null},
+    'null is returned for an undefined response');
+
+  assert.deepEqual(Serializer.normalize(Model, {
```
balajimanoharan: Can we extract the huge objects into vars please?

packages/visualizations/addon/models/table.js
mattgraham1995: I looked into this, but this case is a little different since we're comparing columns that aren't always metrics.

balajimanoharan: Wait a sec, that means the field property is an object when its a metric and a string when its a dimension??
That would be very confusing.

balajimanoharan: Wait a sec, that means the field property is an object when its a metric and a string when its a dimension??
That would be very confusing.

packages/visualizations/addon/serializers/table.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright 2018, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import Ember from 'ember';
```
balajimanoharan: Can we use the new import syntax for `get` and `isPresent`?

packages/visualizations/tests/unit/serializers/table-test.js
```diff
@@ -0,0 +1,97 @@
+import Ember from 'ember';
+import { moduleFor, test } from 'ember-qunit';
+import { setupMock, teardownMock } from '../../helpers/mirage-helper';
+
+const { getOwner } = Ember;
```
balajimanoharan: new import syntax

packages/visualizations/addon/models/table.js
mattgraham1995: I could make dimensions and datetime columns store objects as well, but each object would have a different form so checking for the contents of the field object would still require a type check of the column. Maybe there should be a common `field` object form that all column types use

packages/visualizations/addon/serializers/table.js
```diff
@@ -0,0 +1,46 @@
...
+   * @param type - class type as a DS model
+   * @param visualization - json parsed object
+   * @return {Object} normalized payload
+   */
+  normalize(type, visualization) {
+    if(visualization) {
+      let columns = get(visualization, 'metadata.columns');
+
+      if(isPresent(columns)) {
+        columns.map(column => {
```
jkusa: This is more of a `forEach` than a `map`. Use `map` to return a new array and then assign it to `metadata.columns`

Instead of a switch can we use an object:
```js
const TRANSFORMS = {
   metric:    (column) => column,
   threshold: (column) => assign(column, { field: parseMetricName(column.field) }),
   dimension: (column) => assign(column, { field: { dimension: column.field } })
   dateTime:  (column) => assign(column, { field: { dateTime: column.field } })
};


...


normalize(type, visualization) {
  if(visualization) {
    let columns = get(visualization, 'metadata.columns').map(column => {
      if(typeof get(column, 'field') === 'string') {
        let type = get(column, 'type');
        return TRANSFORMS[type](column);
      }
      return column;
    });
    assign(visualization.metadata, { columns });
  }
  return this._super(type, visualization);
}

```

packages/visualizations/addon/templates/components/navi-visualizations/table.hbs
```diff
@@ -10,7 +10,7 @@
 
           {{#sortable-item
             class='table-header-cell'
-            classNameBindings='column.type column.field'
+            classNameBindings='column.type column.flatField'
```
balajimanoharan: What is `flatField` actually used for?

mattgraham1995:
@jkusa Comments addressed

packages/visualizations/tests/unit/serializers/table-test.js
```diff
@@ -0,0 +1,105 @@
...
+                  type: 'metric',
+                  displayName: 'Page Views'
+                },
+                {
+                  field: {metric: 'revenue', parameters: {currency: 'USD'}},
+                  type: 'threshold',
+                  displayName: 'Revenue (USD)'
+                },
+                {
+                  field: {dimension: 'age'},
```
jkusa: Do we have a test for dimension as a string?

packages/visualizations/tests/unit/serializers/table-test.js
```diff
@@ -0,0 +1,105 @@
...
+                  type: 'dimension',
+                  displayName: 'Age'
+                },
+                {
+                  field: {dateTime: 'dateTime'},
```
jkusa: Do we have a test for dateTime as a string?

mattgraham1995: We do now

PR #88, opened 2018-05-02
Update: Add information txt to the edit side

Xuefeng-Zhu:


packages/visualizations/addon/templates/components/visualization-config/table.hbs
```diff
@@ -51,6 +51,6 @@
   </div>
 {{else}}
   <div class='visualization-config__warn-msg'>
-    No configuration options available.
+    Please use the column headers to config the table
```
cythrawll: can we say `configure` here?
I also not sure about using the please here.

can we say something like: 'You can access more configuration in each of the column headers'

Xuefeng-Zhu: updated

PR #100, opened 2018-05-30
Fix: Dimensions loaded from server now show as selected in dropdown

cythrawll:
* Added an  isEqual function to selectedDimensions so
object equality is not needed.

### DIFF
--- a/packages/data/addon/mirage/routes/bard-lite.js
+++ b/packages/data/addon/mirage/routes/bard-lite.js
@@ -181,11 +181,13 @@ export default function(
 
     // Handle value filters
     if ('filters' in request.queryParams) {
-      // Assume filters to dimension endpoint are in the form "dimension|id-in[...]"
-      let [/* full match */, idString] = request.queryParams.filters.match(/\[(.*)\]/),
-          ids = idString.split(',');
+      let [/* full match */, queryString] = request.queryParams.filters.match(/\[(.*)\]/),
+          values = queryString.split(','),
+          isIdSearch = /\|id/.test(request.queryParams.filters);
 
-      rows = rows.filter(row => ids.includes(row.id));
+      rows = isIdSearch ? 
+        rows.filter(row => values.includes(row.id)) :
+        rows.filter(row => values.some(value => row.description.toLowerCase().includes(value.toLowerCase())));
     }
 
     return { rows };
--- a/packages/data/addon/models/bard-dimension.js
+++ b/packages/data/addon/models/bard-dimension.js
@@ -6,7 +6,11 @@
  */
 import Ember from 'ember';
 
-export default Ember.Object.extend().reopenClass({
+export default Ember.Object.extend({
+  isEqual(other) {
+    return this.id === other.id;
+  }
+}).reopenClass({
 
   /**
    * @property {String} identifierField - name of the dimension identifier field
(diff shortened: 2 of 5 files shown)


packages/reports/addon/components/filter-values/dimension-select.js
```diff
@@ -49,9 +49,13 @@ export default Ember.Component.extend({
         dimensionName = get(this, 'dimensionName'),
         dimensionService = get(this, '_dimensionService');
 
-        // Only fetch dimensions if filter has values
+    // Only fetch dimensions if filter has values
     if (get(dimensionIds, 'length')) {
-      return dimensionService.find(dimensionName, {field: 'id', values: dimensionIds.join(',')});
+      return dimensionService.find(dimensionName, {field: 'id', values: dimensionIds.join(',')}).then(dims => dims.map(dimension => {
+        //adding isEqual to object so power select can find selected values accurately.
+        dimension.isEqual = function(other) { return this.id == other.id; };
```
Xuefeng-Zhu: why not arrow function?

balajimanoharan: `(other) => this.id === other.id`

balajimanoharan: should we add a test for the problem this fixes?

Xuefeng-Zhu: I think arrow function will not work here due to `this` scope

cythrawll: correct arrow function is not appropriate here.

balajimanoharan: ok.. we want `this` to be in the dimension's scope.

jkusa: Can we move `isEqual` to the dimension model:

https://github.com/yahoo/navi/blob/master/packages/data/addon/models/bard-dimension.js#L9

cythrawll:
added tests, but I still need to try adding isEqual to the dimension model.

cythrawll:
Moving the isEqual to the model worked

packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -181,11 +181,13 @@ export default function(
 
     // Handle value filters
     if ('filters' in request.queryParams) {
-      // Assume filters to dimension endpoint are in the form "dimension|id-in[...]"
       let [/* full match */, idString] = request.queryParams.filters.match(/\[(.*)\]/),
```
balajimanoharan: Since it may not be ids, can we call this `queryString`?

packages/data/addon/models/bard-dimension.js
```diff
@@ -6,7 +6,11 @@
  */
 import Ember from 'ember';
 
-export default Ember.Object.extend().reopenClass({
+export default Ember.Object.extend({
+  isEqual: function(other) {
```
balajimanoharan: now we can use the arrow function right?

cythrawll: Nah, `this` isn't available here: gave it a shot and it broke everything.

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1637,3 +1637,35 @@ test('Report with an unknown table doesn\'t crash', function(assert) {
...
+  let options, option, 
+      dropdownSelector = '.filter-values--dimension-select';
+  visit('/reports/new');
+
+  andThen(() => {
+    selectChoose('.navi-table-select__dropdown', 'Table A');
+    click('.grouped-list__item:Contains(EventId) .checkbox-selector__filter');
+  });
+
+  andThen(() => {
```
jkusa: Can we comment these steps?

PR #102, opened 2018-05-31
New: Format table column based on user config

Xuefeng-Zhu:


Xuefeng-Zhu:
@balajimanoharan I think it will be nice to have a hbs helper, but what is helper supposed to do and where should we use it?

balajimanoharan:
I would suggest changing [this](https://github.com/yahoo/navi/blob/master/packages/visualizations/addon/templates/components/cell-renderers/metric.hbs#L2) file to 
```hbs
{{format-value metricValue format}}
```

and you could use the same thing for the threshold cell renderer

Xuefeng-Zhu:
@balajimanoharan I have refactored the code with the helper

packages/visualizations/addon/helpers/format-table-cell.js
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright 2018, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+
+import { helper } from '@ember/component/helper';
+import { isEmpty } from '@ember/utils';
+import numeral from 'numeral';
+import { smartFormatNumber } from 'navi-core/helpers/smart-format-number';
+
+export function formatTableCell([value, format, smartFormat]) {
```
balajimanoharan: Can we call the boolean argument `shouldSmartFormat`?
And can we add jsdoc for this function?

packages/visualizations/tests/integration/helpers/format-table-cell-test.js
```diff
@@ -0,0 +1,34 @@
+
```
balajimanoharan: Can we clean this up? extra `\n`

packages/visualizations/addon/helpers/format-table-cell.js
```diff
@@ -0,0 +1,32 @@
...
+export function formatTableCell([value, format, shouldSmartFormat]) {
+  if (isEmpty(value)) {
+    return '--';
+  }
+
+  if (format) {
+    return numeral(value).format(format);
+  }
+
+  return shouldSmartFormat ? smartFormatNumber([value]) : value;
```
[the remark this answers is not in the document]
Xuefeng-Zhu: for threshold cell, we do not want smart format number

Xuefeng-Zhu:
@jkusa I have updated the code based on discussion

packages/visualizations/addon/templates/components/cell-renderers/threshold.hbs
```diff
@@ -1,2 +1,2 @@
-{{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
-{{metricValue}}
+{{!-- Copyright 2018, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{format-table-cell metricValue format}}
```
balajimanoharan: Now, would it be better to extend the metric cell renderer for this? 
@jkusa what do you think?

jkusa: @balajimanoharan maybe, can `threshold` extend `metric`?

balajimanoharan: it probably could. there only one extra property in threshold, `valueIndicator`. Everything else is the same.

Xuefeng-Zhu: @jkusa updated

jkusa: Do we need the helper now?

Xuefeng-Zhu: @jkusa Do you refer to the template file?

packages/visualizations/addon/templates/components/cell-renderers/metric.hbs
```diff
@@ -1,2 +1,2 @@
-{{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
-{{metricValue}}
+{{!-- Copyright 2018, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{format-table-cell metricValue format}}
```
jkusa: Do we need this helper for a single use? Can we bring this into the metric cell-renderer

Xuefeng-Zhu: yeah. It seems that it is not needed anymore

Xuefeng-Zhu: @jkusa removed

PR #107, opened 2018-06-04
Update: Table modal to make user config persistent

Xuefeng-Zhu:
make user config persistent

### DIFF
--- a/packages/visualizations/addon/models/table.js
+++ b/packages/visualizations/addon/models/table.js
@@ -8,6 +8,7 @@ import VisualizationBase from './visualization';
 import { canonicalizeMetric } from 'navi-data/utils/metric';
 import { validator, buildValidations } from 'ember-cp-validations';
 import { metricFormat } from 'navi-data/helpers/metric-format';
+import keyBy from 'lodash/keyBy';
 
 const {
   A:arr,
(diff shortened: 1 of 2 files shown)


packages/visualizations/addon/models/table.js
```diff
@@ -51,7 +52,15 @@ export default VisualizationBase.extend(Validations, {
    */
   rebuildConfig(request /*, response */) {
     let dimensions = get(request, 'dimensions') || [],
-        metrics = get(request, 'metrics') || [];
+        metrics = get(request, 'metrics') || [],
+        columns = get(this, 'metadata.columns'),
+        columnIndex = keyBy(columns, ({ field, type }) => {
```
balajimanoharan: Can we add some comments here explaining that the key is the dimensionId or metricId or dateTime?

packages/visualizations/addon/models/table.js
```diff
@@ -72,13 +83,17 @@ export default VisualizationBase.extend(Validations, {
       let category = get(metric, 'metric.category'),
           isTrend = ~(category.toLowerCase().indexOf('trend')),
           type = isTrend ? 'threshold' : 'metric',
+          field = metric.toJSON(),
+          column = columnIndex[field.metric],
           longName = get(metric, 'metric.longName'),
-          displayName = metricFormat(metric, longName);
+          displayName = column ? column.displayName : metricFormat(metric, longName),
+          format = column ? column.format : '';
```
balajimanoharan: Can format not be undefined?

Xuefeng-Zhu: If format is set, it will at least be en empty string

PR #108, opened 2018-06-04
Moving report and dashboard mocks to core

balajimanoharan:
Moving report and dashboard mocks to core


Xuefeng-Zhu:
The build is failed, why it is merged?

PR #112, opened 2018-06-11
Fix: Metric table cell default format

Xuefeng-Zhu:
image

### DIFF
--- a/packages/visualizations/addon/components/cell-renderers/metric.js
+++ b/packages/visualizations/addon/components/cell-renderers/metric.js
@@ -12,10 +12,11 @@
 
 import Ember from 'ember';
 import layout from '../../templates/components/cell-renderers/metric';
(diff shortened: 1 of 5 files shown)


packages/visualizations/addon/components/cell-renderers/metric.js
```diff
@@ -52,6 +54,6 @@ export default Ember.Component.extend({
       return numeral(metricValue).format(format);
     }
 
-    return metricValue;
+    return type === 'metric' ? smartFormatNumber([metricValue]) : metricValue;
```
balajimanoharan: Do we need the `type` check?
I think it doesn't matter if the threshold is smart formatted. Can we verify pls?

Xuefeng-Zhu: @balajimanoharan threshold will look weird with smartformat

PR #116, opened 2018-06-12
Add new directory package

balajimanoharan:
Add new directory package
- this package will serve as a repository for the index route

This PR contains:
- addon initialization 
- mirage config for directory


packages/directory/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "start": "ember serve",
+    "test": "export COVERAGE=true; ember test --test-port 0"
+  },
+  "dependencies": {
+    "ember-cli-babel": "^6.6.0"
+  },
+  "devDependencies": {
+    "broccoli-asset-rev": "^2.4.5",
+    "ember-ajax": "^3.0.0",
+    "ember-cli": "~2.18.1",
```
Xuefeng-Zhu: Do we want to use the latest ember?

balajimanoharan: We do, but we still want it to be compatible with the other repositories. We don't know when the other upgrades would happen.
I will keep it at 2.18 for now, and then do the upgrade later with the others.

packages/directory/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "ember-cli-htmlbars-inline-precompile": "^1.0.0",
+    "ember-cli-inject-live-reload": "^1.4.1",
+    "ember-cli-mirage": "^0.4.7",
+    "ember-cli-moment-shim": "^3.7.1",
+    "ember-cli-qunit": "^4.1.1",
+    "ember-cli-shims": "^1.2.0",
+    "ember-cli-sri": "^2.1.0",
+    "ember-cli-uglify": "^2.0.0",
+    "ember-cp-validations": "^4.0.0-beta.1",
+    "ember-data": "^3.1.1",
```
jkusa: Do we want 3.x?

PR #118, opened 2018-06-15
New: Add prettier setup

Xuefeng-Zhu:


packages/data/addon/services/metric-name.js
```diff
@@ -19,7 +19,12 @@ export default Service.extend({
    * @returns {String} - long name for the metric from the metadata
    */
   getLongName(metricId) {
-    return get(this, 'metricMeta').getMetaField('metric', metricId, 'longName', metricId);
+    return get(this, 'metricMeta').getMetaField(
+      'metric',
+      metricId,
+      ' longName',
```
mattgraham1995: `' longName'` -> `'longName'`

packages/data/addon/services/metric-name.js
```diff
@@ -29,7 +34,7 @@ export default Service.extend({
    */
   getDisplayName(metricObject) {
     let metricId = get(metricObject, 'metric'),
-        longName = this.getLongName(metricId);
+      longName = this.getLongName(metricId);
```
mattgraham1995: Won't this spacing conflict with our eslint rule?

Xuefeng-Zhu: yeah, we need to update eslint config

Xuefeng-Zhu:
ready to merge

packages/data/addon/services/metric-name.js
balajimanoharan: I liked how it was before. Can we not configure prettier to behave the same way eslint does currently?

Xuefeng-Zhu: Unfortunately, prettier is not very configurable.

balajimanoharan: Lets discuss this during the navi meeting tomorrow.

Xuefeng-Zhu:
@balajimanoharan I think last time when we discuss this, @jkusa does not want to include `eslint-config-prettier`, since it can cause local test failure

balajimanoharan:
> @balajimanoharan I think last time when we discuss this, @jkusa does not want to include eslint-config-prettier, since it can cause local test failure

As discussed in standup, we do want this.

Xuefeng-Zhu:
@jkusa @balajimanoharan I have added eslint-config-prettier

packages/app/package.json
```diff
@@ -78,6 +78,7 @@
     "ember-tooltips": "^2.9.2",
     "ember-truth-helpers": "^2.0.0",
     "ember-uuid": "1.0.0",
+    "eslint-config-prettier": "^3.0.1",
```
balajimanoharan: Does this need to be in every addon? Or could we just add it to the global package.json

balajimanoharan:
What is the status of this?

packages/app/package.json
Xuefeng-Zhu: If we create a root level eslint config file, I think we just need to add it to the global package.json.

balajimanoharan:
:shipit:

PR #122, opened 2018-06-18
Maintain column order when adding new columns to table

cythrawll:


cythrawll:
any suggestions for that? the existing order tests wouldn't work for this case. wondering if it's really worth adding the testing infrastructure needed to test this thing.

balajimanoharan:
We have a test in visualizations that does something similar.
https://github.com/yahoo/navi/blob/master/packages/visualizations/tests/acceptance/table-test.js#L23

balajimanoharan:
https://github.com/jgwhite/ember-sortable/blob/1.x/test-support/helpers/ember-sortable/test-helpers.js#L2

cythrawll:
tests added and found another bug due to metric change of shape

packages/visualizations/addon/models/table.js
```diff
@@ -92,12 +92,25 @@ export default VisualizationBase.extend(Validations, {
...
+    ];
+
+    if(columns.length) {
+      newColumns.sort((a,b) => {
+        let foundA = columns.findIndex(item => isSameColumn(item, a)),
+            foundB = columns.findIndex(item => isSameColumn(item, b));
+        if(foundA === -1 || foundB === -1) {
+          return 0;
+        }
+        return foundA - foundB;
```
mattgraham1995: Could we do this?
`return foundA > -1 && foundB > -1 ? foundA - foundB : 0;`

packages/visualizations/addon/models/table.js
```diff
@@ -92,12 +92,25 @@ export default VisualizationBase.extend(Validations, {
       };
     });
 
+    let newColumns = [
```
balajimanoharan: I feel `rebuildConfig` here has grown to be a bit too big. Can we split this into multiple smaller methods?
`getMetricColumns`, `getDimensionColumns`...

packages/visualizations/addon/models/table.js
```diff
@@ -124,6 +137,21 @@ function hasAllColumns(request, columns) {
...
+ * @function isSameColumn
+ * @param {Object} colA - column object
+ * @param {Object} colB - column object
+ * @returns {Boolean} - whether the columns are the same field
+ */
+function isSameColumn(colA, colB) {
+  if(colA.type === colB.type) {
+    return colA.type === 'dimension' ? 
+      colA.field.dimension == colB.field.dimension : 
+      canonicalizeMetric(colA.field) == canonicalizeMetric(colB.field);
```
balajimanoharan: Also I feel like this logic is done in multiple places. Could we make use of displayName here instead? `colA.displayName === colB.displayName`

cythrawll:
refactored rebuild component, PR comments addressed

PR #126, opened 2018-06-21
Update: Add margin to bottom of dashboard

mattgraham1995:
Before: 
image

After: 
image

Adds some space between the bottom widget and the bottom of the dashboard body. 
Also updated ember-cli-dependency-checker because the app would not run locally for me otherwise.

### DIFF
--- a/packages/dashboards/app/styles/navi-dashboards/components/navi-dashboard.less
+++ b/packages/dashboards/app/styles/navi-dashboards/components/navi-dashboard.less
@@ -20,6 +20,10 @@
     margin: 3px 10px;
   }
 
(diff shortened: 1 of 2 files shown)


packages/dashboards/app/styles/navi-dashboards/components/navi-dashboard.less
```diff
@@ -20,6 +20,10 @@
     margin: 3px 10px;
   }
 
+  &__widgets {
+    margin-bottom: 10px;
```
balajimanoharan: I think 20px would look a little better.

PR #128, opened 2018-06-22
[Ember Upgrade] Navi Data

balajimanoharan:
Ember Upgrade navi-data to 3.1.0
Major code changes are in Upgrade to 2.18 commit - Request Builder Updates


packages/data/addon/builder/request.js
```diff
@@ -57,7 +70,24 @@ export default Ember.Object.extend({
...
+  /**
+   * Sets a request property in an immutable way
+   *
+   * @method push
+   * @private
+   * @param {String} property - key of array property to push to
+   * @param {Array|Object} value - new value
+   * @returns {Object} copy of request with the new value added
+   */
+  _set(property, value) {
```
jkusa: Could this be `_immutableSet`?

packages/data/addon/builder/request.js
```diff
@@ -57,7 +70,24 @@ export default Ember.Object.extend({
...
+    let requestCopy = this.copy();
```
jkusa: How about something like:
```js
return this.copy({
  [property]: value
});

packages/data/addon/builder/request.js
balajimanoharan: yes, I can change the name

PR #130, opened 2018-06-25
Fix: Metric format on serialized metric mode

Xuefeng-Zhu:
The default name of metric column is not displayed correctly for metric with parameters

### DIFF
--- a/packages/data/tests/unit/helpers/metric-format-test.js
+++ b/packages/data/tests/unit/helpers/metric-format-test.js
@@ -4,26 +4,26 @@ import { metricFormat } from 'navi-data/helpers/metric-format';
 
 module('Unit | Helper | metric formatter', function(hooks) {
   setupTest(hooks);
+});
 
-  test('It formats metrics appropriately', function(assert) {
-    assert.expect(6);
-    assert.equal(metricFormat({metric: 'revenue'}, 'Revenue'),
-      'Revenue',
-      'Formats metrics given longname');
-    assert.equal(metricFormat(null, 'bar'),
-      'bar',
-      'Returns long name in case metric is not available');
-    assert.equal(metricFormat({metric: 'foo'}),
-      '--',
-      'Renders -- if longname is not given');
-    assert.equal(metricFormat({metric: {name: 'foo', longName: 'Foo'}, parameters: {p: 1}}, 'Bar'),
-      'Bar (1)',
-      'Formats metric with parameters');
-    assert.equal(metricFormat({metric: {name: 'foo', longName: 'Foo'}, parameters: {p: 1, m: 3, as: 'ham'}}, 'Bar'),
-      'Bar (1,3)',
-      'Formats multiple params and ignores the `as` parameter');
-    assert.equal(metricFormat({metric: {name: 'foo', longName: 'Foo'}, parameters: {p: 1, m: 3, as: 'ham'}}),
-      '-- (1,3)',
-      'Formats multiple params and ignores the `as` parameter when default longname is allowed');
-  });
+test('It formats metrics appropriately', function(assert) {
+  assert.expect(6);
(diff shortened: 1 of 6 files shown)


Xuefeng-Zhu:
@mattgraham1995 updated

packages/data/addon/helpers/metric-format.js
```diff
@@ -13,6 +13,7 @@ import { hasParameters } from '../utils/metric';
 import { get } from '@ember/object';
 import { isPresent } from '@ember/utils';
 import Helper from '@ember/component/helper';
+import isObject from 'lodash/isObject';
```
mattgraham1995: Do we need this or can we just use `typeof val === 'object'`

cythrawll: `typeof val === 'object'` is pretty nasty if you get nulls and other objects that don't work like the plain ol js objects.

mattgraham1995: or maybe `val === Object(val)` ?

Xuefeng-Zhu: Still prefer _.isObject, since it is easier to read and understand

Xuefeng-Zhu: @mattgraham1995 Actually need to use `isPlainObject` to exclude array case, so `val === Object(val)  ` will not work here

packages/data/tests/unit/helpers/metric-format-test.js
```diff
@@ -17,6 +17,9 @@ test('It formats metrics appropriately', function(assert) {
   assert.equal(metricFormat({metric: {name: 'foo', longName: 'Foo'}, parameters: {p: 1}}, 'Bar'),
     'Bar (1)',
     'Formats metric with parameters');
+  assert.equal(metricFormat({ metric: { name: 'foo', longName: 'Foo' }, parameters: { p: { defaultValue:1 } } }, 'Bar'),
+    'Bar (1)',
+    'Formats metric with parameters');
```
jkusa: @Xuefeng-Zhu Can we update this msg to be more specific than the msg above?

Xuefeng-Zhu: updated

packages/data/tests/unit/helpers/metric-format-test.js
```diff
@@ -17,6 +17,9 @@ test('It formats metrics appropriately', function(assert) {
   assert.equal(metricFormat({metric: {name: 'foo', longName: 'Foo'}, parameters: {p: 1}}, 'Bar'),
     'Bar (1)',
     'Formats metric with parameters');
+  assert.equal(metricFormat({ metric: { name: 'foo', longName: 'Foo' }, parameters: { p: { defaultValue:1 } } }, 'Bar'),
```
jkusa: Are we sure about this object shape?
```js
{ metric: { name: 'foo', longName: 'Foo' }, parameters: { p: { defaultValue:1 } } }
```

Also, can we verify the fix with an acceptance test?

cythrawll: So this shape does look correct.  However I still need to figure out how to produce this for an acceptance test.

cythrawll:
acceptance test added!

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1709,4 +1709,16 @@ test("adding metrics to reordered table keeps order", function(assert) {
       'Total Clicks'
     ], 'The headers are reordered as specified by the reorder');
   });
+});
+
+test('Paramaterized metrics with default displayname are not considered custom', function(assert) {
```
mattgraham1995: Parameterized

PR #131, opened 2018-06-26
Fix Schedule Modal Bugs

balajimanoharan:
- Listen to format changes in the delivery rule
- Apply secondary styles to delete button
- Proper Pill and input styling for schedule modal

screen shot 2018-06-25 at 10 50 27 pm

### DIFF
--- a/packages/dashboards/app/styles/navi-dashboards/common.less
+++ b/packages/dashboards/app/styles/navi-dashboards/common.less
@@ -43,6 +43,7 @@
     margin-right: 10px;
 
     .action {
+      display: flex;
       background: transparent;
       color: @body-font-gray;
       font-size: @font-size-base-large;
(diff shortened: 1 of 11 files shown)


jkusa:
@balajimanoharan how does this look with the styles in Navi App?

packages/dashboards/app/styles/navi-dashboards/common.less
```diff
@@ -43,16 +43,29 @@
...
+      }
+
       .icon {
         font-size: @font-size-base-large;
         position: relative;
         top: -1px;
       }
 
+      &.delete button {
+        font-weight: 400;
```
Xuefeng-Zhu: maybe use `btn-secondary` instead

balajimanoharan: No, the problem here is that `btn` and `btn-secondary` are applied to different elements and thats causing an issue. These were just overrides so that this button looks the same as the other components without having to change all of the styles.

balajimanoharan:
@jkusa checked the app, the buttons looks ok.

PR #141, opened 2018-07-20
Update: Cleanup all jquery extends

balajimanoharan:
Update: Cleanup all jquery extends

### DIFF
--- a/packages/reports/addon/routes/reports/report.js
+++ b/packages/reports/addon/routes/reports/report.js
@@ -3,6 +3,7 @@
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 import Ember from 'ember';
+import merge from 'lodash/merge';
 
 const { computed, get, set } = Ember;
 
@@ -200,7 +201,7 @@ export default Ember.Route.extend({
       let metadata = get(this, 'currentModel.visualization.metadata');
 
       set(this.currentModel, 'visualization.metadata',
-        Ember.$.extend(true, {}, metadata, metadataUpdates)
+        merge({}, metadata, metadataUpdates)
       );
     },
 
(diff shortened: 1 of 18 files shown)


cythrawll:
Do we need to do ember polyfill explicitly? does babel polyfills already include `Object.extend`?

balajimanoharan:
> Do we need to do ember polyfill explicitly? does babel polyfills already include Object.extend?

Do you mean this? 
```
import { assign } from '@ember/polyfills';
```
That was just the new way to import the assign thats already a part of `Ember`.
https://emberjs.com/api/ember/3.2/functions/@ember%2Fpolyfills/assign
I didn't have to bring in a polyfill.

packages/reports/addon/routes/reports/report.js
```diff
@@ -200,7 +201,7 @@ export default Ember.Route.extend({
       let metadata = get(this, 'currentModel.visualization.metadata');
 
       set(this.currentModel, 'visualization.metadata',
-        Ember.$.extend(true, {}, metadata, metadataUpdates)
+        merge({}, metadata, metadataUpdates)
```
Xuefeng-Zhu: why not use merge from Ember?

cythrawll:
> Do you mean this?

> `import { assign } from '@ember/polyfills';`
> That was just the new way to import the assign thats already a part of Ember.
> https://emberjs.com/api/ember/3.2/functions/@ember%2Fpolyfills/assign

Yeah how is that different than the native `Object.assign`? Maybe it works with ember objects?

> I didn't have to bring in a polyfill.

I suspect the polyfill is already there.

packages/reports/addon/routes/reports/report.js
balajimanoharan: `Ember.merge` works only for 2 objects

balajimanoharan:
> Yeah how is that different than the native Object.assign? Maybe it works with ember objects?

Its the same thing. Looks like some guy called @jkusa did this.. https://github.com/emberjs/ember.js/blob/c1f4d5b80279cba9424b83505dc28d65cada7572/packages/%40ember/polyfills/lib/assign.ts#L52

PR #142, opened 2018-07-23
Fix: Export does not open up useless tab

cythrawll:
We don't need target='_blank' here. downloads work without it, and it's smoother and less confusing if it doesn't open up a blank tab that's going to close.


cythrawll:
I tested on firefox and chrome, did not see any issues.

mattgraham1995:
Is this ready to merge?

PR #144, opened 2018-07-24
Integrate directory in app

balajimanoharan:
- Fix Mirage to use the right user
- Integrate Directory in App
- Restyle the landing page

screen shot 2018-07-24 at 1 37 08 pm


cythrawll:
update copyright otherwise looks fine.

PR #160, opened 2018-08-08
Update: Dimension filter uses primaryKey tag

cythrawll:
Dimensions don't always use `id` as their primary key.  We have tags set up to specify which dimension field is the primary key, and therefore we can build the correct dimension filter that gives the intended results.

### DIFF
--- a/packages/data/addon/mirage/routes/bard-lite.js
+++ b/packages/data/addon/mirage/routes/bard-lite.js
@@ -72,10 +72,15 @@ function _fakeDimensionValues(name, count) {
   let fakeValues = [];
 
   for (let i = 0; i < count; i++) {
-    fakeValues.push({
+    let key = null;
+    // used to generate alternative primary keys for dimensions that don't use `id` as their primaryKey (in this case uses `key` instead)
+    if(name === 'multiSystemId') {
+      key = `k${i + 1}`;
+    }
+    fakeValues.push(Object.assign({
       id: `${i + 1}`,
-      description: faker.commerce.productName()
-    });
+      description: faker.commerce.productName(),
+    }, key ? {key} : null));
   }
 
   return fakeValues;
@@ -183,10 +188,10 @@ export default function(
     if ('filters' in request.queryParams) {
       let [/* full match */, queryString] = request.queryParams.filters.match(/\[(.*)\]/),
           values = queryString.split(','),
-          isIdSearch = /\|id/.test(request.queryParams.filters);
+          fieldMatch = request.queryParams.filters.match(/\|(id|key)/);
 
-      rows = isIdSearch ? 
-        rows.filter(row => values.includes(row.id)) :
+      rows = fieldMatch && fieldMatch.length > 0 ? 
+        rows.filter(row => values.includes(row[fieldMatch[1]])) :
         rows.filter(row => values.some(value => row.description.toLowerCase().includes(value.toLowerCase())));
     }
 
--- a/packages/reports/addon/consumers/request/filter.js
+++ b/packages/reports/addon/consumers/request/filter.js
@@ -10,13 +10,7 @@ import { canonicalizeMetric } from 'navi-data/utils/metric';
 import { getSelectedMetricsOfBase, getUnfilteredMetricsOfBase } from 'navi-reports/utils/request-metric';
 import { isEmpty } from '@ember/utils';
 
-const { assign, inject, get, set, setProperties } = Ember;
-
-const DEFAULT_DIM_FILTER = {
-  operator: 'in',
-  field: 'id',
-  values: []
-};
+const { assign, inject, get, set, setProperties, assert } = Ember;
 
 const DEFAULT_METRIC_FILTER = {
   operator: 'gt',
@@ -74,8 +68,9 @@ export default ActionConsumer.extend({
      * @param {Object} dimension - dimension to filter
      */
     [RequestActions.ADD_DIM_FILTER]: ({ currentModel }, dimension) => {
+      assert("Dimension model has correct primaryKeyFieldName", typeof get(dimension, 'primaryKeyFieldName') === 'string');
       get(currentModel, 'request').addFilter(
-        assign({ dimension }, DEFAULT_DIM_FILTER)
+        { dimension, operator: 'in', field: get(dimension, 'primaryKeyFieldName'), values: [] }
       );
     },
 
--- a/packages/visualizations/addon/components/cell-renderers/dimension.js
+++ b/packages/visualizations/addon/components/cell-renderers/dimension.js
@@ -10,14 +10,65 @@
  * }}
  */
 
-import Ember from 'ember';
+import Component from '@ember/component';
+import { computed, get } from '@ember/object'
 import layout from '../../templates/components/cell-renderers/dimension';
 
-export default Ember.Component.extend({
+export default Component.extend({
   layout,
 
   /**
    * @property {Array} classNames - list of component class names
    */
-  classNames: ['table-cell-content', 'dimension']
+  classNames: ['table-cell-content', 'dimension'],
+
+  /**
+   * @property {String} descField - field to find the description
+   */
+  descField: computed('column.field.dimension', function() {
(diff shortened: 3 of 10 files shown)


packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -72,10 +72,14 @@ function _fakeDimensionValues(name, count) {
   let fakeValues = [];
 
   for (let i = 0; i < count; i++) {
-    fakeValues.push({
+    let key = null;
+    if(name === 'multiSystemId') {
```
balajimanoharan: Can we add some comments around this?

packages/reports/addon/consumers/request/filter.js
```diff
@@ -73,9 +73,9 @@ export default ActionConsumer.extend({
      * @param {Object} route - route that has a model that contains a request property
      * @param {Object} dimension - dimension to filter
      */
-    [RequestActions.ADD_DIM_FILTER]: ({ currentModel }, dimension) => {
+    [RequestActions.ADD_DIM_FILTER]: ({ currentModel }, dimension)   => {
       get(currentModel, 'request').addFilter(
-        assign({ dimension }, DEFAULT_DIM_FILTER)
+        assign({ dimension}, DEFAULT_DIM_FILTER, {field: get(dimension, 'primaryKeyFieldName') || 'id'})
```
balajimanoharan: `primaryKeyFieldName` already defaults to `id`.
https://github.com/yahoo/navi/blob/fc6bb19c3b376c9309f49911aa1f79c6291e3a65/packages/data/addon/models/metadata/dimension.js#L87

packages/reports/addon/consumers/request/filter.js
```diff
@@ -27,7 +27,7 @@ export default ActionConsumer.extend({
   /**
    * @property {Ember.Service} requestActionDispatcher
    */
-  requestActionDispatcher: inject.service(),
+  requestActionDispatcher: inject.service(),
```
balajimanoharan: Can we revert this change?

packages/visualizations/addon/templates/components/cell-renderers/dimension.hbs
```diff
@@ -1,6 +1,10 @@
 {{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
 {{#with (get data (concat column.field.dimension '|desc' )) as |description|}}
-    <span title={{description}}>{{description}}</span>
+    {{#with (get data (concat column.field.dimension '|id' )) as |id|}}
+        <span title={{id}}>{{description}}</span>
```
balajimanoharan: Can we do `{{description}} ({{id}})` in the title? I think we follow the same format elsewhere.

packages/visualizations/addon/templates/components/cell-renderers/dimension.hbs
```diff
@@ -1,6 +1,10 @@
 {{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
 {{#with (get data (concat column.field.dimension '|desc' )) as |description|}}
-    <span title={{description}}>{{description}}</span>
+    {{#with (get data (concat column.field.dimension '|id' )) as |id|}}
```
balajimanoharan: This is done on line #9 as well. Can we make this the outer `with`?

packages/reports/addon/consumers/request/filter.js
```diff
@@ -73,9 +73,9 @@ export default ActionConsumer.extend({
      * @param {Object} route - route that has a model that contains a request property
      * @param {Object} dimension - dimension to filter
      */
-    [RequestActions.ADD_DIM_FILTER]: ({ currentModel }, dimension) => {
+    [RequestActions.ADD_DIM_FILTER]: ({ currentModel }, dimension)   => {
```
jkusa: Can we revert this change?

packages/reports/addon/consumers/request/filter.js
jkusa: Does it make sense to have `DEFAULT_DIM_FILTER` now?

cythrawll: So doing that assumes that the consumer is a dimension object, and in practice it seems that it is from where I tested it. However, a lot of tests pass in just the dimension name.  Is this bad mocks? or will we actually run into dimension simply being a string?

packages/visualizations/addon/templates/components/cell-renderers/dimension.hbs
cythrawll: we do, but only when desc and ID are far from each other.  this isn't the case here.

packages/visualizations/addon/templates/components/cell-renderers/dimension.hbs
cythrawll: would that actually fix conditions for all 4 states?

packages/reports/addon/consumers/request/filter.js
jkusa: Bad mocks I hope. I might be time to consider propTypes or flow 😬

packages/visualizations/addon/templates/components/cell-renderers/dimension.hbs
cythrawll: gonna refactor this so that the component makes this decision rather than the template.

packages/visualizations/addon/components/cell-renderers/dimension.js
```diff
@@ -10,14 +10,63 @@
...
+   * @property {String} idField - field to find id
+   */
+  idField: computed('column.field.dimension', function() {
+    return get(this, 'column.field.dimension') + '|id';
+  }),
+
+  /**
+   * @property {String} title - value that should be used in hoverover title
+   */
+  title: computed('descField', 'idField', 'data', function() {
```
balajimanoharan: screen shot 2018-08-10 at 10 23 55 am

I wasn't sure what you meant by that. But I will explain what I meant.
In the table, unless you choose an column that displays the `id`, the `id` is not seen when a description is available. When we talked to some of the spotlight users, they were more familiar with the `id` than the `desc`. So on hover, if we can show both, that would be an improvement. `description (id)`.

packages/visualizations/addon/components/cell-renderers/dimension.js
```diff
@@ -10,14 +10,63 @@
...
+        idField = get(this, 'idField'),
+        data = get(this, 'data');
+
+    if(data[idField]) {
+      return data[idField];
+    } else if(data[descField]) {
+      return data[descField];
+    }
+
+    return '--';
```
balajimanoharan: Do we need a `title` in this case?

packages/visualizations/addon/components/cell-renderers/dimension.js
```diff
@@ -10,14 +10,63 @@
...
+  }),
+
+  /**
+   * @property {String} value - value that should be displayed in table cell
+   */
+  value: computed('descField', 'idField', 'data', function() {
```
balajimanoharan: Is this the exact same as `title`?

balajimanoharan: Do we already have tests for these changes?

packages/visualizations/addon/components/cell-renderers/dimension.js
cythrawll: fixed

packages/visualizations/addon/components/cell-renderers/dimension.js
cythrawll: I was just saying desc is right next to the id in this case during hover over, but I changed this behavior. to be `desc (id)` if both are present.

packages/visualizations/addon/components/cell-renderers/dimension.js
cythrawll: display priority is different.

Tests added for titles

PR #171, opened 2018-08-17
Update: Refactor report deletion to use a consumer

mattgraham1995:
This is the first step in moving away from route-actions in the `reports` route. 
The delete report action will now be handled in a report action consumer. 
I renamed the `report-action` helper to `update-report-action` since that helper fires off the given action when the `onUpdateReport` action is called. I made my own helper called `report-action` that will immediately fire off the passed action. 
Also, I moved the `buildReportUrl` action into the `navi-action-list` component directly.


Xuefeng-Zhu:
So the ultimate goal is replacing `update-report-action` in this PR with `report-action `?

mattgraham1995:
Issue added. #172

packages/reports/addon/helpers/report-action.js
```diff
@@ -1,26 +1,29 @@
...
+   * @returns {Function} Closure action
    */
-  compute([reportAction, ...params]) {
-    let actionName = ReportActions[reportAction];
-    Ember.assert(`The action name "${actionName}" is not a valid report action`, actionName);
-    return this._super([ROUTE_ACTION, actionName, ...params]);
+  compute([action, ...params]) {
+    let actionName = ReportActions[action];
+    assert(`The action name "${action}" is not a valid report action`, actionName);
+    return (() => get(this, 'reportActionDispatcher').dispatch(actionName, ...params));
```
balajimanoharan: I am not sure if this is right? Do we want to create a new dispatcher service for the report-actions?
This service is configured to dispatch actions to update report visualization and request attributes.

This seems like its the same as `update-report-action` except we have removed the middle man `onUpdateReport`.
https://github.com/yahoo/navi/blob/7acd8157bbde0414f1492c0bf80b00d75e17225f/packages/reports/addon/routes/reports/report.js#L223

mattgraham1995: It is essentially the same as `update-report-action` but it just returns a function instead of extending a route-action. The service exists independently of the helper/route-action so I don't see why we'd need a separate one. It works the same way as all the other action-dispatchers that we have.

balajimanoharan: True, but the dispatcher that we have right now dispatches route action. By adding this new consumer we have introduced a normal action that would also be dispatched by the same dispatcher. I feel that the contract for the consumer actions could get confusing if we allow multiple different types of actions to be handled by the same dispatcher, unless ofcourse if we can come up with a contract that is suitable.

mattgraham1995: Basically, calling `update-report-action` returns the `onUpdateReport` route-action after appending a list of actions to dispatch from within that route action. The actions dispatched by `report-action-dispatcher` are still normal actions.

balajimanoharan: Sure, maybe the term route action was incorrect. I mean all the actions currently are always passed a route as the first argument, so they are always associated with a route.

mattgraham1995: I don't see where we're passing a route/routeName anywhere. In the `update-report-action` helper we pass that `onUpdateReport` route-action name, but I don't see anywhere else. The dispatcher takes the same arguments that all the other dispatchers take so I'm not sure how it's any different besides being utilized by the `update-report-action` helper

balajimanoharan: I have the link above where we pass in `this`, which is the route.

packages/reports/addon/helpers/report-action.js
```diff
@@ -1,26 +1,29 @@
...
-   * @returns {Function} - Closure action
+   * @method reportAction
+   * @param {Array} array with the name of the action to dispatch
+   * @returns {Function} Closure action
    */
-  compute([reportAction, ...params]) {
-    let actionName = ReportActions[reportAction];
-    Ember.assert(`The action name "${actionName}" is not a valid report action`, actionName);
-    return this._super([ROUTE_ACTION, actionName, ...params]);
+  compute([action, ...params]) {
```
jkusa: When making an action helper, make sure to:
- execute in the runloop: 
https://github.com/DockYard/ember-route-action-helper/blob/v2.0.6/addon/helpers/route-action.js#L50

- make function a "real" action 
https://github.com/DockYard/ember-route-action-helper/blob/v2.0.6/addon/helpers/route-action.js#L53

packages/reports/addon/components/navi-action-list.js
```diff
@@ -3,15 +3,37 @@
...
    */
   tagName: 'ul',
+
+  actions: {
+    /**
+     * @action buildReportUrl
+     * @param {Object} report - model with id
+     * @returns {String} url for given report
+     */
+    buildReportUrl(report) {
```
jkusa: Could this have been in a separate commit?

packages/reports/addon/services/report-action-dispatcher.js
```diff
@@ -1,20 +1,23 @@
...
 
 export default RequestActionDispatcher.extend({
+
   /**
    * @property {Array} consumers - concatenate consumers in request
    */
   consumers: [
-    'report/table-visualization'
+    'report/table-visualization',
+    'report/report'
```
jkusa: Don't these two consumers have different contracts?

https://github.com/yahoo/navi/blob/master/packages/reports/addon/consumers/report/table-visualization.js#L22

Also, the `RequestActionDispatcher` inherits all  `RequestActionDispatcher` consumers.

packages/reports/addon/helpers/report-action.js
```diff
@@ -6,7 +6,13 @@ import Helper from '@ember/component/helper';
 import { inject as service } from '@ember/service';
 import { get } from '@ember/object';
 import { assert } from '@ember/debug';
+import { run } from '@ember/runloop';
 import { ReportActions } from 'navi-reports/services/report-action-dispatcher';
+import Ember from 'ember';
```
jkusa: Can we do: 
```js
import { join } from '@ember/runloop';
```

packages/reports/addon/helpers/report-action.js
```diff
@@ -1,26 +1,41 @@
...
-import RouteAction from 'ember-route-action-helper/helpers/route-action';
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+import { get } from '@ember/object';
+import { assert } from '@ember/debug';
+import { join } from '@ember/runloop';
 import { ReportActions } from 'navi-reports/services/report-action-dispatcher';
+import Ember from 'ember';
+
+const ClosureActionModule = 'ember-glimmer/helpers/action' in Ember.__loader.registry ?
```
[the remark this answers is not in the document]
balajimanoharan: This should be in all CAPS.

packages/reports/tests/unit/consumers/report/report-test.js
```diff
@@ -0,0 +1,153 @@
...
+  let Consumer = this.subject({
+    naviNotifications: {
+      add({ message }) {
+        assert.equal(message,
+          'Report "Hyrule News" deleted successfully!',
+          'A notification is sent containing the report title');
+      }
+    },
+    router: {
+      transitionTo() { }
```
balajimanoharan: transitionTo: () => {}

packages/reports/addon/helpers/report-action.js
```diff
@@ -1,26 +1,41 @@
...
    */
-  compute([reportAction, ...params]) {
-    let actionName = ReportActions[reportAction];
-    Ember.assert(`The action name "${actionName}" is not a valid report action`, actionName);
-    return this._super([ROUTE_ACTION, actionName, ...params]);
+  compute([action, ...params]) {
+    let actionName = ReportActions[action];
+    assert(`The action name "${action}" is not a valid report action`, actionName);
+    let reportAction = (() => join(() => get(this, 'reportActionDispatcher').dispatch(actionName, ...params))),
+        ACTION = ClosureActionModule.ACTION;
```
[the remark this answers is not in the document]
mattgraham1995: Same name as the property it given the value of.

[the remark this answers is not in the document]
balajimanoharan: yeah, but its kind of misleading. We have been using CAPS for global vars

balajimanoharan:
And your build is failing.

mattgraham1995:
Oops forgot to squash before merge.

Issue #172, opened 2018-08-21
Directory Route: Move needed actions to consumers and subroutes


mattgraham1995:
**Problem: The new directory route does not have the actions defined that are required for the navi-actions-list and dashboard-actions-list components.** 

These actions are defined as actions in the reports and dashboards routes. 

Simple Solution: Add all route actions from reports and dashboards routes to the directory route. 

Problems with this solution:

- Fills the directory route with route-actions used for deleting reports/dashboards and handling their delivery rules
- Multiple copies of these same actions will be necessary to support use in the report/dashboard view as well as in the directory view
    - E.g. deleteReport will be an action defined in the directory route and the reports route. 
    - This wouldn’t be a problem if the report view route was a child of the directory route, but that isn’t necessarily a good solution either. 

**Solution to avoid these problems**: Create consumers to handle any report or dashboard actions (this will just include their deletion actions for now) and create a subroute for the schedule modal that will contain the actions needed for handling delivery rules.

**Necessary changes:** 

- Two new action consumers
    - navi-reports/consumers/report/report
    - navi-dashboards/consumers/dashboard/dashboard
- One new action dispatcher service
    - navi-reports/services/report-action-dispatcher (Already exists, will just register new  consumer with this)
    - navi-dashboards/services/dashboard-action-dispatcher
- Two new helpers
    - navi-dashboards/helpers/dashboard-action
        - Validates an action name and then passes it to dashboard-action-dispatcher for dispatching
    - navi-reports/helpers/report-action
        - This helper already exists, but the existing helper will be renamed to “update-report-action”. This is because the existing helper sends an action to the route that will dispatch when the route-action “onUpdateReport” is called.
        - Validates an action name and then passes it to report-action-dispatcher for immediate dispatching
- One new route
    - navi-reports/routes/schedule
        - Will use the routable modal from liquid-tether and contain all the delivery-rule actions

PR #173, opened 2018-08-21
Update: Refactor dashboard deletion to use a consumer

mattgraham1995:
See #172 

This covers the dashboard section of that issue and is related to PR #171


mattgraham1995:
@balajimanoharan I had considered that when creating the action dispatchers, but my thinking was that we wanted to keep dashboard related services/helpers in the dashboards package and reports stuff in the reports package. I suppose I could just make a single dispatcher in navi-core though

balajimanoharan:
The dashboards addon inherits everything in reports already. So we could just maintain this dispatcher in the reports addon and make use of it here.

packages/dashboards/addon/components/dashboard-action-list.js
```diff
@@ -1,14 +1,42 @@
...
+   */
+  router: service(),
+
+  actions: {
+    /**
+     * @action buildDashboardUrl
+     * @param {Object} dashboard - model with id
+     * @returns {String} url for given dashboard
+     */
+    buildDashboardUrl(dashboard) {
```
balajimanoharan: Isn't this also similar to the method for reports?
Is it possible to extend that and make use of it here, since our route structures are very similar.

packages/dashboards/addon/templates/components/dashboard-action-list.hbs
```diff
@@ -65,7 +65,7 @@
       classNames='action delete'
       model=dashboard
       warnMsg='Are you sure you want to delete this dashboard and all of its widgets?'
-      deleteAction=(route-action 'deleteDashboard' dashboard)
+      deleteAction=(item-action 'DELETE_ITEM' dashboard)
```
balajimanoharan: I feel like we need a more descriptive name for these type of actions instead of just `item-action`.

packages/dashboards/tests/acceptance/share-link-test.js
```diff
@@ -0,0 +1,26 @@
+import { test } from 'qunit';
+import { teardownModal } from '../helpers/teardown-modal';
+import moduleForAcceptance from '../../tests/helpers/module-for-acceptance';
+
+moduleForAcceptance('Acceptance | share link', {
+  beforeEach() {
+    return visit('/dashboards');
```
balajimanoharan: Can we do this inside the test, since there is only one.?

mattgraham1995:
I agree we need better names, but that's been an unanswered question for a while: What do we call a parent class of reports and dashboards? The only answers that I've heard so far are `document`, `item` or `deliverableItem`. Item is vague, but I don't see a clearly better alternative right now unless you've got one in mind.

packages/dashboards/addon/components/dashboard-action-list.js
mattgraham1995: I suppose I could create a base component that both `navi-action-list` and `dashboard-action-list` extend from, but that seems unnecessary two combine two small actions that were separately defined in the routes before my changes.

balajimanoharan: You don't need a base component, how about something simple like a mixin with an action/method that would have the buildUrl logic.

PR #174, opened 2018-08-22
Fix: Removing a parametrized metric now removes sort

daniellevinson:


packages/reports/addon/consumers/request/sort.js
```diff
@@ -46,16 +46,42 @@ export default ActionConsumer.extend({
...
     /**
      * @action REMOVE_METRIC
      * @param {Object} route - route that has a model that contains a request property
      * @param {Object} metric - metadata model of metric to remove
      */
     [RequestActions.REMOVE_METRIC](route, metric) {
       // Find and remove any `sorts` attached to the metric
-      let metricName = get(metric, 'name');
+      if (get(metric, 'hasParameters')) {
+        // מot passing parameters to remove sorts with any parameter
```
mattgraham1995: What is this funky `n` character? `מ`

packages/reports/addon/consumers/request/sort.js
```diff
@@ -46,16 +46,42 @@ export default ActionConsumer.extend({
...
+    },
+    /**
+     * @action REMOVE_METRIC_WITH_PARAM
+     * @param {Object} route - route that has a model that contains a request property
+     * @param {Object} metric - metadata model of metric to remove
+     * @param {Object} parameters - metric parameters
+     */
+    [RequestActions.REMOVE_METRIC_WITH_PARAM](route, metric, parameters) {
+      // Find and remove any `sorts` attached to the metric and parameters
+      get(this, 'requestActionDispatcher').dispatch(RequestActions.REMOVE_SORT_WITH_PARAM, route, metric, parameters);
```
mattgraham1995: If all this action does is dispatch another action, isn't this action just a middle man?

mattgraham1995: Nevermind, I see what's going on

packages/reports/addon/models/bard-request/request.js
```diff
@@ -459,6 +459,27 @@ export default Fragment.extend(Validations, {
...
+   * Removes the sort from the sort array using the metric model and parameters
+   *
+   * @method removeSortMetricByModelAndParam
+   * @param {DS.Model} metricModel
+   * @param {Object} [parameters] - parameters [optional]
+   * @returns {Void}
+   */
+  removeSortMetricByModelAndParam(metricModel, parameters) {
+    if (parameters) {
+      let canonicalizedMetric = canonicalizeMetric({
```
mattgraham1995: I think we can just call `get(metricModel, 'canonicalName')`  instead of `canonicalizeMetric`

packages/reports/tests/unit/consumers/request/sort-test.js
```diff
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import { moduleFor, test } from 'ember-qunit';
 import { RequestActions } from 'navi-reports/services/request-action-dispatcher';
 
-const { get, getOwner } = Ember;
+const { getOwner } = Ember;
```
mattgraham1995: We're slowly moving towards using the new ember import styles so could we change this to
`import { getOwner } from '@ember/application'`
and then get rid of the `import Ember from 'ember'`?

packages/reports/addon/consumers/request/sort.js
```diff
@@ -46,16 +46,42 @@ export default ActionConsumer.extend({
 
       request.removeSortByMetricName(metricName);
     },
+    /**
```
cythrawll: Make sure you update copyright in changed files.

packages/reports/addon/models/bard-request/request.js
cythrawll: This metric should have a computed property of 'canonicalName', see if you can use that instead

packages/reports/addon/consumers/request/sort.js
daniellevinson: The year?

mattgraham1995: Yeah, whenever we update a file, just bump the year to 2018

packages/reports/tests/unit/consumers/request/sort-test.js
```diff
@@ -50,52 +50,112 @@ test('UPSERT_SORT', function(assert) {
...
 
-  consumer.send(RequestActions.REMOVE_SORT, { currentModel }, 'click', 'desc');
+  consumer.send(RequestActions.REMOVE_SORT, { currentModel }, 'click');
+});
+
+test('REMOVE_SORT_WITH_PARAM', function(assert) {
+  assert.expect(2);
+
+  let parameters = { currency: 'USD' },
+      RevenueUSD = {
```
Xuefeng-Zhu: Why capitalize first letter?

packages/reports/addon/models/bard-request/request.js
daniellevinson: so actually we don't have `canonicalName` here, maybe that's why `removeRequestMetricWithParam` was written the same way.

packages/reports/addon/models/bard-request/request.js
```diff
@@ -459,6 +459,27 @@ export default Fragment.extend(Validations, {
...
+        metric: get(metricModel, 'name'),
+        parameters
+      });
+      this.removeSortByMetricName(canonicalizedMetric);
+    } else {
```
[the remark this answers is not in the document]
daniellevinson: because `removeSortByMetricName` finds by `canonicalName`, while we want to filter by the name and remove sorted metrics with all parameters. In other words the if block is called with specific parameter, the else block is called when you unselect the metric itself when it can have more than one param selected

packages/reports/tests/unit/consumers/request/sort-test.js
daniellevinson: it's all over so I didn't change it but not really necessary

packages/reports/addon/models/bard-request/request.js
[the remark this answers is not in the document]
balajimanoharan: hmm.. makes sense. My concern is that the method name now can be a little confusing. its called `removeSortMetricWithParam` where parameters is an optional argument.

packages/reports/addon/models/bard-request/request.js
```diff
@@ -459,6 +459,27 @@ export default Fragment.extend(Validations, {
     return this.removeSort(sortObj);
   },
 
+  /**
+   * Removes the sort from the sort array using the metric model and parameters
+   *
+   * @method removeSortMetricWithParam
+   * @param {DS.Model} metricModel
```
balajimanoharan: Can we add a description to this to say `metric metadata model`? We have two metric model, one is a fragment which is part of the request, the other is an `Ember.Object`, the metric metadata model.

I see other places in this file have this too, but can we correct this line to say `{Ember.Object}` instead of `{DS.Model}`.

packages/reports/addon/models/bard-request/request.js
```diff
@@ -2,21 +2,21 @@
...
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
-
-import Ember from 'ember';
 import DS from 'ember-data';
 import MF from 'model-fragments';
 import isEqual from 'lodash/isEqual';
 import { validator, buildValidations } from 'ember-cp-validations';
 import Interval from 'navi-core/utils/classes/interval';
+import { A, makeArray } from '@ember/array';
```
jkusa: Small thing: we usually rename `A` to `arr`

```js
import { A as arr } from '@ember/array';
```
https://github.com/yahoo/navi/search?q=A%3Aarr&unscoped_q=A%3Aarr

packages/reports/tests/unit/consumers/request/sort-test.js
```diff
@@ -50,52 +52,112 @@ test('UPSERT_SORT', function(assert) {
...
+});
+
+test('REMOVE_SORT_WITH_PARAM', function(assert) {
+  assert.expect(2);
+
+  let parameters = { currency: 'USD' },
+      revenueUSD = {
+        metric: { name: 'revenue' },
+        parameters,
+        hasParameters: true,
```
jkusa: What shape is this object? [Request Metric](https://github.com/yahoo/navi/blob/fc6bb19c3b376c9309f49911aa1f79c6291e3a65/packages/reports/addon/models/bard-request/fragments/metric.js) or [Metric Metadata](https://github.com/yahoo/navi/blob/fc6bb19c3b376c9309f49911aa1f79c6291e3a65/packages/data/addon/models/metadata/metric.js)?

jkusa:
Can we add an acceptance test with using the table?

packages/reports/addon/models/bard-request/request.js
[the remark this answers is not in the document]
balajimanoharan: https://github.com/yahoo/navi/pull/174/files#diff-c41a86855938e5b6dd5b8089ddb920caL458
One suggestion, can we update this method to remove all sorts with the same metricName.
Switch the `findBy` to `filterBy` and loop through the sorts and remove them.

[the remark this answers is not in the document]
daniellevinson: did you mean `removeSortByMetricName`?

packages/reports/addon/models/bard-request/request.js
daniellevinson: 👍

packages/reports/addon/models/bard-request/request.js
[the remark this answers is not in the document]
balajimanoharan: Yes, can we move the else logic there?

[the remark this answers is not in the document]
daniellevinson: Sounds like a good idea

packages/reports/tests/unit/consumers/request/sort-test.js
daniellevinson: didn't notice we have two models (@balajimanoharan also mentioned it in another comment here), I'll check again and adjust the test and method description accordingly

packages/reports/addon/models/bard-request/request.js
daniellevinson: Sure, still need to familiarize myself with the project a little more though :)

packages/reports/tests/unit/consumers/request/sort-test.js
daniellevinson: it's Metric Metadata, I'll remove the `canonicalName` property

packages/reports/addon/models/bard-request/request.js
```diff
@@ -459,6 +460,28 @@ export default Fragment.extend(Validations, {
...
+   */
+  removeSortMetricWithParam(metricModel, parameters) {
+    let sorts = get(this, 'sort').filterBy('metric.metric', metricModel),
+        canonicalizedMetric = canonicalizeMetric({
+          metric: get(metricModel, 'name'),
+          parameters
+        });
+
+    sorts.forEach(sortMetric => {
+      if(!parameters || get(sortMetric, 'metric.canonicalName') === canonicalizedMetric) {
```
balajimanoharan: I still feel that removing a metric without parameters should not be a part of this method.

https://gist.github.com/balajimanoharan/9f6319eb0371cf72ad4099294219fba3
Would something like this work?

daniellevinson: Yeah looks good, I'll refactor

daniellevinson:
@balajimanoharan it does, when clicking a table header (from `headerClicked`)

packages/reports/tests/acceptance/report-visualizations-test.js
```diff
@@ -69,18 +69,120 @@ test('Table column sort', function(assert) {
...
+    click('.metric-config__dropdown-container .grouped-list__item:contains(Euro) .grouped-list__item-label');
+    click('.navi-report__run-btn');
+  });
+
+  andThen(() => {
+    assert.notOk(find('.table-header-cell.metric:contains(Platform Revenue (USD)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--desc'),
+      'Metric with first parameter is not sorted desc');
+    assert.notOk(find('.table-header-cell.metric:contains(Platform Revenue (USD)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--asc'),
+      'Metric with first parameter is not sorted asc');
+    assert.notOk(find('.table-header-cell.metric:contains(Platform Revenue (EUR)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--desc'),
```
jkusa: Can we add a newline between column asserts for readability?

packages/reports/tests/acceptance/report-visualizations-test.js
```diff
@@ -69,18 +69,120 @@ test('Table column sort', function(assert) {
...
+  andThen(() => {
+    assert.ok(find('.table-header-cell.metric:contains(Platform Revenue (USD)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--desc'),
+      'Metric with first parameter is sorted desc after first sort click');
+    assert.notOk(find('.table-header-cell.metric:contains(Platform Revenue (USD)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--asc'),
+      'Metric with first parameter is not sorted asc after first sort click');
+    assert.notOk(find('.table-header-cell.metric:contains(Platform Revenue (EUR)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--desc'),
+      'Metric with second parameter is not sorted desc after first sort click');
+    assert.notOk(find('.table-header-cell.metric:contains(Platform Revenue (EUR)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--asc'),
+      'Metric with second parameter is not sorted asc after first sort click');
+    click('.navi-report__copy-api-btn .get-api__btn');
```
jkusa: Can the`click` stmts can be move outside the `andThen`? They will still execute in the same order

packages/reports/tests/acceptance/report-visualizations-test.js
```diff
@@ -69,18 +69,120 @@ test('Table column sort', function(assert) {
...
+  andThen(() => {
+    assert.ok(find('.table-header-cell.metric:contains(Platform Revenue (USD)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--desc'),
+      'Metric with first parameter is sorted desc after sorting by both parameters');
+    assert.ok(find('.table-header-cell.metric:contains(Platform Revenue (EUR)) .navi-table-sort-icon').hasClass('navi-table-sort-icon--desc'),
+      'Metric with first parameter is sorted desc after sorting by both parameters');
+    // remove the metric
+    click('.report-builder__metric-selector .grouped-list__item:contains(Platform Revenue) .grouped-list__item-label');
+    click('.navi-report__copy-api-btn .get-api__btn');
+  });
+
```
jkusa: Can we also verify that the table returns and not an error?

PR #175, opened 2018-08-27
Update: Ember-tooltips version to support Ember 3.x

mattgraham1995:
The fix for Ember 3.0 that I submitted to ember-tooltips got merged and they came out with a release today.

### DIFF
--- a/packages/dashboards/package.json
+++ b/packages/dashboards/package.json
@@ -71,7 +71,7 @@
     "ember-tag-input": "^1.2.1",
     "ember-tether": "^0.4.1",
     "ember-toggle": "5.2.1",
-    "ember-tooltips": "^2.9.2",
+    "ember-tooltips": "^2.10.0",
     "ember-truth-helpers": "^2.0.0",
     "ember-uuid": "^1.0.0",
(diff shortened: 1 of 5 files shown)


mattgraham1995:
I could do `~2.10.0` but I don't imagine there will be a `2.10.1` release before `3.x` is released. We need 2.10 or above for the Ember 3.x support.

packages/dashboards/package.json
```diff
@@ -71,9 +71,10 @@
     "ember-tag-input": "^1.2.1",
     "ember-tether": "^0.4.1",
     "ember-toggle": "5.2.1",
-    "ember-tooltips": "^2.9.2",
+    "ember-tooltips": "^2.10.0",
     "ember-truth-helpers": "^2.0.0",
     "ember-uuid": "^1.0.0",
+    "klaw": "^3.0.0",
```
balajimanoharan: What is this needed for?

mattgraham1995: When I ran dashboards before, it told me it was missing that package, but it doesn't seem to need it now 🤔

PR #181, opened 2018-09-07
Update: Move dependecies from blueprints to direct npm dependencies

balajimanoharan:
Moving dependencies from blueprints to direct npm dependencies
Fixes #179

PR #184, opened 2018-09-10
Update: Remove blueprints navi-data

balajimanoharan:
Removing blueprints from repositories individually


balajimanoharan:
Approved in #181

PR #185, opened 2018-09-10
Tooltip data binding fix

cythrawll:
Found issue with the data that was being closed over by the tooltip component could become out of date.  To fix this we bind the object instead.  Also, we use a unique registry entry for the component so tooltips have the right data bound when there are multiple charts.

### DIFF
--- a/packages/core/addon/chart-builders/date-time.js
+++ b/packages/core/addon/chart-builders/date-time.js
@@ -242,7 +242,7 @@ export default Ember.Object.extend({
    * @returns {Object} object with tooltip template and rendering context
    */
   buildTooltip() {
-    let byXSeries = get(this, 'byXSeries');
+    let builder = this;
 
     return Ember.Mixin.create({
       layout: tooltipLayout,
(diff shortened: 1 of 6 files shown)


packages/core/tests/unit/components/navi-visualizations/line-chart-test.js
```diff
@@ -376,3 +378,138 @@ test('single data point', function(assert) {
...
+      index: 5,
+      seriesIndex: 0,
+      value: 172933788,
+      x: '2016-06-03 00:00:00.000'
+    }],
+    x: '2016-06-03 00:00:00.000',
+    seriesConfig: (component.get('seriesConfig')),
+    request
+  });
+  Ember.run(() => {
```
mattgraham1995: Could we import run directly?

packages/core/addon/chart-builders/date-time.js
```diff
@@ -253,7 +253,7 @@ export default Ember.Object.extend({
       rowData: Ember.computed('x', 'tooltipData', function() {
         return get(this, 'tooltipData').map(series => {
           // Get the full data for this combination of x + series
-          let dataForSeries = byXSeries.getDataForKey(get(this, 'x') + series.name) || [];
+          let dataForSeries = get(builder, 'byXSeries').getDataForKey(get(this, 'x') + series.name) || [];
```
Xuefeng-Zhu: why is this change necessary?

cythrawll: Because closing over the data as a pojo in the tooltip builder, means that the tooltip won't see updates when `byXSeries` in the object changes. Therefore, if `byXSeries` changes the tooltip can't access the row data correctly.  By closing over the builder instead it can always extract up-to-date `byXSeries`

PR #201, opened 2018-09-24
Update: Rename my-directory and lower font size

mattgraham1995:
screen shot 2018-09-24 at 2 27 44 pm

- Bring font size down to match font size from reports list
- Remove `>` symbol from sidebar until foldering is supported
- Rename `My Directory` to `My Data` based on feedback from navi users


balajimanoharan:
Do we want to eventually remove all references to directory to have routes named `data` instead?
`data.my-data`?

PR #202, opened 2018-09-24
Update: Extract computation of metric column total to an overridable method

daniellevinson:
The new public method is `computeColumnTotal`.

### DIFF
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -9,7 +9,6 @@ module.exports = {
     browser: true
   },
   rules: {
-    indent: [2, 2, { VariableDeclarator: { var: 2, let: 2, const: 3 } }],
     'multiline-comment-style': ['error', 'starred-block']
   },
   globals: {
--- a/packages/core/addon/components/navi-visualizations/table.js
+++ b/packages/core/addon/components/navi-visualizations/table.js
@@ -25,14 +25,14 @@ import { canonicalizeMetric } from 'navi-data/utils/metric';
 import { getColumnDefaultName } from 'navi-core/helpers/default-column-name';
 
 const NEXT_SORT_DIRECTION = {
-  'none': 'desc',
-  'desc': 'asc',
-  'asc': 'none',
+  none: 'desc',
+  desc: 'asc',
+  asc: 'none'
 };
 
 const HEADER_TITLE = {
-  'grandTotal': 'Grand Total',
-  'subtotal': 'Subtotal'
+  grandTotal: 'Grand Total',
+  subtotal: 'Subtotal'
 };
 
 export default Component.extend({
@@ -66,7 +66,9 @@ export default Component.extend({
   /**
    * @property {Number} totalRows - total rows for the request
    */
-  totalRows: computed.readOnly('model.firstObject.response.meta.pagination.numberOfResults'),
+  totalRows: computed.readOnly(
+    'model.firstObject.response.meta.pagination.numberOfResults'
+  ),
 
   /**
    * @property {Number} rowsInResponse - rows in response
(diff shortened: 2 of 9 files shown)


packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -66,63 +66,96 @@ export default Component.extend({
...
+   * Returns sum of values, may be overriden with your own logic
+   *
+   * @param {Array} data
+   * @param {String} metricName
+   * @param {Object} totalRow
+   * @param {Object} column
+   * @param {String} type - `grandTotal` || `subtotal`
+   * @returns {*} sum
+   */
+  computeColumnTotal(data, metricName /*, totalRow, column, type*/) {
```
balajimanoharan: Why do we want this to be a public method?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -66,63 +66,96 @@ export default Component.extend({
...
+    //add totalRow indication to meta
+    set(
+      totalRow,
+      '__meta__',
+      Object.assign({}, get(totalRow, '__meta__'), { isTotalRow: true })
+    );
+
+    //if partial data add indication to meta
+    if (hasPartialData) {
+      set(totalRow, '__meta__.hasPartialData', true);
```
balajimanoharan: What are we doing with this metadata? Do we still want to display the totals even if we have partial data?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -165,32 +200,35 @@ export default Component.extend({
...
-    }
+      if (!get(this, 'options.showTotals.grandTotal')) {
+        return tableData;
+      }
 
-    return [...tableData, this._computeTotal(rawData, 'grandTotal')];
-  }),
+      return [...tableData, this._computeTotal(rawData, 'grandTotal')];
+    }
+  ),
```
balajimanoharan: Can we revert these changes?

mattgraham1995: I think prettier made these changes

packages/core/.eslintrc.js
```diff
@@ -9,7 +9,6 @@ module.exports = {
     browser: true
   },
   rules: {
-    indent: [2, 2, { VariableDeclarator: { var: 2, let: 2, const: 3 } }],
```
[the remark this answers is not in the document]
daniellevinson: Prettier changes indentations when committing the code and makes tests fail on linting.
I ran `node_modules/.bin/eslint --print-config addon | node_modules/.bin/eslint-config-prettier-check` and got 
```
The following rules are unnecessary or might conflict with Prettier:

- indent
```

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -66,63 +66,96 @@ export default Component.extend({
...
+   * @param {String} metricName
+   * @param {Object} totalRow
+   * @param {Object} column
+   * @param {String} type - `grandTotal` || `subtotal`
+   * @returns {*} sum
+   */
+  computeColumnTotal(data, metricName /*, totalRow, column, type*/) {
+    return data.reduce((sum, row) => {
+      let number = Number(row[metricName]);
+      if (!Number.isNaN(number)) {
```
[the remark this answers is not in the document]
daniellevinson: 👍

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: Currently we're not displaying totals if we have partial data, I haven't changed this behavior. Only moved it out of the reduce function, setting it once per row is enough.

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: yeah, prettier... haven't changed anything here

PR #204, opened 2018-09-25
 Update: Prettier all the things  

jkusa:
- Prettier all the things


jkusa:
@balajimanoharan the only change I made was setting the the `printWidth` to `120`:

https://github.com/yahoo/navi/blob/ca8bef67c77c11fe8519f9175c839d2bdd46b2bf/prettier.config.js#L3

Build is passing now.

PR #213, opened 2018-09-28
Fix: Unsaved report links in directory

mattgraham1995:
An issue was found where trying to open an unsaved report from the directory view wouldn't do anything. This should restore that functionality and I added a couple tests.

image
Unsaved report at the bottom

### DIFF
--- a/packages/directory/addon/components/dir-item-name-cell.js
+++ b/packages/directory/addon/components/dir-item-name-cell.js
@@ -34,6 +34,11 @@ export default Component.extend({
   }),
 
   /**
+   * @property {String} itemId - the id of the model or the tempId of an unsaved model
+   */
+  itemId: computed.oneWay('value.modelId'),
+
+  /**
    * @property {String} type - the type of the item
    */
   type: computed('value', function() {
--- a/packages/directory/addon/templates/components/dir-item-name-cell.hbs
+++ b/packages/directory/addon/templates/components/dir-item-name-cell.hbs
@@ -1,8 +1,12 @@
 {{!-- Copyright 2018, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
-{{#link-to itemLink (get value 'id') classNames='dir-item-name-cell'}}
+{{#link-to itemLink itemId classNames='dir-item-name-cell'}}
   {{navi-icon iconClass classNames=(concat 'dir-item-name-cell__icon dir-icon__' type)}}
   {{get value 'title'}}
-  {{#if (get value 'isFavorite')}}
-    {{navi-icon 'star' classNames='dir-item-name-cell__fav-icon dir-icon__favorites'}}
-  {{/if}}
 {{/link-to}}
+
+{{#if (get value 'tempId')}}
+  <span class='dir-item-name-cell__unsaved-label'>Unsaved</span>
+{{/if}}
+{{#if (get value 'isFavorite')}}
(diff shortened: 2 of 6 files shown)


packages/directory/addon/components/dir-item-name-cell.js
```diff
@@ -33,6 +33,15 @@ export default Component.extend({
     return `${type}.${singularType}`;
   }),
 
+  /**
+   * @property {String} itemId - the id of the model or the tempId of an unsaved model
+   */
+  itemId: computed('value', function() {
+    let model = get(this, 'value');
+
+    return get(model, 'tempId') ? get(model, 'tempId') : get(model, 'id');
```
cythrawll: `return get(model, 'tempId') || get(model, 'id')`

mattgraham1995:
I've already made an issue for a generic id in the models. #214

packages/directory/addon/components/dir-item-name-cell.js
```diff
@@ -33,6 +33,15 @@ export default Component.extend({
     return `${type}.${singularType}`;
   }),
 
+  /**
+   * @property {String} itemId - the id of the model or the tempId of an unsaved model
+   */
+  itemId: computed('value', function() {
+    let model = get(this, 'value');
+
+    return get(model, 'tempId') || get(model, 'id');
```
balajimanoharan: It should be the other way. `id` if present, otherwise `tempId`.

packages/directory/addon/templates/components/dir-item-name-cell.hbs
```diff
@@ -1,8 +1,12 @@
...
   {{get value 'title'}}
-  {{#if (get value 'isFavorite')}}
-    {{navi-icon 'star' classNames='dir-item-name-cell__fav-icon dir-icon__favorites'}}
-  {{/if}}
 {{/link-to}}
+
+{{#if (get value 'tempId')}}
+  <span class='dir-item-name-cell__unsaved-label'>Unsaved</span>
+{{/if}}
+{{#if (get value 'isFavorite')}}
```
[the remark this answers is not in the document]
balajimanoharan: hmm. good question. 
Looks like we cannot. So this should be fine.

packages/directory/tests/unit/components/dir-item-name-cell-test.js
```diff
@@ -0,0 +1,47 @@
...
+    );
+
+    assert.equal(unsavedReportComponent.get('itemId'), '12345', 'Component uses the tempId when the id is falsy');
+
+    let savedDashboardComponent = subject.create({
+      value: {
+        serialize() {
+          return { data: { type: 'dashboards' } };
+        },
+        id: 1
```
balajimanoharan: Can we test with a model having both `tempId` and `id`?

packages/directory/addon/components/dir-item-name-cell.js
mattgraham1995: It shouldn't matter since `tempId` will be `null` if `id` is present.
https://github.com/yahoo/navi/blob/e1fdfa433f0d1038fc51345c18f7524ad04d3fc7/packages/core/addon/models/report.js#L38

packages/directory/tests/unit/components/dir-item-name-cell-test.js
mattgraham1995: That shouldn't ever happen

packages/directory/addon/components/dir-item-name-cell.js
balajimanoharan: Can we make use of `modelId` now? #217

PR #217, opened 2018-10-01
New: Add modelId property

### LINKED ISSUE #214: Create generic id computed property in models
Have one computed property in asset models that will return the temp id or id.

balajimanoharan:
Add modelId property to deliverable-item

resolves #214


packages/core/tests/unit/models/deliverable-item-test.js
```diff
@@ -0,0 +1,21 @@
+import { moduleForModel, test } from 'ember-qunit';
+
+moduleForModel('deliverable-item', 'Unit | Model | deliverable item', {
+  needs: ['model:delivery-rule']
+});
+
+test('it exists', function(assert) {
```
cythrawll: better test name

PR #218, opened 2018-10-03
Fix: Remove line leftover from rebase

mattgraham1995:
These lines were accidentally included when rebasing my occlusion PR branch. We only want one favorite icon


mattgraham1995:
@balajimanoharan No this is purely cosmetic. There were two favorite icons. What build is failing?

PR #220, opened 2018-10-04
Update: Add overrideable empty message component

cythrawll:
screen shot 2018-10-04 at 10 23 45 am


packages/directory/addon/templates/components/dir-empty.hbs
```diff
@@ -0,0 +1,3 @@
+{{#navi-info-message}}
+    <div>Welcome to Navi, get started by creating a new {{#link-to reportRoute}}report{{/link-to}}</div>
```
mattgraham1995: Is the div wrapper needed here?

packages/directory/tests/integration/components/dir-empty-test.js
```diff
@@ -0,0 +1,14 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Component | dir-empty', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
```
mattgraham1995: Could we use a better test name?

cythrawll: "Will it blend"?

cythrawll: Usually you're right, but this test description is apt in this case.

PR #228, opened 2018-10-10
Update: Metric formatter only forces decimals when needed.

cythrawll:
* Detects if value has a decimal point in it.
* Uses that knowledge to figure if it should force decimal points in the formatter for numbers 0, and 1 through 100.

### DIFF
--- a/packages/core/addon/helpers/smart-format-number.js
+++ b/packages/core/addon/helpers/smart-format-number.js
@@ -19,6 +19,7 @@ import numeral from 'numeral';
  */
 export function smartFormatNumber([value]) {
   // type safe check
+  let hasPoint = value !== undefined && JSON.stringify(value).indexOf('.') !== -1;
   if (typeof value !== 'number') {
     value = parseFloat(value);
   }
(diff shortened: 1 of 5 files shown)


packages/core/addon/helpers/smart-format-number.js
```diff
@@ -30,7 +31,7 @@ export function smartFormatNumber([value]) {
   let absValue = Math.abs(value);
 
   if (absValue === 0 || (absValue >= 1 && absValue < 100)) {
-    return numeral(value).format('0,0.00');
+    return numeral(value).format(hasPoint ? '0,0.00' : '0,0[.]00');
```
balajimanoharan: Does fili always return either an integer or a float? if it sends a mixture of integers or floats, how would this work?

cythrawll: So each cell will make it's own determination of the format.  So you can see columns formatted differently for each row if the webservice returns a mix of both.  But this puts some responsibility on the web service to return floats if the type is floaty consistently.

packages/core/addon/helpers/smart-format-number.js
```diff
@@ -19,6 +19,7 @@ import numeral from 'numeral';
  */
 export function smartFormatNumber([value]) {
   // type safe check
+  let hasPoint = value !== undefined && JSON.stringify(value).indexOf('.') !== -1;
```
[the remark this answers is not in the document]
cythrawll: So the problem is this erases the context of it coming in with decimal places in the first place (IE coming in a string of "0.0e99" which the webservice could potentially give us as a hint that it should be formatted as decimal.

PR #236, opened 2018-10-15
Update: Maybe fix build?

mattgraham1995:
Possibly fix the build issues in directory


packages/dashboards/tests/acceptance/print-dashboard-test.js
```diff
@@ -1,7 +1,11 @@
 import { test } from 'qunit';
 import moduleForAcceptance from '../../tests/helpers/module-for-acceptance';
 
-moduleForAcceptance('Acceptance | print dashboard');
+moduleForAcceptance('Acceptance | print dashboard', {
+  afterEach() {
+    server.shutdown();
```
balajimanoharan: Where is these server instances instantiated?

mattgraham1995: I'm not sure exactly, but every acceptance test was printing a warning message about having multiple Pretender instances running.

mattgraham1995: I think it's this server instance
https://github.com/samselikoff/ember-cli-mirage/blob/2703e767bcb3e47c8694691aae2c076c45fe0d6c/app/initializers/ember-cli-mirage.js#L16

mattgraham1995: We probably won't need these server.shutDown() calls once mirage is upgraded
https://github.com/samselikoff/ember-cli-mirage/issues/915

PR #237, opened 2018-10-16
Rename new route to edit

jkusa:
This PR renames the routes `{assets}.{asset}.new` to `{assets}.{asset}.edit`. This will provide a route for the app to transition to when canceling a request. It also will be the route when creating a new asset (Ex: `reports/a5bda8f0-d14c-11e8-be8f-711dd8cdcf5c/edit`).

This is a breaking change and apps using `navi-reports` and `navi-dashboards` will need to update their router.

TODO:
- [x] Rename routes

This PR needs #230


balajimanoharan:
Can we update the blueprints?

jkusa:
@balajimanoharan blueprints are already good to go:
- https://github.com/yahoo/navi/blob/master/packages/dashboards/blueprints/navi-dashboards/index.js#L75
- https://github.com/yahoo/navi/blob/master/packages/reports/blueprints/navi/index.js#L55

PR #239, opened 2018-10-16
Update: Upgrade node to 8 in travis build

mattgraham1995:
Cuts the build time by a few minutes for each package 

The difference:

image


balajimanoharan:
This comes with the upgrade. But as long as nothing breaks, i guess we could use this.

PR #241, opened 2018-10-18
Fix: Directory Sidebar spacing issue in Firefox

mattgraham1995:
Before:
image

After:
image

### DIFF
--- a/packages/directory/app/styles/navi-directory/components/dir-sidebar.less
+++ b/packages/directory/app/styles/navi-directory/components/dir-sidebar.less
@@ -16,7 +16,7 @@
 .dir-sidebar {
   color: @navi-blue-300;
   display: grid;
-  height: fit-content;
+  grid-template-rows: fit-content(50px);
 
   &__group {
     cursor: pointer;


packages/directory/app/styles/navi-directory/components/dir-sidebar.less
```diff
@@ -16,6 +16,7 @@
 .dir-sidebar {
   color: @navi-blue-300;
   display: grid;
+  grid-template-rows: fit-content(50px);
   height: fit-content;
```
balajimanoharan: Is this needed now?

packages/directory/app/styles/navi-directory/components/dir-sidebar.less
```diff
@@ -16,6 +16,7 @@
 .dir-sidebar {
   color: @navi-blue-300;
   display: grid;
+  grid-template-rows: fit-content(50px);
```
balajimanoharan: Do we have to update the px when we add elements to the sidebar?

mattgraham1995: No the `50px` refers to the height of one row in the grid.

PR #243, opened 2018-10-19
Update: Vertical collection in table component

daniellevinson:


packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -312,6 +353,177 @@ export default Component.extend({
...
+   * @private
+   * @returns {void}
+   */
+  _syncHeadersWidth() {
+    //in editing mode the real table headers are visible, no need to sync
+    if (get(this, 'isEditingMode')) {
+      return;
+    }
+
+    let widths = this.$('thead th')
```
cythrawll: So, I think we should move off of jquery. Community is trying to move away from it.  An alternative would be to use `this.element` which is native dom, and you can do querySelector/querySelectorAll off of it.

cythrawll:
So far so good, but the sizing is a little off when I did a report that only had two columns, the last column was cut off and I couldn't scroll.  (this was running the reports dummy app)
screen shot 2018-10-31 at 2 35 34 pm

daniellevinson:
Hmm. I'll take a look. Chrome right?

cythrawll:
Yup

jkusa:
@daniellevinson are you aware of the header flickering issue when scrolling in edit mode?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -273,24 +320,14 @@ export default Component.extend({
...
-
-    //Create a set of row dimensions for each row of data
-    let rowDimensions = new Array(get(this, 'tableData.length'));
-    for (let i = 0; i < rowDimensions.length; i++) {
-      rowDimensions[i] = rowDimension;
-    }
-
-    return rowDimensions;
-  }),
+  bufferSize: 0,
```
jkusa: Should we make this 10 or 20?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -312,6 +349,156 @@ export default Component.extend({
...
+        headerElm.style.width = `${widths[i]}px`;
+        widthChanged = true;
+      }
+    });
+
+    //set total width to headers wrapper div
+    let headerRowElm = this.element.querySelector('div.table-header-row'),
+      headerRowStyle = parseInt(headerRowElm.style.width);
+
+    if (!headerRowStyle || widthChanged) {
```
jkusa: what is the case that we need `parseInt`?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -312,6 +349,156 @@ export default Component.extend({
...
+    if (get(this, 'isEditingMode')) {
+      return;
+    }
+
+    let widths = Array.from(this.element.querySelectorAll('th')).map(elm => elm.getBoundingClientRect().width);
+
+    let prevWidths = get(this, 'columnsWidth'),
+      widthChanged = widths.length !== prevWidths.length; //if number of columns has changed, set to true to re-set width
+
+    Array.from(this.element.querySelectorAll('div.table-header-row .table-header-cell')).map((headerElm, i) => {
```
jkusa: do we need `div` in the seletor?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -312,6 +349,156 @@ export default Component.extend({
...
+      let headerElmWidth = parseInt(headerElm.style.width);
+
+      if (!headerElmWidth || widths[i] !== prevWidths[i]) {
+        headerElm.style.width = `${widths[i]}px`;
+        widthChanged = true;
+      }
+    });
+
+    //set total width to headers wrapper div
+    let headerRowElm = this.element.querySelector('div.table-header-row'),
```
jkusa: Could this be cached or does it change too often?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -312,6 +349,156 @@ export default Component.extend({
...
+
+  /**
+   * Synchronizes scroll position of table headers and table body
+   *
+   * @method _syncScroll
+   * @private
+   * @returns {void}
+   */
+  _syncScroll() {
+    let scrollableWrapper = this.element.querySelector(`.${get(this, 'tableWrapperClass')}`),
```
jkusa: Could this element query be cache as a cp?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -312,6 +349,156 @@ export default Component.extend({
...
+        elementToScroll = tableHeaderRow;
+      }
+
+      elementToScroll.scrollLeft = scrollLeft;
+    }
+
+    if (get(this, 'isEditingMode')) {
+      //reposition real headers for editing mode
+      let translate = `translate(0,${scrollableWrapper.scrollTop - 2}px)`;
+      Array.from(this.element.querySelectorAll('th')).map(headerElm => (headerElm.style.transform = translate));
```
jkusa: `forEach`

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: `200px` for example

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: the `thead > tr` has the same class

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: So I was wondering if caching selectors is a good practice in Ember like it is in React for example, because in no tutorial so far have I seen this. I'll try here

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: It was flickery when not all rows had the same height. I can play with it a little more

daniellevinson:
@jkusa re the flickering issue - yes I'm aware, tried to explain it in the PR summary - I 'm using the real headers for edit mode because of horizontal scrolling and overflow-y issues for pop ups, and re-positioning them is not very smooth. If this is a no-go let me see if I can try to think of a different approach here.

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: you mean the DOM element itself? it's removed from the DOM on edit mode actually

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: changed to 10

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -262,6 +262,27 @@ export default Component.extend({
...
+  isVerticalCollectionEnabled: computed(function() {
+    return featureFlag('enableVerticalCollectionTableIterator');
+  }),
+
+  /**
+   * @property {String} tableRenderer
+   */
+  tableRenderer: computed('isVerticalCollectionEnabled', function() {
+    return `table-renderer${get(this, 'isVerticalCollectionEnabled') ? '-vertical-collection' : ''}`;
+  }),
```
daniellevinson: make sure to test print

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,250 @@
...
+  /**
+   * during both render and re-render, sync headers width and scroll position
+   *
+   * @method didRender
+   * @override
+   */
+  didRender() {
+    this._super(...arguments);
+
+    run.scheduleOnce('render', () => {
```
daniellevinson: can we try without `scheduleOnce`

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,250 @@
...
+  },
+
+  /**
+   * Synchronizes the width of visible table headers with real (hidden) table header elements
+   *
+   * @method _syncHeadersWidth
+   * @private
+   * @returns {void}
+   */
+  _syncHeadersWidth() {
```
daniellevinson: see why it's not working

daniellevinson:
look into loader that is not rendering at 100% width

packages/core/addon/components/table-renderer-vertical-collection.js
daniellevinson: changed to `minWidth` here

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,250 @@
...
+      headerRowStyle = parseInt(headerRowElm.style.minWidth);
+
+    if (!headerRowStyle || widthChanged) {
+      set(this, 'columnsWidth', widths);
+      headerRowElm.style.minWidth = `${this.element.querySelector('table').getBoundingClientRect().width}px`;
+    }
+  },
+
+  /**
+   * attach scroll and view resize even listeneres when first rendered
```
mattgraham1995: typos

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,250 @@
...
+      elm.removeEventListener(SCROLL_EVENT, () => this._syncScroll())
+    );
+
+    //turn off view resize listener
+    get(this, 'resize').off(RESIZE_EVENT);
+
+    this._super(...arguments);
+  },
+
+  actions: {
```
mattgraham1995: Could we just use the `mut` helper to change the value of `isDragged` instead of using these actions?

packages/core/addon/components/table-renderer-vertical-collection.js
daniellevinson: seems to be working fine without, will be removed

packages/core/addon/components/table-iterator.js
```diff
@@ -4,15 +4,20 @@
  *
  * Usage:
  * {{#table-iterator
```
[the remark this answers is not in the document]
daniellevinson: yeah, or just yielding the data if `occlusion=false`

packages/core/addon/components/table-renderer-vertical-collection.js
daniellevinson: Sure, wasn't familiar with it

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: still having an issue here with a vertical scroller appearing on the pdf output

daniellevinson:
@mattgraham1995 I've addressed your comments, thanks
still there's an issue with a vertical scroller when exporting to pdf -(

packages/core/addon/components/navi-visualizations/table.js
cythrawll: Can we see a sample of the pdf?

cythrawll: Okay I just made a pdf and it looks fine. (no better or worse than we have now) if you look at the print/ route for table now, it has a scrollbar too, it's probably fine.

daniellevinson: so that's the thing, with what we have now (feature flag is off) the route has a scrollbar but the pdf doesn't (try it with a report with ~500 rows)
however if you turn the flag on, the pdf does have a scrollbar, trying to see where's the difference

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -313,6 +326,13 @@ export default Component.extend({
   },
 
   actions: {
+    /**
+     * @action onUpdateReport
+     */
+    onUpdateReport() {
+      this.attrs.onUpdateReport(...arguments);
```
jkusa: Can we just pass `onUpdateReport` prop to the template?

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -340,12 +360,14 @@ export default Component.extend({
      * @action updateColumnDisplayName
      */
     updateColumnDisplayName(column, displayName) {
-      this.attrs.onUpdateReport(
-        'updateColumn',
-        assign({}, column, {
-          displayName
-        })
-      );
+      run.scheduleOnce('afterRender', () => {
```
jkusa: What changed here? Double update in a render loop?

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,228 @@
...
+
+  /**
+   * @property {String} tableWrapperClass - class of table wrapper div
+   */
+  tableWrapperClass: 'table-widget__table-wrapper',
+
+  /**
+   * @property {Object} tableWrapperDomElement - table wrapper div DOM element
+   */
+  tableWrapperDomElement: computed('tableWrapperClass', function() {
```
jkusa: `tableWrapperClass` is private correct? Do we expected it to update? If not, can we remove the listener?

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,228 @@
...
+
+  /**
+   * @property {String} tableHeadersClass - class of outer div of visible table headers
+   */
+  tableHeadersClass: 'table-widget__table-headers',
+
+  /**
+   * @property {Object} tableHeadersDomElement - outer div of visible table headers DOM element
+   */
+  tableHeadersDomElement: computed('tableHeadersClass', function() {
```
jkusa: Same comment as above

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,228 @@
...
+    if (get(this, 'isEditingMode')) {
+      return;
+    }
+
+    let widths = Array.from(this.element.querySelectorAll('th')).map(elm => elm.getBoundingClientRect().width);
+
+    let prevWidths = get(this, 'columnsWidth'),
+      widthChanged = widths.length !== prevWidths.length; //if number of columns has changed, set to true to re-set width
+
+    Array.from(this.element.querySelectorAll('div.table-header-row-vc .table-header-cell')).map((headerElm, i) => {
```
jkusa: Can this be:
```js
this.element.querySelectorAll('div.table-header-row-vc .table-header-cell')).forEach((headerElm, i) => ...
```

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -0,0 +1,228 @@
...
+      let headerElmWidth = parseInt(headerElm.style.minWidth);
+
+      if (!headerElmWidth || widths[i] !== prevWidths[i]) {
+        headerElm.style.minWidth = `${widths[i]}px`;
+        widthChanged = true;
```
jkusa: Did you consider a trandtional for loop so you can do a `break` here?

packages/core/addon/components/table-renderer-vertical-collection.js
jkusa: Do we need to do this if the above is true?

packages/core/addon/templates/components/navi-visualizations/table.hbs
```diff
@@ -4,93 +4,23 @@
...
+  columns=columns
+  occlusion=occlusion
+  estimateHeight=estimateHeight
+  bufferSize=bufferSize
+  isEditingMode=isEditingMode
+  cellRendererPrefix=cellRendererPrefix
+  headerClicked=(action 'headerClicked')
+  updateColumnOrder=(action 'updateColumnOrder')
+  updateColumnDisplayName=(action 'updateColumnDisplayName')
+  onUpdateReport=(action 'onUpdateReport')
```
jkusa: Can this be:
```hbs
onUpdateReport=(action onUpdateReport)
```

packages/core/tests/acceptance/table-test.js
```diff
@@ -37,7 +45,7 @@ test('visiting /table', function(assert) {
 
   andThen(() => {
     assert.deepEqual(
-      find('.table-header-cell__title')
+      find('div.table-header-row-vc .table-header-cell__title')
```
jkusa: Do we need to be this specific?

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1964,7 +1977,7 @@ test('adding metrics to reordered table keeps order', function(assert) {
 
   andThen(() => {
     assert.deepEqual(
-      find('.table-header-cell__title')
+      find('div.table-header-row-vc .table-header-cell__title')
```
jkusa: Do we need to be this specific for selectors in this file?

packages/core/addon/templates/components/navi-visualizations/table.hbs
daniellevinson: I had tests failing in the dashboards package because we don't pass this action to the table component there
I've mentioned this bug on slack
once we fix that then sure, didn't want it to be part of this PR

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: That's how I did it initially, see my comment about the bug in dashboards

packages/core/addon/components/navi-visualizations/table.js
daniellevinson: For some reason with the vertical collection tests were failing with `You modified "columns" twice in a single render.`
Scheduling it in a separate render loop seemed to solve the issue. Did you have another suggestion?
I'll admit I have to read more about the run loop and understand how it works

packages/core/addon/components/table-renderer-vertical-collection.js
daniellevinson: We don't, by listener you mean the `'tableWrapperClass'` part?

packages/core/tests/acceptance/table-test.js
daniellevinson: Yes, you have the same class inside `tr.table-header-row-vc th`, we want to verify the visible one

packages/reports/tests/acceptance/navi-report-test.js
daniellevinson: Same answer as above

packages/core/addon/templates/components/navi-visualizations/table.hbs
jkusa:
```js
onUpdateReport=(action optional (onUpdateReport))
```

PR #246, opened 2018-10-25
Fix: Smartkey dimension bulk-import fix

cythrawll:
When we have a dimension with a smartkey, we need people to do bulk import of searchable ids.  The bulk importer needs to search against the right field.  This change makes sure that bulk-import works for smart key dimensions by looking to see if they have an 'id' tagged field and searching against that.  Otherwise, we search against primaryKey (searching against primaryKey was previous behavior).

Also found an issue which putting ids that were comma separated with spaces was sending bad requests to fili.  The splitting has been improved.

### DIFF
--- a/packages/data/tests/unit/services/bard-dimensions-test.js
+++ b/packages/data/tests/unit/services/bard-dimensions-test.js
@@ -341,8 +341,8 @@ module('Unit | Service | Dimensions', function(hooks) {
     });
   });
 
-  test('searchValueField: contains search', function(assert) {
-    assert.expect(6);
+  test('searchValueField: contains search', async function(assert) {
+    assert.expect(7);
 
     let response3 = {
       rows: [{ id: 'v1', desc: 'value1' }, { id: 'v2', desc: 'value2' }],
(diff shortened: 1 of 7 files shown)


packages/reports/addon/components/dimension-bulk-import.js
```diff
@@ -77,6 +77,12 @@ export default Ember.Component.extend({
    */
   _dimensionService: Ember.inject.service('bard-dimensions'),
 
+  /**
+   * @private
+   * @property {Ember.Service} bardMetadata
+   */
+  _bardMetadata: Ember.inject.service('bard-metadata'),
```
mattgraham1995: Could we import `service` using the new import styles?

packages/reports/addon/components/dimension-bulk-import.js
```diff
@@ -111,6 +120,19 @@ export default Ember.Component.extend({
...
+   * Gets the searchable id field if dimension has an 'id' tag, primaryKey otherwise
+   * @private
+   * @method _searchableIdField
+   * @returns {String} - Field to search against
+   */
+  _searchableIdField() {
+    const meta = get(this, '_bardMetadata').getById('dimension', get(this, 'dimension.name'));
+    //see if this field has a tagged id field
+    const field = meta.getFieldsForTag('id')[0] || {};
+    return get(field, 'name') || get(meta, 'primaryKeyFieldName');
```
jkusa: Did you consider adding a metadata computed property called `idFieldName` that would do:
```js
return this.getFieldsForTag('id')[0] || thi.get('primaryKeyFieldName')
```
That way we don't have to repeat this logic elsewhere in the future

cythrawll: I did consider that, however, was concerned about blowing up with unnecessary computed properties for every tag that we may support in the future.  Since the look up was a one liner, I went with this direction for now.

jkusa: My concern is this logic isn't something that `dimension-bulk-import` should have to know about. If we are going to support a tag as a first class citizen, we should formally support it in the metadata model.

cythrawll: Fixed

packages/data/tests/unit/services/bard-dimensions-test.js
```diff
@@ -341,51 +341,59 @@ module('Unit | Service | Dimensions', function(hooks) {
...
+        let [, field, operator, values] = req.queryParams.filters.match(QUERY_PARAM_REGEX),
+          rows = A(response3.rows).filterBy(field, values);
 
-      assert.equal(operator, 'contains', 'Search is done using `contains`');
+        assert.equal(operator, 'contains', 'Search is done using `contains`');
 
-      return [200, { 'Content-Type': 'application/json' }, JSON.stringify({ rows })];
+        return [200, { 'Content-Type': 'application/json' }, JSON.stringify({ rows })];
+      } catch (e) {
+        return [400, { 'Content-Type': 'text/plain' }, 'Invalid filter'];
```
jkusa: Should we do something like:
```js
assert.notOk(true, '`searchValueField` provided an invalid filter');
```

packages/reports/addon/components/dimension-bulk-import.js
```diff
@@ -111,6 +120,17 @@ export default Ember.Component.extend({
     set(this, '_loadingPromise', DS.PromiseObject.create({ promise }));
   },
 
+  /**
+   * Gets the searchable id field if dimension has an 'id' tag, primaryKey otherwise
+   * @private
+   * @method _searchableIdField
+   * @returns {String} - Field to search against
+   */
+  _searchableIdField() {
```
jkusa: Can this be a CP?

PR #251, opened 2018-10-29
Fix: size of loading view in Firefox

kevinhinterlong:
Firefox before/after:
 

Chrome:

---

There's still a small issue with the horizontal scroll bar in Firefox, but I'm not sure what's causing it

### DIFF
--- a/packages/core/app/styles/navi-core/components/navi-loader.less
+++ b/packages/core/app/styles/navi-core/components/navi-loader.less
@@ -12,13 +12,7 @@
   }
 
   &__spinner {
-    align-items: center;
-    background-color: @navi-container-gray;
-    display: flex;
-    flex: 1;
-    height: 300px;
-    justify-content: center;
-    width: 100%;
+    .spinner;
   }
 
   &__bounce {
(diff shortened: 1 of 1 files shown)


packages/core/app/styles/navi-core/components/navi-loader.less
```diff
@@ -15,7 +15,7 @@
     align-items: center;
     background-color: @navi-container-gray;
     display: flex;
-    flex: 1;
+    flex: 1 1 auto;
```
mattgraham1995: We could just do `flex: auto` since this seems to be equivalent to that. 
https://css-tricks.com/almanac/properties/f/flex/

kevinhinterlong: I think `&__spinner` just be the same as 

https://github.com/yahoo/navi/blob/a62677653c9168e22258d0981a58f9f9cba10354/packages/core/app/styles/navi-core/common/spinner.less#L6-L13

(the results are the same)

Is there any way to directly reference that, or should I just make the attributes for `&__spinner` the same as `.spinner`

mattgraham1995: You can just say 
`&__spinner {`
`.spinner;`
`}`

mattgraham1995:
https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint

kevinhinterlong:
No more horizontal scroll bar

PR #253, opened 2018-11-01
Update: Add date-dimension filter builder for dimensions of datatype date

mattgraham1995:
Uses the navi-date-picker to allow selection of date/interval values for dimensions with date values. 

The supported operators are Since, Before, and Between.

screen shot 2018-11-01 at 6 40 30 pm

screen shot 2018-11-02 at 11 24 40 am

screen shot 2018-11-02 at 11 25 11 am

PR #255, opened 2018-11-02
Update: Add date range filter value for dimensions with date values

mattgraham1995:
Part 2 of the split of PR #253 

Similar to the date-time interval selector. I add a computed property to the navi-date-picker component to accept either a string or interval object as its interval attribute. This way, we can set the filter value to be a string representing the interval and have the component update accordingly.


packages/reports/addon/components/filter-values/date-range.js
```diff
@@ -26,9 +27,19 @@ export default Ember.Component.extend({
...
+        let intervalStr = interval.asStrings('YYYY-MM-DD');
+
+        this.attrs.onUpdateFilter({
+          values: [`${get(intervalStr, 'start')}/${get(intervalStr, 'end')}`]
+        });
+      } else {
+        //Set the date range for the datetime filter
+        this.attrs.onUpdateFilter({
+          interval
+        });
```
jkusa: Can we normalize the contract here and change shape in the action handler?

packages/reports/addon/components/navi-date-range-picker.js
```diff
@@ -44,7 +44,23 @@ export default Ember.Component.extend({
...
    * @property {Object} customRange - custom range interval, or default interval
    */
-  customRange: computed.or('interval', 'defaultInterval'),
+  customRange: computed.or('intervalInstance', 'defaultInterval'),
+
+  /**
+   * @property {Object} intervalInstance - Interval Object that is either passed in as an attribute or built from a string
+   */
+  intervalInstance: computed('interval', function() {
+    let interval = get(this, 'interval');
```
jkusa: Can we normalize to a interval class and change from string when we pass the prop down?

packages/core/tests/unit/helpers/as-interval-test.js
```diff
@@ -0,0 +1,16 @@
...
+
+test('interval accepts interval or string', function(assert) {
+  assert.expect(2);
+  const expectedInterval = Interval.parseFromStrings('2018-10-31', '2018-11-01');
+
+  assert.ok(
+    expectedInterval.isEqual(asInterval(['2018-10-31/2018-11-01'])),
+    'Interval string is turned into an interval class instance'
+  );
+  assert.ok(expectedInterval.isEqual(asInterval([expectedInterval])), 'Interval class instance is passed through');
```
jkusa: Can we add error cases? `null`, `undefined`, numbers, object, bad string?

PR #265, opened 2018-12-10
Fix: Vertical Collection header scroll gesture

cythrawll:
Scrolling via the header via trackpad gesture or vertical scroll wheel didn't work.

This fixes that as both seem to be capturable with the 'wheel' event.  best of all this event is fired even when the container isn't scrollable! So we take the data from the wheel event and fire a scroll event to the table body.

### DIFF
--- a/packages/core/addon/components/table-renderer-vertical-collection.js
+++ b/packages/core/addon/components/table-renderer-vertical-collection.js
@@ -34,6 +34,11 @@ const SCROLL_EVENT = 'scroll';
  */
 const RESIZE_EVENT = 'didResize';
 
+/**
+ * @constant {String} WHEEL_EVENT - event for mouse and trackpad gestures
+ */
+const WHEEL_EVENT = 'wheel';
+
 export default Component.extend({
   layout,
 
@@ -93,6 +98,8 @@ export default Component.extend({
     [get(this, 'tableWrapperDomElement'), get(this, 'tableHeadersDomElement')].forEach(elm =>
       elm.addEventListener(SCROLL_EVENT, () => this._syncScroll())
     );
+
+    get(this, 'tableHeadersDomElement').addEventListener(WHEEL_EVENT, e => this._headerWheelSync(e));
   },
 
   /**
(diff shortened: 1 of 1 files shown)


packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -178,6 +185,21 @@ export default Component.extend({
...
+   * Event handler for wheel events in header to scroll the table body vertically.
+   * Works for mouse wheels and vertical scrolling via mousepad gestures.
+   * @param {Event} event - wheel event
+   * @private
+   * @returns {Boolean} false
+   */
+  _headerWheelSync(event) {
+    let table = get(this, 'tableWrapperDomElement');
+
+    table.scrollTo(table.scrollLeft + event.deltaX, table.scrollTop);
```
daniellevinson: we can use `scrollLeft` here but otherwise looks good

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -178,6 +185,21 @@ export default Component.extend({
...
+    event.preventDefault(); // Prevents the page navigation gesture in Mac OSX
```
mattgraham1995: Do we need both `event.preventDefault()` and `return false`?

cythrawll: Nope, just me remembering old fashioned event interface, bad habit. Removed.

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -178,6 +185,20 @@ export default Component.extend({
     }
   },
 
+  /**
+   * Event handler for wheel events in header to scroll the table body vertically.
+   * Works for mouse wheels and vertical scrolling via mousepad gestures.
+   * @param {Event} event - wheel event
+   * @private
+   * @returns {Boolean} false
```
mattgraham1995: [One more thing](https://66.media.tumblr.com/04319def23cf65fdb5fbb183632b9ba0/tumblr_mve9fsMfbg1qc5y70o1_500.gif)

PR #266, opened 2018-12-18
Update: Add header line wrap limit

cythrawll:
* Truncates and removes limit on hover.

Long column names in headers on small columns (think metrics with a huge trend description) were wrapping A LOT.  causing the header to take up a lot of vertical space leaving little for row data.
This css fix will truncate wrapping after 3 lines and show the whole header name on hover.

firefox doesn't support the `line-clamp` so it needed the @supports work around.

headers


mattgraham1995:
Could you post a screenshot?

packages/core/app/styles/navi-core/visualizations/table.less
```diff
@@ -326,3 +342,13 @@
     font-size: @font-size-xsmall;
   }
 }
+
+@supports not (-moz-line-clamp: @navi-table-max-wrap) {
```
jkusa: `-webkit-line-clamp`

PR #280, opened 2019-01-08
Fix: Build timing out in reports and core

mattgraham1995:
We should statically determine what the display month is for a week rather than depend on the date of runtime. 

It seems that ember-sortable 1.12.4 released about a week ago and it was causing some of our tests to time out.

### DIFF
--- a/packages/reports/package.json
+++ b/packages/reports/package.json
@@ -82,7 +82,7 @@
     "ember-qunit-assert-helpers": "^0.2.0",
     "ember-resize": "0.0.17",
     "ember-resolver": "^4.0.0",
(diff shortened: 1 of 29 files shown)


packages/reports/package.json
```diff
@@ -82,7 +82,7 @@
     "ember-qunit-assert-helpers": "^0.2.0",
     "ember-resize": "0.0.17",
     "ember-resolver": "^4.0.0",
-    "ember-sortable": "^1.12.1",
+    "ember-sortable": "1.12.3",
```
Alex-Aralis: This package is in more than just reports and core.  Should it be pinned in all the subpackages as well?

PR #282, opened 2019-01-08
Update: Build table visualization model using default show clause dimension fields

daniellevinson:


jkusa:
@daniellevinson can we get a screenshot?

packages/core/addon/components/cell-renderers/dimension.js
```diff
@@ -58,12 +72,17 @@ export default Component.extend({
...
-    let descField = get(this, 'descField'),
+  value: computed('dimensionField', 'descField', 'idField', 'data', function() {
+    let dimensionField = get(this, 'dimensionField'),
+      descField = get(this, 'descField'),
       idField = get(this, 'idField'),
       data = get(this, 'data');
 
-    if (data[descField]) {
+    if (dimensionField) {
+      if (data[dimensionField]) {
```
jkusa: what if the dim is `=== 0`?

packages/core/addon/components/cell-renderers/dimension.js
```diff
@@ -58,12 +72,17 @@ export default Component.extend({
   /**
    * @property {String} value - value that should be displayed in table cell
    */
-  value: computed('descField', 'idField', 'data', function() {
-    let descField = get(this, 'descField'),
+  value: computed('dimensionField', 'descField', 'idField', 'data', function() {
+    let dimensionField = get(this, 'dimensionField'),
+      descField = get(this, 'descField'),
       idField = get(this, 'idField'),
       data = get(this, 'data');
```
jkusa: Do we need to fetch these values if we don't use them?

packages/core/addon/models/table.js
```diff
@@ -77,22 +78,63 @@ export default VisualizationBase.extend(Validations, {
...
+/**
+ * returns default dimension fields (show clause)
+ * @param {Object} dimension - dimension from request
+ * @returns {Object} - list of field names
+ */
+function getDefaultDimensionFields(dimension) {
+  let fields = get(dimension, 'dimension.fields');
+
+  if (isArray(fields)) {
+    return fields.filter(field => isArray(field.tags) && field.tags.includes('show')).map(field => field.name);
```
jkusa: Is `dimension` a dimension metadata object? Can we use `getFieldsForTag('show')` here?

https://github.com/yahoo/navi/blob/master/packages/data/addon/models/metadata/dimension.js#L82

packages/core/addon/models/table.js
```diff
@@ -77,22 +78,63 @@ export default VisualizationBase.extend(Validations, {
...
+}
+
+/**
+ * builds a single dimension column
+ * @param {Object} dimension - dimension from request
+ * @param {Object} column
+ * @param {String} fieldName
+ * @returns {Object} - dimension column
+ */
+function dimensionColumn(dimension, column, fieldName) {
```
jkusa: should we call this `buildDimensionColumn`?

packages/data/addon/utils/dimension.js
```diff
@@ -0,0 +1,23 @@
...
+
+export function canonicalizeDimension(dimensionName, fieldName) {
+  if (fieldName) {
+    return `${dimensionName}(${fieldName})`;
+  }
+
+  return dimensionName;
+}
+
+export function dimensionFormat(dimensionName, fieldName) {
```
jkusa: should we call this `formatDimensionName`?

packages/core/addon/models/table.js
```diff
@@ -77,22 +78,63 @@ export default VisualizationBase.extend(Validations, {
...
+      dimensionName = get(dimension, 'dimension.name');
+
+    if (!defaultFields.length) {
+      let column = columnIndex[dimensionName];
+      return dimensionColumn(dimension, column);
+    } else {
+      //return an array of columns for each dimension field
+      return defaultFields.map(fieldName => {
+        let column = columnIndex[canonicalizeDimension(dimensionName, fieldName)];
+        return merge({}, { field: { show: fieldName } }, dimensionColumn(dimension, column, fieldName));
```
jkusa: should this logic be in `dimensionColumn`?

jkusa: I am not sure about the property `show`. `show` sounds like a boolean. `show` in fili is a metadata entity and a mechinism for fetching dimension `field`s. We want to model and describe the entity itself not the mechinism. 

You wouldn't say `Give me the device with the id and description shows`. Rather you would say `Give me the device with the id and description fields`.

Could we make the shape something like:
```js
{ 
   type: 'dimension',
   dimension: dimensionObject,
   field: 'id',
   displayName: 'Device (name)'
}

File: packages/core/addon/components/cell-renderers/dimension.js (hunk 1 above)
Comment (daniellevinson) [reply packages/core/addon/components/cell-renderers/dimension.js]: how is it different from the old logic (`if (data[descField])...` or `else if (data[idField])...)`?

File: packages/core/addon/components/cell-renderers/dimension.js (hunk 2 above)
Comment (daniellevinson) [reply packages/core/addon/components/cell-renderers/dimension.js]: you mean fetch the rest only if `dimensionField` is null?

File: packages/core/addon/models/table.js (hunk 4 above)
Comment (daniellevinson) [reply packages/core/addon/models/table.js]: yeah tried to avoid having `buildDimensionColumns` and `buildDimensionColumn` but I also prefer `buildDimensionColumns`

File: packages/core/addon/models/table.js (hunk 6 above)
Comment (daniellevinson) [reply packages/core/addon/models/table.js]: sounds better

Comment (daniellevinson) [reply packages/core/addon/models/table.js]: I agree, also didn't see a reason for `field: {dimension: ... }}`

Comment (daniellevinson) [reply packages/core/addon/models/table.js]: `dimensionObject` is actually the dimension id, do we need an object here?

Comment (daniellevinson) [reply packages/core/addon/models/table.js]: or maybe
```
{
  type: 'dimension or metric',
  [dimension or metric]: {
    name: 'advanceTrafficCurrency',
    parameters: {} // for metric
    field: 'id', // for dimension - or should it be at the root level?
  }
  format: '' // for metric
  displayName: 'Traffic Currency'
}
```

daniellevinson: @jkusa @cythrawll @mattgraham1995 @balajimanoharan feel free to comment

cythrawll: Yeah I wouldn't mind switching the shape so that we can use `type`, and `dimension`. freeing up `field` for this. it would be a lot cleaner.    I think I favor @daniellevinson last comment.

daniellevinson: @jkusa I like it. In any case it might be a breakable change

packages/core/addon/components/cell-renderers/dimension.js
daniellevinson: how is it different from `if (data[descField])` or `if (data[idField])`?

packages/core/addon/models/table.js
daniellevinson: @jkusa @cythrawll I've temporarily changed `show` to `field` which makes more semantical sense, so for now for dimension we have ```{
  type: 'dimension',
  field: {
    dimension: 'advanceTrafficCurrency',
    field: 'id'
  }
  displayName: 'Platform (id)'
}```
the more I'm diving into the shape changing the more I'm starting to think maybe we should separate it to its own PR, as this one might get a little messy since we rely on the `field` key for all column types, existing reports have `format` at the root level, change calls to `canonicalizeMetric`, update tests etc.

daniellevinson:
screenshot: 
screen shot 2019-01-17 at 4 46 52 pm

packages/core/addon/models/table.js
cythrawll: I agree on probably making it a separate PR just for the breaking change. as you're right, it might get hairy. 

I'm fine having the subfield as `field` for the short term.

packages/core/addon/components/cell-renderers/dimension.js
jkusa: Can we use `isEmpty`?

PR #286, opened 2019-01-16
Update: Favorites filter should link to my-data route then add filter

mattgraham1995:
The favorites filter in the directory route should specifically be applied to `my-data`

### DIFF
--- a/packages/directory/addon/templates/components/dir-sidebar.hbs
+++ b/packages/directory/addon/templates/components/dir-sidebar.hbs
@@ -14,7 +14,7 @@
 </section>
 <section class='dir-sidebar__filters'>
   {{#each filters as | filter |}}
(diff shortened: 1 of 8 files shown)


packages/directory/addon/templates/components/dir-sidebar.hbs
```diff
@@ -14,7 +14,7 @@
 </section>
 <section class='dir-sidebar__filters'>
   {{#each filters as | filter |}}
-    {{#link-to (query-params filter=filter.queryParams.filter sortBy=filter.queryParams.sortBy)
+    {{#link-to 'directory.my-data' (query-params filter=filter.queryParams.filter sortBy=filter.queryParams.sortBy)
```
jkusa: Do we have any tests around this?

PR #292, opened 2019-01-18
Update: Remove all `this.sendActions`s

Alex-Aralis:
- `this.sendAction('myAction', myValue) ` ==> `this.get('myAction')(myValue)`
- renamed actions with `on`
- explicitly pass action handlers

### DIFF
--- a/packages/core/addon/components/pick-container.js
+++ b/packages/core/addon/components/pick-container.js
@@ -20,8 +20,7 @@
  */
 import Ember from 'ember';
 import Layout from '../templates/components/pick-container';
-
-const { get, computed } = Ember;
+import { get, computed } from '@ember/object';
 
 export default Ember.Component.extend({
   layout: Layout,
--- a/packages/core/addon/components/pick-container.js
+++ b/packages/core/addon/components/pick-container.js
@@ -1,9 +1,17 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
+ * `pick-container` no longer sends up onUpdteSelection or onFormToggled actions.  Instead they
+ * must be handled explicetly via passing handlers.
+ *
  * Usage:
- *   {{#pick-container selection=selectedValue as |selection actions|}}
+ *   {{#pick-container
+ *      selection=selectedValue
+ *      onUpdateSelection=(action handleUpdateSelection)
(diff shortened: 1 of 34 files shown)


packages/core/addon/components/pick-object-container.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
```
balajimanoharan: Can we revert the changes in this file?

packages/core/addon/components/pick-container.js
```diff
@@ -151,14 +161,15 @@ export default Ember.Component.extend({
 
   /**
    * @method toggleForm - toggles pick-form between open and closed states
-   * Sends `formToggled` action with state to parents
+   * Calls onFormToggled handler.
    */
   toggleForm() {
     this.toggleProperty('isFormOpen');
     this._updateFormVisibility();
+    const handleFormToggled = get(this, 'onFormToggled');
```
balajimanoharan: Can we do this instead to avoid the `if` conditional below?
```
const handleFormToggled = get(this, 'onFormToggled') || (() => {});
```

packages/core/addon/components/pick-container.js
```diff
@@ -13,6 +21,8 @@
 import Ember from 'ember';
 import Layout from '../templates/components/pick-container';
 
+const { get, computed } = Ember;
```
jkusa: `import { computed, get } from '@ember/object';`

packages/reports/addon/templates/components/navi-date-range-picker.hbs
```diff
@@ -1,5 +1,5 @@
-{{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
-{{#pick-container updateSelection=this.attrs.setInterval as |selection container|}}
+{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{#pick-container onUpdateSelection=this.attrs.onSetInterval as |selection container|}}
```
jkusa: Since we are touching this, can we remove all the `this.attrs`? It is now a private API

packages/core/addon/components/pick-container.js
Alex-Aralis: Odd, i don't even remember adding this import o_O

mattgraham1995:
Is this ready to be merged?

PR #301, opened 2019-01-31
Upgrade and Rework Mirage

mattgraham1995:
This PR upgrades mirage to 0.4.x. The mirage route handlers have been reworked to handle the polymorphism of the delivery rule `deliveredItem` relationship as well as separating the routehandlers for reports and users.


packages/core/app/mirage/routes/dashboard.js
```diff
@@ -1,29 +1,15 @@
...
 export default function() {
-  this.get('dashboards/:id', ({ dashboards, dashboardWidgets, deliveryRules }, request) => {
+  this.get('dashboards/:id', ({ dashboards }, request) => {
     let id = request.params.id,
       dashboard = dashboards.find(id);
 
-    assignWidgets(dashboard, dashboardWidgets);
-    assignDeliveryRules(dashboard, deliveryRules);
+    if (!dashboard) {
+      return new Mirage.Response(RESPONSE_CODES.NOT_FOUND, {}, { errors: [`Unknown identifier '${id}'`] });
```
jkusa: Can we import:
```js
import Response from 'ember-cli-mirage/response';
```

packages/core/app/mirage/routes/dashboard.js
```diff
@@ -59,25 +50,22 @@ export default function() {
...
-    });
+    author.update(dashboards, author.dashboards.add(dashboard));
 
     // Init properties
-    db.dashboards.update(dashboard.id, {
+    dashboard.update({
       widgetIds: [],
-      createdOn: moment.utc().format(TIMESTAMP_FORMAT),
-      updatedOn: moment.utc().format(TIMESTAMP_FORMAT)
+      createdOn: null,
```
jkusa: what is the reason for this change?

packages/core/app/mirage/routes/delivery-rules.js
```diff
@@ -67,28 +55,34 @@ export default function() {
...
+      deliveredItems = { reports, dashboards },
+      deliveredItem = deliveredItems[`${deliveredItemType}s`].find(deliveryRule.deliveredItemId.id),
+      user = users.find(deliveryRule.ownerId);
 
-    // Delete deliveryRule from user
-    db.users.update(deliveryRule.ownerId, {
-      deliveryRules: owner.deliveryRules.filter(id => id.toString() !== deliveryRule.id)
-    });
+    if (!deliveryRule) {
+      return new Mirage.Response(RESPONSE_CODES.NOT_FOUND, {}, { errors: [`Unknown identifier '${id}'`] });
```
jkusa: Same import comment as before

packages/core/app/mirage/routes/dashboard.js
mattgraham1995: screen shot 2019-02-04 at 3 19 06 pm

Some tests were comparing to the returned dashboard model and expecting null values. But now that I look into the reports route, it has the same test but returns the timestamp. I'm not sure why the test is getting null values for `createdOn` and `updatedOn` for reports but not dashboards, so I'll look into that further.

jkusa: That was probably done b/c we do a full object compare and would not know the expected dates in the test.

PR #303, opened 2019-02-04
Move webservice from Maven to Gradle

### LINKED ISSUE #294: Move webservice build from mvn to gradle
We have a multi module product we need to build

1) models -> this contains the beans, permissions etc that people can use in their own servlet setups with elide.
2) app -> standalone demo app that uses ^^ models.

I started out building this with maven because that's what we use internally, but this would be better served using gradle build system.

kevinhinterlong:
resolves #294


cythrawll:
So trying this out, problem I have is gradle test seems to fail. it seems it can't find the hibernate.cfg

```
Caused by: org.hibernate.internal.util.config.ConfigurationException: Specified cfg.xml file [/private/tmp/navi/packages/webservice/app/../settings/hibernate.cfg.xml] does not exist
```

PR #306, opened 2019-02-06
Breaking: Directory title should toggle based on chosen directory

mattgraham1995:
The directories override has been moved to a service so that Navi apps only need to override them in one place. 
screen shot 2019-02-06 at 4 48 26 pm
screen shot 2019-02-06 at 4 48 29 pm


packages/directory/addon/controllers/directory.js
```diff
@@ -4,9 +4,23 @@
...
+import { getOwner } from '@ember/application';
+import { inject as service } from '@ember/service';
+import { A as arr } from '@ember/array';
 
 export default Controller.extend({
+  /**
+   * @property {Ember.Controller} application
+   */
+  application: computed(function() {
+    return getOwner(this).lookup('controller:application');
```
cythrawll: does `get(this, 'target.currentPath')` work if you're only fetching the application to get the current path?

mattgraham1995: Yes it does. I didn't know about that property

packages/directory/addon/controllers/directory.js
```diff
@@ -51,10 +57,14 @@ export default Controller.extend({
   /**
    * @property {String} title - Title for the table
    */
-  title: computed('filter', function() {
-    let title = dirInfo[0].name,
+  title: computed('filter', 'target.currentPath', function() {
+    const currentPath = get(this, 'target.currentPath'),
```
jkusa: Can we use this: 
https://www.emberjs.com/api/ember/3.7/classes/RouterService/properties/currentRoute?anchor=currentRouteName

I think it is a little clearer.

cythrawll: why inject a service when the property exists in controllers already?  That's kind of like knocking on your neighbors door so he can tell you what color your own kitchen walls are.

packages/directory/addon/services/directories.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright 2019, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import Service from '@ember/service';
```
balajimanoharan: Is this more of an util rather than a service?

packages/directory/addon/controllers/directory.js
mattgraham1995: If clarity is the issue, I can add a comment to that line to make it more clear.

packages/directory/addon/services/directories.js
mattgraham1995: It is mostly a util, but the static file of directories that it depends on is harder to override in an app without overriding the entire util. As a service, the app just needs to extend this service with their own directories.

packages/directory/addon/controllers/directory.js
jkusa: To me, `target` is one of those obscure legacy concepts that would be nice to be cleaned up. 

Just look at the doc definition:

> The object to which actions from the view should be sent.

Nothing indicates "I have knowledge of the current route".

When reading the code the router service feels like a more obvious way to get the current route. It also has better documentation:

https://www.emberjs.com/api/ember/3.7/classes/RouterService/properties/currentRoute?anchor=currentRouteName

Also, the router service is a more portable technique. If we move this logic to component unlike `target` the router service can still be used.

>That's kind of like knocking on your neighbors door so he can tell you what color your own kitchen walls are.

Keep in mind that the target is set (injected) into the controller as well.

cythrawll: Is there indication that this is legacy as in going to be deprecated?

The doc for `target` also says: 
> by default, the value of the target property is set to the router

Which does give indication that target knows about the route.

No argument on better documentation.

> Also, the router service is a more portable technique. If we move this logic to component unlike `target` the router service can still be used.

agreed it's more portable, but it's odd that it's injecting a service to get info from the router, when the router is already available in the object.  It's completely redundant.  Router service is nice to have for components and classes that don't already have access to the router. But why inject it when the object already has the router?

packages/directory/addon/services/directories.js
balajimanoharan: Should we just move the enums under here then as a default for directories? Keep it closer to this service?

mattgraham1995: Okay I've moved the default directories into the service so now the utils file isn't needed

PR #310, opened 2019-02-08
Breaking: Table visualization model change (fixes #300)

### LINKED ISSUE #300: Refactor dimension and metric representation in table visualization model
Currently for dimension and metric we have:
```
{
  type: 'dimension',
  field: {
    dimension: 'dimensionName',
    field: 'desc' // <- this was added for show clause
  },
  displayName: 'Display Name'
}
```
```
{
  type: 'metric',
  field: {
    metric: 'metricName',
    parameters: {...}
  },
  displayName: 'Display Name',
  format: '...'
}
```

We would like to have something more like
```
{
  type: 'dimension',
  dimension: {
    name: 'advanceTrafficCurrency',
    field: 'id'
  },
  displayName: 'Traffic Currency'
}
```
and
```
{
  type: 'metric',
  metric: {
    name: 'pageViews',
    format: '0.0a'
  },
  displayName: 'Traffic Currency'
}
```

daniellevinson:
This changes the columns representations to:
```
{
  type: 'dimension',
  attributes: {
    name: 'advanceTrafficCurrency',
    field: 'id'
  },
  displayName: 'Traffic Currency'
}
```

```
{
  type: 'metric',
  attributes: {
    name: 'pageViews',
    format: '0.0a',
    parameters: {...}
  },
  displayName: 'Page Views'
}
```

```
{
  type: 'dateTime',
  attributes: {
    name: 'dateTime'
  },
  displayName: 'Date'
}
```

### DIFF
--- a/packages/core/addon/components/cell-renderers/dimension.js
+++ b/packages/core/addon/components/cell-renderers/dimension.js
@@ -12,6 +12,7 @@
 
 import Component from '@ember/component';
 import { computed, get } from '@ember/object';
+import { readOnly } from '@ember/object/computed';
 import layout from '../../templates/components/cell-renderers/dimension';
 import { isEmpty } from '@ember/utils';
 
@@ -24,28 +25,30 @@ export default Component.extend({
   classNames: ['table-cell-content', 'dimension'],
 
   /**
+   * @property {String} name - dimension name
+   */
+  name: readOnly('column.attributes.name'),
+
+  /**
    * @property {String} descField - field to find the description
    */
-  descField: computed('column.field.dimension', function() {
-    return get(this, 'column.field.dimension') + '|desc';
+  descField: computed('name', function() {
+    return get(this, 'name') + '|desc';
   }),
 
   /**
    * @property {String} idField - field to find id
    */
-  idField: computed('column.field.dimension', function() {
-    return get(this, 'column.field.dimension') + '|id';
+  idField: computed('name', function() {
+    return get(this, 'name') + '|id';
   }),
 
   /**
    * @property {Object} dimensionField - field to find dimension field
    */
-  dimensionField: computed('column.field.dimension', 'column.field.field', function() {
-    if (!get(this, 'column.field.field')) {
-      return null;
-    }
-
-    return `${get(this, 'column.field.dimension')}|${get(this, 'column.field.field')}`;
+  dimensionField: computed('name', 'column.attributes.field', function() {
+    let field = get(this, 'column.attributes.field');
+    return field ? `${get(this, 'name')}|${field}` : null;
   }),
 
   /**
--- a/packages/core/addon/components/cell-renderers/metric.js
+++ b/packages/core/addon/components/cell-renderers/metric.js
@@ -15,7 +15,7 @@ import layout from '../../templates/components/cell-renderers/metric';
 import { computed, get } from '@ember/object';
 import { oneWay } from '@ember/object/computed';
 import { isEmpty } from '@ember/utils';
-import { canonicalizeMetric } from 'navi-data/utils/metric';
+import { canonicalizeColumnAttributes } from 'navi-data/utils/metric';
 import { smartFormatNumber } from 'navi-core/helpers/smart-format-number';
 import numeral from 'numeral';
 
@@ -28,22 +28,17 @@ export default Ember.Component.extend({
   classNames: ['table-cell-content', 'metric'],
 
   /**
-   * @property {String} metric - metric name used to fetch the value for
+   * @property {Object} attributes - metric name and optional parameters used to fetch the value for
    */
-  metric: oneWay('column.field'),
-
-  /**
-   * @property {String} format - format the number should be rendered
-   */
-  format: oneWay('column.format'),
+  attributes: oneWay('column.attributes'),
 
   /**
    * @property {Number} - value to be rendered on the cell
    */
-  metricValue: computed('data', 'metric', 'format', function() {
-    let format = get(this, 'format'),
+  metricValue: computed('data', 'attributes', function() {
+    let format = get(this, 'attributes.format'),
       type = get(this, 'column.type'),
-      canonicalName = canonicalizeMetric(get(this, 'metric')),
+      canonicalName = canonicalizeColumnAttributes(get(this, 'attributes')),
       metricValue = get(this, `data.${canonicalName}`);
 
     if (isEmpty(metricValue)) {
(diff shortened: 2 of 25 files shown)


packages/core/addon/components/cell-renderers/dimension.js
```diff
@@ -23,29 +23,38 @@ export default Component.extend({
    */
   classNames: ['table-cell-content', 'dimension'],
 
+  /**
+   * @property {String} name - dimension name
+   */
+  name: computed('column.attributes.name', function() {
```
mattgraham1995: We could just use `computed.oneWay` or `computed.readOnly` here.

packages/core/addon/components/cell-renderers/dimension.js
```diff
@@ -23,29 +23,38 @@ export default Component.extend({
...
 
   /**
    * @property {Object} dimensionField - field to find dimension field
    */
-  dimensionField: computed('column.field.dimension', 'column.field.field', function() {
-    if (!get(this, 'column.field.field')) {
+  dimensionField: computed('name', 'column.attributes.field', function() {
+    let field = get(this, 'column.attributes.field');
+
+    if (!field) {
```
mattgraham1995: Could we simplify this to ``return field ? `${get(this, 'name')}|${field}` : null;`` ?

packages/core/addon/serializers/table.js
```diff
@@ -5,14 +5,35 @@
 import VisualizationSerializer from 'navi-core/serializers/visualization';
 import { parseMetricName } from 'navi-data/utils/metric';
 import { get } from '@ember/object';
+import { isPresent, isNone } from '@ember/utils';
+import omit from 'lodash/omit';
 
-const TRANSFORMS = {
+const MAP_STRING_FIELD = {
   metric: column => Object.assign(column, { field: parseMetricName(column.field) }),
```
mattgraham1995: Do these functions have the right column shape?

packages/core/addon/serializers/table.js
```diff
@@ -23,12 +44,18 @@ export default VisualizationSerializer.extend({
...
+          mapped = column;
+
         if (typeof get(column, 'field') === 'string') {
-          let type = get(column, 'type');
+          mapped = MAP_STRING_FIELD[type](column);
+        }
 
-          return TRANSFORMS[type](column);
+        if (isPresent(get(mapped, 'field'))) {
+          mapped = mapFieldToAttributes(type === 'threshold' ? 'metric' : type, mapped);
```
mattgraham1995: Do we need to use `mapFieldToAttributes`? I feel like we should just change the functions within `MAP_STRING_FIELD` to return the right shape.

packages/data/tests/unit/utils/metric-test.js
```diff
@@ -147,4 +148,47 @@ module('Unit - Utils - Metrics Utils', function() {
       'Parser throws an error for corrupted metric'
     );
   });
+
+  test('map column attributes to object', function(assert) {
+    assert.expect(6);
+
+    assert.deepEqual(
+      mapColumnAttributes({ name: 'base' }),
+      { metric: 'base' },
```
mattgraham1995: What do you think of always including a parameters property to the returned metric object? I think it would be good to have a consistent shape for the metric object whether it has parameters or not.

packages/core/addon/serializers/table.js
daniellevinson: the input may be either one of the two previous "versions":
`field: 'revenue(currency=USD)'` or 
`field: { metric: 'revenue', parameters: { currency: 'USD' } }`

packages/core/addon/serializers/table.js
daniellevinson: Not against it, thought the change itself might be more clear if I just added another mapping on top of what was in place

packages/core/addon/serializers/table.js
mattgraham1995: I was thinking of the output of each function. We're still outputting an object with `field` instead of `attributes: { name: ... `

daniellevinson: Yes it will be confusing, will change

packages/core/addon/components/cell-renderers/dimension.js
```diff
@@ -23,29 +23,33 @@ export default Component.extend({
    */
   classNames: ['table-cell-content', 'dimension'],
 
+  /**
+   * @property {String} name - dimension name
+   */
+  name: computed.readOnly('column.attributes.name', function() {
```
balajimanoharan:
```js
import  { readOnly } from '@ember/object/computed';
...
name: readOnly('column.attributes.name'),
...
```

packages/data/addon/utils/metric.js
```diff
@@ -122,3 +122,22 @@ export function parseMetricName(canonicalName) {
...
   };
 }
+
+/**
+ * Returns a metric object given column attributes
+ * @function mapColumnAttributes
+ * @param {Object} attributes - column attributes
+ * @param {String} attributes.name - metric name
+ * @param {Object} attributes.parameters - metric parameters
+ * @returns {Object} - metric object with metric (name) and parameters
```
balajimanoharan: Should we just say `object with metric name and params` here? metric object made me think of the metric metadata object or the fragment.

packages/core/addon/components/cell-renderers/metric.js
```diff
@@ -28,22 +28,17 @@ export default Ember.Component.extend({
...
   /**
    * @property {Number} - value to be rendered on the cell
    */
-  metricValue: computed('data', 'metric', 'format', function() {
-    let format = get(this, 'format'),
+  metricValue: computed('data', 'attributes', function() {
+    let format = get(this, 'attributes.format'),
       type = get(this, 'column.type'),
-      canonicalName = canonicalizeMetric(get(this, 'metric')),
+      canonicalName = canonicalizeMetric(mapColumnAttributes(get(this, 'attributes'))),
```
balajimanoharan: We do this quite a lot, makes me wonder if we should just create a method that would return a canonicalName given the column attributes.

daniellevinson: Yeah it would make sense having one

packages/core/addon/components/cell-renderers/dimension.js
```diff
@@ -26,7 +27,7 @@ export default Component.extend({
   /**
    * @property {String} name - dimension name
    */
-  name: computed.readOnly('column.attributes.name', function() {
+  name: readOnly('column.attributes.name', function() {
```
jkusa: ` name: readOnly('column.attributes.name')`

PR #311, opened 2019-02-08
Fix: Serialization tests.

cythrawll:
Tests were failing when run in isolation as models were not
setup right in the container.  Creating a record in the store
fixes the static properties issue and the right propeties
get populated in the model class.


cythrawll:
to reproduce in core run this:

```
ember t --server --filter 'Unit | Serializer'
```

packages/core/tests/unit/serializers/line-chart-test.js
```diff
@@ -12,12 +11,14 @@ moduleFor('serializer:line-chart', 'Unit | Serializer | line chart', {
...
     'validator:request-dimension-order',
-    'validator:length'
+    'validator:length',
+    'validator:request-filters'
   ],
   beforeEach() {
     setupMock();
     Serializer = this.subject();
-    Model = getOwner(this).factoryFor('model:line-chart').class;
+    const store = this.container.lookup('service:store');
```
jkusa: Can we do ` getOwner(this).lookup('service:store')`?

PR #320, opened 2019-02-18
Update: Remove Flag that shouldn't be published

mattgraham1995:
We have `isDevelopingAddon` set to true in only dashboards, reports, and data. These shouldn't be true when we publish our addons. With Ember 3, this flag will cause Navi's template files to be run through the template linter for apps that use Navi.

More info:
https://github.com/ember-template-lint/ember-template-lint/issues/119#issuecomment-258620041

### DIFF
--- a/packages/dashboards/index.js
+++ b/packages/dashboards/index.js
@@ -10,9 +10,5 @@ module.exports = {
     var target = parentAddon || app;
 
     target.import('vendor/loader.css');
-  },
-
-  isDevelopingAddon: function() {
-    return true;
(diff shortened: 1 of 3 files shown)


daniellevinson:
this was set to true for live reload?

mattgraham1995:
@daniellevinson I'm not sure why it was set to true. Only reports, dashboards, and data had it set to true though. Live reload works fine on core and others so I don't think this flag is necessary and at the very least it shouldn't be true in the published version of navi.

PR #321, opened 2019-02-20
Breaking: Make navi more compatible with Ember 3.x apps

mattgraham1995:
The dynamic segments of the routes should be in snake case for Ember 3. 

Also there was an issue with setting the report state. The [issue](https://github.com/emberjs/ember.js/issues/17545) is that the contract for the `transition.send` function changed. The first argument in the function definition is `ignoreFailure` which used to be defined as a boolean. So I think since we were calling `transition.send` and passing a string as the first argument, the compiler would know that the string wasn't meant to be assigned to `ignoreFailure`. There was a change recently that no longer defines `ignoreFailure` as a boolean, but gives it a default value of `false`. Now that means that when we pass a string as the first argument, that string gets passed as `ignoreFailure` to `transition.send` and everything blows up. The issue I mentioned says that they solved the issue by bumping up the version of `router_js`, but the [line](https://github.com/tildeio/router.js/blob/2d611a583b74b8f8d8290146cf2816f175e251e3/lib/router/transition.ts#L323) still has the default value and not a boolean specified like it did before. So for now, we will pass `false` as the first parameter to `transition.send`.

### DIFF
--- a/packages/app/app/router.js
+++ b/packages/app/app/router.js
@@ -17,20 +17,20 @@ Router.map(function() {
     this.route('my-data');
   });
 
-  this.route('dashboardCollections', function() {
-    this.route('collection', { path: '/:collectionId' });
+  this.route('dashboard-collections', function() {
+    this.route('collection', { path: '/:collection_id' });
   });
 
   this.route('dashboards', function() {
     this.route('new');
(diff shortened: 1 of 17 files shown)


jkusa:
Can we break this into multiple commits?

mattgraham1995:
@jkusa Done

PR #328, opened 2019-02-22
App Ember 3.7 upgrade

### LINKED ISSUE #308: Upgrade App to ~3.7+

cythrawll:
Resolves #308

## Description

Upgrade App to ember 3.7! incrementally upgrade and run codemods, and get everything buttoned up and running.

## Proposed Changes

Run ember update and codemods.


packages/app/.travis.yml
```diff
@@ -0,0 +1,27 @@
+---
```
jkusa: Can we remove this?

packages/app/config/environment.js
```diff
@@ -59,11 +58,12 @@ module.exports = function(environment) {
     ENV.APP.LOG_VIEW_LOOKUPS = false;
 
     ENV.APP.rootElement = '#ember-testing';
+    ENV.APP.autoboot = false;
   }
 
   if (environment === 'production') {
     ENV['ember-cli-mirage'] = {
-      enabled: false //set to true for gh-pages
+      enabled: true //set to true for gh-pages
```
jkusa: Is this change intentional?

packages/app/tests/acceptance/custom-reports-test.js
```diff
@@ -1,57 +1,47 @@
...
-  assert.expect(3);
+  test('Viewing saved reports', async function(assert) {
+    assert.expect(3);
 
-  visit('/reports');
-  andThen(() => {
+    await visit('/reports');
     assert.ok(
-      Ember.isPresent(find('.navi-reports-index .navi-collection table')),
+      isPresent(find('.navi-reports-index .navi-collection table')),
```
jkusa: Could this be:
```js
assert.dom('.navi-reports-index .navi-collection table').exists();
```

packages/app/tests/acceptance/custom-reports-test.js
```diff
@@ -1,57 +1,47 @@
...
+    await visit('/reports');
+    await click(link('New Report'));
+    assert.ok(
+      currentURL().match(
+        /^\/reports\/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\/edit/
+      ),
+      'Clicking "New Report" button brings the user to the report builder'
+    );
+
+    assert.ok(isPresent(find('.report-builder')), 'Custom report builder is visible');
```
jkusa: same here

packages/app/config/environment.js
cythrawll: see comment, wondering why it was ever set to false.

packages/app/.travis.yml
cythrawll: the entire file?

packages/app/tests/helpers/contains-helpers.js
```diff
@@ -0,0 +1,29 @@
+import { findAll } from '@ember/test-helpers';
```
jkusa: Should we put this in core so other packages can use this?

packages/app/config/environment.js
jkusa: For the `gh-pages` demo we need to enable the mock for `prod` builds manually. However, as a user of the Navi app, when you clone the repo and configure your web services, you would want to disable mirage for prod builds.

packages/app/.travis.yml
jkusa: Yes, we use the top level .travis.yml file.

cythrawll: we didn't remove navi-data one.  Also, do we really want to remove this on every ember-upgrade?

packages/app/app/services/navi-notifications.js
```diff
@@ -34,7 +34,7 @@ export default Ember.Service.extend({
    * @returns {Ember.Service} notification service
    */
   add(options = {}) {
-    let notificationService = get(this, 'notificationService');
+    let notificationService = this.notificationService;
```
mattgraham1995: Why did this change?

packages/app/app/services/navi-notifications.js
```diff
@@ -56,6 +56,6 @@ export default Ember.Service.extend({
    * @returns {Ember.Service} navi notification service
    */
   clearMessages() {
-    return get(this, 'notificationService').clearMessages();
+    return this.notificationService.clearMessages();
```
mattgraham1995: Same here

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -0,0 +1,29 @@
+import { findAll } from '@ember/test-helpers';
+
+/**
+ * Finds selector by innerText
+ * @param {String} selector - selector to search
+ * @param {String} text - text to search by
+ * @returns {Element} - first matching dom element
+ */
+export default function findByContains(selector, text) {
+  return findAll(selector).filter(el => el.innerText.includes(text))[0];
```
mattgraham1995: We could just use `find` instead of filter and then we wouldn't need the `[0]` at the end

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Finds button by innnerText
+ * @param {String} text - text to search for
+ * @returns {Element} - first matching dom element
+ */
+export function button(text) {
```
mattgraham1995: Maybe `button` and `link` would be better with more descriptive names. They make sense in the scope of this file, but anywhere it's imported to wouldn't necessarily have the context to tell what these functions do.

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: Also what does this return if the element isn't found? Would an ` || null;` at the end be necessary?

packages/app/app/services/navi-notifications.js
cythrawll: gets are no longer needed.

packages/app/app/services/navi-notifications.js
cythrawll: same

packages/core/test-support/helpers/contains-helpers.js
cythrawll: we need all the dom elements so the filter works.

also it would return undefined which is appropriate I think.

mattgraham1995: But we're only returning the first one that matches the filter which is exactly what find does
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

mattgraham1995: `return findAll(selector).find(el => el.innerText.includes(text));`

PR #335, opened 2019-03-07
Multi select dimension fix

cythrawll:
## Description
Found bug with smart keys and mutli-select box in dimension filter selector.   When you select a dimension with the same id (but different primary Identifier) all dimensions with that id would be considered selected.

## Proposed Changes
Use the configured identifier field instead of id when comparing bard-dimension


packages/data/tests/unit/models/bard-dimension-test.js
```diff
@@ -14,4 +15,22 @@ module('Unit | Model | Bard Dimension', function() {
       'toString correctly provides a description of the dimension model factory'
     );
   });
+
+  test('isEqual', function(assert) {
+    assert.expect(3);
```
jkusa: 2

PR #337, opened 2019-03-08
Upgrade: Directory to Ember v3.7

### LINKED ISSUE #192: Update Directory to the latest ember version
Update Directory to the latest ember version

daniellevinson:
Resolves #192

## Description
Upgrade navi-directory to Ember 3.7


packages/directory/index.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
-  name: 'navi-directory'
+  name: require('./package').name
```
cythrawll: what's going on here, did a codemod do this?  Does this really resolve to the same thing?

cythrawll: nm I see that it would end up being the same thing.

packages/directory/.travis.yml
```diff
@@ -0,0 +1,59 @@
...
+    - env: EMBER_TRY_SCENARIO=ember-canary
+    - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
+
+before_install:
+  - npm config set spin false
+  - npm install -g npm@4
+  - npm --version
+
+script:
+  - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO
```
Alex-Aralis: This file is not used.

packages/directory/addon/components/dir-asset-row-actions.js
```diff
@@ -24,9 +24,9 @@ export default Component.extend({
    * @property {String} type - the type of the asset
    */
   type: computed('value', function() {
-    let value = get(this, 'value');
+    let { value } = this;
```
Alex-Aralis: In the past destructures have failed for me when the 3.x addons have been used with 2.x apps.  Does this work with zig or teller?

packages/directory/addon/mixins/search-filter.js
```diff
@@ -19,8 +19,6 @@ export default Mixin.create({
    */
   searchResults: computed('directory.q', 'sortedItems', function() {
     let queryString = get(this, 'directory.q');
-    return get(this, 'sortedItems').then(
-      items => (isEmpty(queryString) ? items : searchRecords(items, queryString, 'title'))
-    );
+    return this.sortedItems.then(items => (isEmpty(queryString) ? items : searchRecords(items, queryString, 'title')));
```
Alex-Aralis: `(isEmpty(queryString))` => `isEmpty(queryString)`

packages/directory/config/ember-try.js
```diff
@@ -0,0 +1,79 @@
...
+          npm: {
+            devDependencies: {
+              '@ember/jquery': '^0.5.1'
+            }
+          }
+        }
+      ]
+    };
+  });
+};
```
Alex-Aralis: File not used.

packages/directory/package.json
```diff
@@ -6,87 +6,95 @@
...
   "devDependencies": {
-    "broccoli-asset-rev": "^2.4.5",
-    "ember-ajax": "^3.0.0",
+    "@ember/jquery": "^0.6.0",
+    "@ember/optional-features": "^0.6.3",
+    "broccoli-asset-rev": "^2.7.0",
     "ember-basic-dropdown": "1.0.1",
     "ember-c3": "0.2.3",
-    "ember-cli": "^3.1.4",
+    "ember-cli": "~3.7.1",
```
Alex-Aralis: can we go up to 3.8?  nothing much changes.

packages/directory/addon/components/dir-asset-row-actions.js
daniellevinson: teller is already upgraded but I'll check with zig

packages/directory/config/ember-try.js
daniellevinson: do we need the `ember-try` dependency then? a standard addon has a `"test:all": "ember try:each"` npm script

packages/directory/addon/components/dir-asset-row-actions.js
mattgraham1995: Ziggurat doesn't use directory anyway

daniellevinson: I've probably changed to this syntax in a few other places

packages/directory/config/ember-try.js
Alex-Aralis: hmm, not sure.  The tests are being run by lerna via the top level package.json.  I don't think we are currently using ember-try in our tests.

cythrawll: I kind of think we should keep these things in place even if we don't use them.  Ember upgrades will just put them back, and nott worth the effort to remember to remove them. let's reduce what needs to be done with each upgrade.

packages/directory/addon/components/dir-asset-row-actions.js
daniellevinson: oh wait I forgot it's only a directory upgrade :laughing:

packages/directory/config/ember-try.js
Alex-Aralis: I tried it with core and it works fine.  Reverting for now.

packages/directory/addon/components/dir-asset-row-actions.js
daniellevinson: Teller's looking good so far

packages/directory/addon/mixins/search-filter.js
daniellevinson: It's a `prettier` thing

packages/directory/config/ember-try.js
daniellevinson: I agree with @cythrawll that it makes sense to keep these

packages/directory/tests/helpers/start-app.js
```diff
@@ -1,12 +1,12 @@
...
 
 export default function startApp(attrs) {
-  let attributes = merge({}, config.APP);
+  let attributes = assign({}, config.APP);
   attributes.autoboot = true;
-  attributes = merge(attributes, attrs); // use defaults, but you can override;
+  attributes = assign(attributes, attrs); // use defaults, but you can override;
 
   return run(() => {
     let application = Application.create(attributes);
```
balajimanoharan: Is this file used? This used to be the old way to run acceptance tests.

packages/directory/.travis.yml
jkusa: Can we add this to the gitignore?

packages/directory/config/ember-try.js
jkusa: In general, if we don't want a generated file we should add an entry for it in the gitignore file. 

In this case, this file can be helpful for future upgrades in determining supported Ember versions.

packages/directory/tests/helpers/start-app.js
daniellevinson: hmm don't think so, tests still pass when removing this one and destroy-app.js

PR #339, opened 2019-03-14
Fix: Reinclude bower in the npmignore.

Alex-Aralis:
## Description
the `bower.json` file was not pushed to NPM because it was listed in the `.npmignore` file.  This breaks the node-core install.

## Proposed Changes
- Remove `bower.json` and `.bowerrc` from `.npmignore`.

### DIFF
--- a/packages/core/.npmignore
+++ b/packages/core/.npmignore
@@ -6,7 +6,7 @@
 /bower_components/
 
 # misc
-/**/.gitkeep
+/.bowerrc
 /.editorconfig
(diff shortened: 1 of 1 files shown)


packages/core/.npmignore
```diff
@@ -6,8 +6,7 @@
 /bower_components/
 
 # misc
-**/.gitkeep
-/.bowerrc
```
daniellevinson: /.bowerrc can actually stay

PR #342, opened 2019-03-20
Fix: Missed one.

cythrawll:
Resolves #

## Description
Found an issue where dates weren't being formatted correctly in widgets

## Proposed Changes
Same as yesterday's bug fix, just missed one.


mattgraham1995:
Truly inspiring work.

PR #345, opened 2019-03-26
New: Add test helper that selects on elements based on content using jquery notation

mattgraham1995:
## Description
Right now, findByContains is not very helpful for complicated selectors with multiple `:contains` calls in its jquery equivalent.

## Proposed Changes
- Add `findContains` to support using jquery style contains selectors without using jquery
- Eventually replace `findByContains` implementation with this one

### DIFF
--- a/packages/core/test-support/helpers/contains-helpers.js
+++ b/packages/core/test-support/helpers/contains-helpers.js
@@ -27,3 +27,62 @@ export function buttonContains(text) {
 export function linkContains(text) {
   return findByContains('a', text);
 }
+
+/**
+ * This function allows us to keep using the jquery notation for :contains without using jquery
+ *
+ * Example:
+ * Jquery: $(.class-thing:contains(blah) .metric .test:contains(other))
+ *
+ * findByContains:
+ * [...findByContains('.class-thing', 'blah').querySelectorAll('.metric .test')].find(el => el.innerText.trim().includes('other'))
+ *
+ * Non-Jquery Equivalent: [
+ *    ...[
+ *      ...document.querySelectorAll(
+ *        (.class-thing)
+ *      )].find(el => el.innerText.trim().includes('blah'))
+ *    .querySelectorAll('.metric .test')
+ *  ]
+ *  .find(el => el.innerText.trim().includes('other'))
+ *
+ * findContains: findContains(.class-thing:contains(blah) .metric .test:contains(other))
+ *
+ * @param {String} selector - jquery or native style selector
+ * @param {HTMLElement} [baseElement] - optional parameter to use as base element to select off of
+ * @returns {HTMLElement} - Element based on text it (or its parent) contains
+ */
+export function findContains(selector, baseElement) {
+  let nativeSelector,
+    parentElement = baseElement || document;
+
+  if (selector.includes(':contains(')) {
+    let splits = selector.split(':contains(');
+    let firstSelector = splits.shift(); //Selector without the :contains operator
+    let restOfSelector = splits.join(':contains('); //We only want to split on the first instance of :contains(, so join the rest back together if there were more
+    let endOfContentIndex = findEndOfContentIndex(restOfSelector);
+    let content = restOfSelector.substring(0, endOfContentIndex);
+    let nextSelector = restOfSelector.substring(endOfContentIndex + 1).trim();
+    let filteredElement = [...parentElement.querySelectorAll(firstSelector)].find(el =>
+      el.innerText.trim().includes(content)
+    );
+
+    /**
+     * If the :contains was used at the very end of the selector, return the element we found here
+     * Else recursively find element from rest of selector using the found element as the base
+     */
+    return nextSelector ? findContains(nextSelector, filteredElement) : filteredElement;
+  } else {
+    nativeSelector = selector;
+  }
+
+  return parentElement.querySelector(nativeSelector);
+}
+
+/**
+ * @param {String} str - String with the content ended with a ) and possibly another selector following that
+ * @returns {Number} Index of the ) character that closes the :contains( expression that precedes the input string
+ */
+function findEndOfContentIndex(str) {
+  return str.includes(':contains(') ? str.lastIndexOf(')', str.indexOf(':contains')) : str.lastIndexOf(')');
+}
--- a/packages/core/test-support/helpers/contains-helpers.js
+++ b/packages/core/test-support/helpers/contains-helpers.js
@@ -53,22 +53,46 @@ export function linkContains(text) {
  * @returns {HTMLElement} - Element based on text it (or its parent) contains
  */
 export function findContains(selector, baseElement) {
-  let parentElement = baseElement || document;
+  return findAllContains(selector, baseElement)[0];
+}
 
-  if (selector.includes(':contains(')) {
-    let regex = /^(.*?):contains\((["']?)(.*?)\2\)(.*)$/;
+/**
+ * Same as findContains but returns ALL elements that match rather than just the first one
+ *
+ * @param {String} selector
+ * @param {HTMLElement} baseElement
+ * @returns {HTMLElement[]} - Array of all elements that match the selector and text content
+ */
+export function findAllContains(selector, baseElement) {
+  let parentElement = baseElement || document;
+  let hasContains = /:[cC]ontains\(/;
+
+  if (hasContains.exec(selector)) {
+    let regex = /^(.*?):[cC]ontains\((["']?)(.*?)\2\)(.*)$/;
     let matches = regex.exec(selector);
     let firstSelector = matches[1];
     let content = matches[3];
     let nextSelector = matches[4];
-    let filteredElement = [...parentElement.querySelectorAll(firstSelector)].find(el => el.innerText.includes(content));
+    let filteredElements = [...parentElement.querySelectorAll(firstSelector)].filter(el =>
+      el.innerText.includes(content)
+    );
 
     /**
      * If the :contains was used at the very end of the selector, return the element we found here
      * Else recursively find element from rest of selector using the found element as the base
      */
-    return nextSelector ? findContains(nextSelector, filteredElement) : filteredElement;
+    return nextSelector
+      ? flattenDeep(filteredElements.map(el => findAllContains(nextSelector, el)))
+      : flattenDeep(filteredElements);
   }
 
-  return parentElement.querySelector(selector);
+  return [...parentElement.querySelectorAll(selector)];
+}
+
+/**
+ * Deep flatten array
+ * @param {Array} arr - Array to flatten
+ */
+function flattenDeep(arr) {
+  return arr.reduce((acc, val) => (Array.isArray(val) ? acc.concat(flattenDeep(val)) : acc.concat(val)), []);
 }
--- a/packages/core/test-support/helpers/contains-helpers.js
+++ b/packages/core/test-support/helpers/contains-helpers.js
@@ -31,11 +31,11 @@ export function linkContains(text) {
 /**
  * This function allows us to keep using the jquery notation for :contains without using jquery
  *
+ * Need quotes around content with parentheses -- e.g. class:contains("(content)")
+ *
+ *
  * Example:
  * Jquery: $(.class-thing:contains(blah) .metric .test:contains(other))
- *
- * findByContains:
- * [...findByContains('.class-thing', 'blah').querySelectorAll('.metric .test')].find(el => el.innerText.trim().includes('other'))
  *
  * Non-Jquery Equivalent: [
  *    ...[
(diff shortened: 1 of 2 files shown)


packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,62 @@ export function buttonContains(text) {
...
+  }
+
+  return parentElement.querySelector(nativeSelector);
+}
+
+/**
+ * @param {String} str - String with the content ended with a ) and possibly another selector following that
+ * @returns {Number} Index of the ) character that closes the :contains( expression that precedes the input string
+ */
+function findEndOfContentIndex(str) {
```
jkusa: Can we add a few tests?

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,62 @@ export function buttonContains(text) {
...
 }
+
+/**
+ * This function allows us to keep using the jquery notation for :contains without using jquery
+ *
+ * Example:
+ * Jquery: $(.class-thing:contains(blah) .metric .test:contains(other))
+ *
+ * findByContains:
+ * [...findByContains('.class-thing', 'blah').querySelectorAll('.metric .test')].find(el => el.innerText.trim().includes('other'))
```
Alex-Aralis: This documentation is out of date.

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,62 @@ export function buttonContains(text) {
...
+  if (selector.includes(':contains(')) {
+    let splits = selector.split(':contains(');
+    let firstSelector = splits.shift(); //Selector without the :contains operator
+    let restOfSelector = splits.join(':contains('); //We only want to split on the first instance of :contains(, so join the rest back together if there were more
+    let endOfContentIndex = findEndOfContentIndex(restOfSelector);
+    let content = restOfSelector.substring(0, endOfContentIndex);
+    let nextSelector = restOfSelector.substring(endOfContentIndex + 1).trim();
+    let filteredElement = [...parentElement.querySelectorAll(firstSelector)].find(el =>
+      el.innerText.trim().includes(content)
+    );
```
Alex-Aralis: Why not use findByContains here?

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,62 @@ export function buttonContains(text) {
...
+
+    /**
+     * If the :contains was used at the very end of the selector, return the element we found here
+     * Else recursively find element from rest of selector using the found element as the base
+     */
+    return nextSelector ? findContains(nextSelector, filteredElement) : filteredElement;
+  } else {
+    nativeSelector = selector;
+  }
```
Alex-Aralis: else not needed because true block returns directly

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,62 @@ export function buttonContains(text) {
...
+ */
+export function findContains(selector, baseElement) {
+  let nativeSelector,
+    parentElement = baseElement || document;
+
+  if (selector.includes(':contains(')) {
+    let splits = selector.split(':contains(');
+    let firstSelector = splits.shift(); //Selector without the :contains operator
+    let restOfSelector = splits.join(':contains('); //We only want to split on the first instance of :contains(, so join the rest back together if there were more
+    let endOfContentIndex = findEndOfContentIndex(restOfSelector);
```
Alex-Aralis: `/^(.*?):contains\((.+?)\)(.*)$/`

Group 1: firstSelector
Group 2: content
Group 3: nextSelector

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: How so?

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: We're going to get rid of findByContains

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: Considered using a regex for this and decided against it. Can't remember why though

mattgraham1995: Now I remember, we select on some elements that have parentheses in them. The regex cuts off the content at the closing parenthesis. I figured it was safer to just find the content manually.

Alex-Aralis: The contains with parentheses are quoted right?

Alex-Aralis: `/^(.*?):contains\(\s*("(.*?)"|(.*?))\s*\)(.*)$/`

Group 1: firstSelector
Group 2:  Whatever is in the contains, trimmed
Group 3: If quoted, has text inside quote, otherwise undefined
Group 4: If not quoted, whatever is inside contains trimmed, otherwise undefined
Group 5: nextSelector

packages/core/test-support/helpers/contains-helpers.js
Alex-Aralis: findByContains is going away. Also, findByContains is not the function being documented.

packages/core/test-support/helpers/contains-helpers.js
Alex-Aralis: okay, here we go.

`/^(.*?):contains\(\s*("(.*?)"|(.*?))\s*\)((:contains\(\s*(".*?"|.*?)\s*\)|\[.+?(=".*?"|.*?)?\]|[^\s+>~,])*)(.*)$/`

```perl
/^

# Lazy match everything before contains
# Group 1: firstSelectorPart1
(.*?)

# Match :contains()
# Group 2: Whatever is in the contains, trimmed
# Group 3: If quoted, has text inside quote, otherwise undefined
# Group 4: If not quoted, whatever is inside contains trimmed, otherwise undefined
:contains\(\s*("(.*?)"|(.*?))\s*\)

# Match trailing conjunctive selectors
# Group 5: firstSelectorPart2
((:contains\(\s*(".*?"|.*?)\s*\)|\[.+?(=".*?"|.*?)?\]|[^\s+>~,])*)

# Match everything after first selectors group
# Group 9: nextSelector
(.*)

$/
```

Now `firstSelector = firstSelectorPart1 + firstSelectorPart2`

Even this does not handle `,` selector operators properly.

packages/core/test-support/helpers/contains-helpers.js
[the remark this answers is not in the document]
Alex-Aralis: I've provided a regex that handles the above case by taking `.someclass:contains(sometext).otherclass` -> firstSelector = `.someclass.otherclass` and content = `sometext`

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: The contains with parentheses are not quoted, no

packages/core/test-support/helpers/contains-helpers.js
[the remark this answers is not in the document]
mattgraham1995: That's a different selector though, isn't it? That would be equivalent to `.someclass .otherclass:contains(sometext)` instead of `.someclass:contains(sometext) .otherclass`

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: Your regex still doesn't grab the closing parentheses correctly and cuts the content short when there's a `)` in it. 
Screen Shot 2019-03-31 at 2 18 25 PM

Alex-Aralis: The idea is that parentheses need to be in quotes

packages/core/test-support/helpers/contains-helpers.js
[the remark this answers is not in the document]
Alex-Aralis: those are equivalent selectors.

packages/core/test-support/helpers/contains-helpers.js
Alex-Aralis: After all, there is no guaranty that the contains text will have balanced parens.

packages/core/test-support/helpers/contains-helpers.js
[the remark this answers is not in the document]
mattgraham1995: They are not equivalent. 
Screen Shot 2019-04-01 at 1 29 21 PM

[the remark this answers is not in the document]
Alex-Aralis: you put an extra descendants operator, but you are correct.

packages/core/test-support/helpers/contains-helpers.js
mattgraham1995: I have no issue with using quotes, but I don't think adding quotes actually solves the problem, but rather changes it to a different problem. How would escaping the quotation marks change what the regex is parsing? It would still see it as a quote in the same way as if it was unescaped. 

Quotation syntax is optional in jquery and only used partly in navi so it shouldn't be a requirement here, but again, it's not the syntax that I mind. I just don't think it actually solves the problem it is meant to.

Alex-Aralis: The regex captures both with and without quotes.

Alex-Aralis: After taking a closer look at the jquery :contains behavior `/^(.*?):contains\((\s*"(.*?)"\s*|(.*?))\)(.*)$/` is a closer fit for their behavior.

It captures things like `:contains(skdjf " sldjf) -> 'skdjf " sldjf'` as expected, but can also properly capture `:contains("sldfjsd") -> 'sldfjsd'` as expected.

Alex-Aralis: jQuery does support escaping characters.  I don't think we need to support that functionality at this time.  Especially when we have a way of representing double quotes `:contains(")` and parens `:contains("()")`.

Alex-Aralis: It seems like the only difference in behavior with the jQuery capturing would be the quote escaping.  In jQuery they allow `:contains("text \" othertext") -> 'text " othertext'`.  They also allow `:contains('text " othertext') -> 'text " othertext'`.

Alex-Aralis: `/^(.*?):contains\((\s*"(.*?)"\s*|\s*'(.*?)'\s*|(.*?))\)(.*)$/` would grab single quote contents. 

And finally `/^(.*?):contains\((\s*"((?:\\.|.)*?)"\s*|\s*'((?:\\.|.)*?)'\s*|((?:\\.|.)*?))\)(.*)$/` grabs escaped characters the same way that jquery does.  This final method will require resolving the escaped characters in a second step.

Alex-Aralis: Here's a more compact version.  I've removed the white space trimming before and after quoted arguments because that is the jQuery behavior.

```javascript
/^(.*?):contains\((["']?)((?:\\.|.)*?)\2\)(.*)$/
```

Group 1: `firstSelector`
Group 3: `content` (without quotes)
Group 4: `nextSelector`

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,48 @@ export function buttonContains(text) {
...
+ *  ]
+ *  .find(el => el.innerText.trim().includes('other'))
+ *
+ * findContains: findContains(.class-thing:contains(blah) .metric .test:contains(other))
+ *
+ * @param {String} selector - jquery or native style selector
+ * @param {HTMLElement} [baseElement] - optional parameter to use as base element to select off of
+ * @returns {HTMLElement} - Element based on text it (or its parent) contains
+ */
+export function findContains(selector, baseElement) {
```
Alex-Aralis: Problem: Consider the case
```html
<div class="a">dog</div>
<div class="b">dog</div>
```
and we do a `findContains(':contains(dog).b')`.  

We would expect to find the second division, but it is lost because the iteration is done on only the first find of the contains.

Solution: Do the recursion with a `findContainsAll` function that takes an array of selectors instead of a singleton, then make `findContains` a special case of that function.

packages/core/test-support/helpers/contains-helpers.js
```diff
@@ -27,3 +27,48 @@ export function buttonContains(text) {
...
+  let parentElement = baseElement || document;
+
+  if (selector.includes(':contains(')) {
+    let regex = /^(.*?):contains\((["']?)(.*?)\2\)(.*)$/;
```
Alex-Aralis: `/^(.*?):contains\((["']?)(.*?)\2\)(.*)$/i`

cythrawll:
Closing as we decided to use Sizzle

PR #346, opened 2019-03-27
Fix: Save As modal closes after success

cythrawll:
## Description
Save As modal no longer closes after successful save.

## Proposed Changes
Change modal state to controller, send actions to controller

### DIFF
--- a/packages/reports/addon/controllers/reports/report.js
+++ b/packages/reports/addon/controllers/reports/report.js
@@ -15,6 +15,11 @@ const REPORT_STATE = {
 
 export default Controller.extend({
   /**
+   * @property {String} showSaveAs - whether the save as dialog is showing
+   */
+  showSaveAs: false,
+
+  /**
    * @property {String} reportState - state of the the report
    */
   reportState: computed({
@@ -55,5 +60,14 @@ export default Controller.extend({
    */
   didReportFail: computed('reportState', function() {
     return get(this, 'reportState') === REPORT_STATE.FAILED;
-  })
+  }),
+
+  actions: {
+    /**
+     * Closes save as modal
+     */
+    closeSaveAs() {
+      this.set('showSaveAs', false);
+    }
+  }
 });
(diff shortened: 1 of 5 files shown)


packages/reports/addon/controllers/reports/report.js
```diff
@@ -14,6 +14,11 @@ const REPORT_STATE = {
 };
 
 export default Controller.extend({
+  /**
+   * @property {String} showSaveAs - whether the save as dialog is showing
```
daniellevinson: Boolean

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -404,6 +404,8 @@ test('Save As report', function(assert) {
       '(New Copy) Hyrule News',
       'New Saved Report is being viewed'
     );
+
+    assert.notOk(find('.save-as__save-as-modal-btn').is(':visible'), 'Save As Modal not visible after save');
```
mattgraham1995: Does this work? I thought `.is` was a jquery function

cythrawll: it is because we're still using jquery in tests atm and this is just an inverse of an earlier assertion.  let me see if I can update them both to non jquery version.

cythrawll: Ok pushed this so that upgrading will be easier.

packages/reports/addon/controllers/reports/report.js
```diff
@@ -14,6 +14,11 @@ const REPORT_STATE = {
 };
 
 export default Controller.extend({
+  /**
+   * @property {Boolean} showSaveAs - whether the save as dialog is showing
+   */
+  showSaveAs: false,
```
jkusa: Can we check that this state is reset when loading the view?

cythrawll: Done

PR #348, opened 2019-03-29
Update: Hide datetime column on all timegrain

mattgraham1995:
## Description
We recently decided that when the request has a timegrain of `all`, we will hide the dateTime column in the table visualization.

## Proposed Changes
- Change table model validations to check for the existence of the datetime column based on the timegrain
- Change `rebuildConfig` in table model to handle adding or removing the datetime column when the timegrain changes

### DIFF
--- a/packages/core/addon/models/table.js
+++ b/packages/core/addon/models/table.js
@@ -23,7 +23,11 @@ const Validations = buildValidations(
         let request = get(options, 'request');
         return request && hasAllColumns(request, arr(columns));
       },
-      dependentKeys: ['model._request.dimensions.[]', 'model._request.metrics.[]']
+      dependentKeys: [
+        'model._request.dimensions.[]',
+        'model._request.metrics.[]',
+        'model._request.logicalTable.timeGrain.name'
+      ]
     })
   },
   {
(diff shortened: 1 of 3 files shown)


packages/core/addon/models/table.js
```diff
@@ -54,19 +58,21 @@ export default VisualizationBase.extend(Validations, {
...
-      dateColumn,
-      ...buildDimensionColumns(dimensions, columnIndex),
-      ...buildMetricColumns(metrics, columnIndex)
-    ];
+    let newColumns = [...buildDimensionColumns(dimensions, columnIndex), ...buildMetricColumns(metrics, columnIndex)];
+
+    //Add DateTime column when on timegrain other than All
+    if (timeGrain !== 'all') {
+      newColumns.unshift(dateColumn);
+    }
```
jkusa: Have you seen this technique?
```js
const newColumns = [
...(timeGrain !== 'all') && [dateColumn], 
...buildDimensionColumns(dimensions, columnIndex),
...buildMetricColumns(metrics, columnIndex)
];
```

mattgraham1995: I haven't. I didn't know you could do that with the spread operator

jkusa:
@mattgraham1995 as mentioned offline, can you double check this won't have a heavy impact to column customization for reports with `all` time grain. 

Also, please check dashboards, as they don't verify the visualization model.

packages/core/addon/models/table.js
mattgraham1995: Screen Shot 2019-04-01 at 3 40 42 PM

Couldn't get the spread to work like that so I changed it a different way to not use `unshift`

mattgraham1995:
@jkusa Checked dashboards and it works as expected. Checked existing reports with customized tables and they all retain their customizations while dropping the date column.

PR #351, opened 2019-04-04
Fix: Stop dimension filter initial loading of large dimensions.

### LINKED ISSUE #344: UX issue when filtering in dimension dropdown
When I type to filter on a dimension name, the combobox list is filtered by what I've typed.

After I select what I want, the search text doesn't go away but the combobox filter is dropped and everything is shown.

Alex-Aralis:
Resolves #344 

## Description
For large dimensions you must paste the term 2 or 3 times before the value shows up in the results. In some cases it briefly flashes the value but then disappears from the results list.

## Proposed Changes
- Stop loading all dimensions on initial filter dimension selector focus for large dimensions as defined by `config.navi.searchThresholds.contains`.

## Screenshots

No initial load all on focus for large cardinality dimension:
image

Load after value input:
image

### DIFF
--- a/packages/reports/addon/components/filter-values/dimension-select.js
+++ b/packages/reports/addon/components/filter-values/dimension-select.js
@@ -53,8 +53,6 @@ export default Ember.Component.extend({
       dimensionService = get(this, '_dimensionService'),
       metadataService = get(this, '_metadataService');
 
-    console.log(get(metadataService.getById('dimension', dimensionName), 'cardinality'));
-
     if (get(metadataService.getById('dimension', dimensionName), 'cardinality') <= LOAD_CARDINALITY) {
       return dimensionService.all(dimensionName);
     }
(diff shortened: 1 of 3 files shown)


packages/reports/addon/components/filter-values/dimension-select.js
```diff
@@ -42,9 +50,14 @@ export default Ember.Component.extend({
    */
   dimensionOptions: computed('filter.subject', function() {
     let dimensionName = get(this, 'dimensionName'),
-      dimensionService = get(this, '_dimensionService');
+      dimensionService = get(this, '_dimensionService'),
+      metadataService = get(this, '_metadataService');
+
+    if (get(metadataService.getById('dimension', dimensionName), 'cardinality') <= LOAD_CARDINALITY) {
+      return dimensionService.all(dimensionName);
```
daniellevinson: this works fine for low cardinality dimensions since searching also waits for `all` to resolve, right?

cythrawll: correct see: https://github.com/yahoo/navi/blob/master/packages/data/addon/services/bard-dimensions.js#L332

PR #352, opened 2019-04-04
Update: Check if report needs rerun

daniellevinson:
## Description
This tries to solve 2 UX issues:
1. We don't always need to rerun the report, the "Run request to update Table" message should not be displayed after changes like column title/format and the run button should be disabled.
2. editing mode of table remains "on" even if report is changed and `visualization-config` is hidden:
Apr-04-2019 14-33-39

## Proposed Changes
Bringing back here something similar to what we had before - a property controller `hasRequestRun` (`hasReportRun` is too confusing with the report completed status)
the main difference is that `previousRequest` lives in the `reports.report` controller rather that the `view` router so it can be accessed from both `runReport` in the `view` route and `onUpdateReport` in the `report` route. Not sure if it's a bad practice to read controller properties from child routes.

## Screenshots
Apr-04-2019 14-38-54

## Tests
TODO, would like to get some input before

### DIFF
--- a/packages/reports/addon/routes/reports/report.js
+++ b/packages/reports/addon/routes/reports/report.js
@@ -73,7 +73,10 @@ export default Route.extend({
    */
   setupController(controller) {
     this._super(...arguments);
-    controller.set('showSaveAs', false);
+    controller.setProperties({
+      showSaveAs: false,
+      modifiedRequest: null
+    });
   },
 
   /**
@@ -208,6 +211,15 @@ export default Route.extend({
       controller.set('reportState', state);
     },
 
+    /*
+     * @action setModifiedRequest
+     * @param {Object} request
+     */
+    setModifiedRequest(request) {
+      const controller = this.controllerFor(this.routeName);
+      controller.set('modifiedRequest', request);
+    },
+
     /**
      * @action onUpdateVisualizationConfig
      *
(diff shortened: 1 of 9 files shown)


packages/reports/addon/routes/reports/report.js
```diff
@@ -208,6 +209,24 @@ export default Route.extend({
       controller.set('reportState', state);
     },
 
+    /*
+     * @action setHasRequestRun
+     * @param {Boolean} value
+     */
+    setHasRequestRun(value) {
+      const controller = this.controllerFor(this.routeName);
+      controller.set('hasRequestRun', value);
```
jkusa: Is possible to compare the model's request to the current request to see if the current request has run?

daniellevinson: @jkusa so I've tried a few things
1. a property in the `view` controller, but can't think of a way to access the request of the parent model without setting it explicitly.
something like `computed('reportController.model.request' ...` or a cp `reportController.request` doesn’t seem to trigger any change
2. pass down `request` along with `response` to `view.hbs` and have a cp there that compares it with `report.request` - also doesn't work

either I'm missing something or the explicit `set` of the request was there for a reason, will try to think of another way to approach this.

In any case we _can_ get rid of `hasRequestRun` while still setting the request explicitly from `onUpdateReport`, I just think `hasRequestRun` might be useful to other child routes in the future

PR #353, opened 2019-04-08
Create findContains and findAllContains utils. (Alex version)

Alex-Aralis:
## Description
Right now, findByContains is not very helpful for complicated selectors with multiple `:contains` calls in its jquery equivalent.  I've created a new PR that handles, and tests, more edge cases and and has better behavior.

## Proposed Changes
- Add `findContains` and `findAllContains` to support using jquery style contains selectors without using jquery
- Eventually replace `findByContains` implementation with this one

### DIFF
--- a/packages/core/test-support/helpers/contains-helpers.js
+++ b/packages/core/test-support/helpers/contains-helpers.js
@@ -27,3 +27,62 @@ export function buttonContains(text) {
 export function linkContains(text) {
   return findByContains('a', text);
 }
+
+/**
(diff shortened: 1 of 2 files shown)


jkusa:
Instead of rebuilding jQ***y functionality, can we investigate the feasibility of using [sizzle](https://github.com/jquery/sizzle) directly in our tests?

cythrawll:
might as well just keep the j-word around then :/

cythrawll:
Closing as we decide to go Sizzle

PR #354, opened 2019-04-08
Update: Remove metric requirement validation

Alex-Aralis:
## Description
Fili is removing the requirement to choose a metric. We will be too. this will allow dimension only reporting.

Update ember data model for request to require selecting a metric.

Make sure alerts is good. make sure adapter handles empty metrics correctly.

## Proposed Changes
- Remove request model validation that requires at least one metric.

## Screenshots
image

image

image

### DIFF
--- a/packages/core/addon/models/bard-request/request.js
+++ b/packages/core/addon/models/bard-request/request.js
@@ -39,8 +39,8 @@ const Validations = buildValidations({
   metrics: [
     validator('has-many'),
     validator('length', {
-      min: 1,
-      message: 'At least one metric should be selected'
+      min: 0,
+      message: 'Metrics can be emtpy'
     })
   ],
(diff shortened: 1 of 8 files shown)


jkusa:
The last two screen shots show chart visualization available. Do they break?

packages/core/addon/models/bard-request/request.js
```diff
@@ -39,8 +39,8 @@ const Validations = buildValidations({
   metrics: [
     validator('has-many'),
     validator('length', {
-      min: 1,
-      message: 'At least one metric should be selected'
+      min: 0,
```
jkusa: Do we need the length? 
Would this work:
```js
validator('collection', {
  collection: true,
  message: 'must be a collection'
})
```

PR #356, opened 2019-04-11
Fix: Better table header sync in Chrome

cythrawll:
## Description

Found a bug where events don't fire quite at the right time to allow sync between invisible table headers and visible on in table visualizations

## Proposed Changes
This is a fix for Chrome only at the moment.  Might consider using a polyfill here if we see the same bug in other browers.


packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -100,6 +105,15 @@ export default Component.extend({
     );
 
     get(this, 'tableHeadersDomElement').addEventListener(WHEEL_EVENT, e => this._headerWheelSync(e));
+
+    //chrome better behavior
+    if (typeof window.ResizeObserver !== 'undefined') {
```
daniellevinson: should we still use the service in case it's supported?

cythrawll: I'm gonna look into a more comprehensive solution with the polyfill and yeah it won't use resize service when I'm done with it.

daniellevinson: oops missed the do not review label

cythrawll:
Now works in Firefox! added observer polyfill. removed the resize service, and added a mutation observer to observe when the contents of the hidden header changes (reorders, adding and removing columns). So it reregisters the resize observers.

packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -45,9 +40,14 @@ export default Component.extend({
   classNames: ['table-widget-container', 'table-widget-container--vc'],
 
   /**
-   * @property {Service} resize - Resize service to respond to view resize events
+   * @property {ResizeObserver} -- chrome resize observer for header changes
```
balajimanoharan: Is this an object? `@property {Object} `??

cythrawll: It's an object of the global ResizeObserver class.

PR #357, opened 2019-04-12
Fix: Next day selection bug.

cythrawll:
## Description
Bug reported that opening date picker in report builder, they couldn't select the next day if they already specified an interval.

reproduce step:
* Create a report with a custom date range and run it.
* open up custom date range and try to pick the day after your previous range selection as the end range
* Note that it won't let you pick the date.

## Proposed Changes
Found code that won't adjust the end correctly if date selected is the same day as the previous interval. Because the switching between inclusive-inclusive and inclusive-exclusive this doesn't work well.

Noticed that the advanced editor when you select a date, it will increase the date by 1 if you focus out.  Since this already is inclusive-exclusive range it shouldn't adjust. So created a separate setEnd action. So that logic from selecting a date from date picker, and selecting a date with advanced editor is handled separatly.

### DIFF
--- a/packages/reports/addon/components/pick-range-container.js
+++ b/packages/reports/addon/components/pick-range-container.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Description:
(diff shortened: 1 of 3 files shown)


packages/reports/addon/components/pick-range-container.js
```diff
@@ -103,16 +103,29 @@ export default PickObjectContainer.extend({
       this.send('stagePropertyChange', '_start', start);
     },
 
+    /**
+     * Used to set the end of the advanced range selector
+     * @param {String} - date entered into box, could be date or macro.
+     */
+    setAdvancedEnd(end) {
```
jkusa: Can we call this `setExclusiveEnd` unlike the other which is inclusive. I think it will make more sense why we have one start setter for both the date picker the and the adv input and two different individual end setters for the date picker and the adv input.

cythrawll: I think I can manage that.

PR #359, opened 2019-04-19
Apply dashboard filters

Alex-Aralis:
## Description
Apply global dashboard filters to widget requests if applicable.  Otherwise inject an error into the request response for that widget request.

## Proposed Changes
- Apply global dashboard filters to widget requests if applicable.
- Otherwise inject an error into the request response for that widget request.
- Test

### DIFF
--- a/packages/dashboards/addon/services/dashboard-data.js
+++ b/packages/dashboards/addon/services/dashboard-data.js
@@ -56,25 +56,22 @@ export default Service.extend({
     // Construct hash of widget id to data
     widgets.forEach(widget => {
       const widgetId = get(widget, 'id');
+      const dashboard = get(widget, 'dashboard');
       const widgetDataPromises = get(widget, 'requests').map(request => {
         //construct custom header for each widget with uuid
         options.customHeaders = {
           uiView: `dashboard.${dashboardId}.${uuid}.${widgetId}`
         };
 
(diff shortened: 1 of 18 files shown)


packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -47,25 +50,33 @@ export default Ember.Service.extend({
...
+        const requestDecorated = this._decorate(decorators, requestWithFilters.serialize());
 
-          request = this._decorate(decorators, request.serialize());
+        const filterErrors = this._getFilterErrors(dashboard, request);
 
-          return this._fetch(request, options);
+        return this._fetch(requestDecorated, options).then(result => {
+          const serverErrors = getWithDefault(result, 'response.errors', []);
+
+          return merge({}, result, { response: { errors: [...serverErrors, ...filterErrors] } });
```
cythrawll: I think we wanted this under the `meta` key under the response, any reason why we went with `errors`?

Alex-Aralis: I was following the JSON API errors object spec. https://jsonapi.org/examples/#error-objects-basics

cythrawll: But these aren't errors, just information the user should know about the filters that may/may not been applied.

PR #364, opened 2019-04-23
Fix: Unset height and min-height to allow vertical overflow of table when printing

daniellevinson:


jkusa:
@daniellevinson are these changes compatible with pdf printing via PhantomJS?

daniellevinson:
@kilroy they are compatible, although they don't solve the vertical scroller issue when there are many rows, my guess is it's something with @page/body/table margins

Apr-24-2019 14-58-52

PR #377, opened 2019-05-08
Update: Field selection on contains filter

cythrawll:
## Description
Allows user to choose the field of a dimension when using the "Contains" operator. 

## Proposed Changes
Update the dimension filter builder to allow for field selection of the operator supports it. We must also make sure we switch the field back to the primary key when choosing operators that do not support field choice.

## Screenshots
Screen Shot 2019-05-08 at 11 25 22 AM

### DIFF
--- a/packages/reports/addon/components/filter-builders/base.js
+++ b/packages/reports/addon/components/filter-builders/base.js
@@ -55,6 +55,11 @@ export default Ember.Component.extend({
         Ember.assign(changeSet, { values: [] });
       }
 
+      //switch field to primary key if operator does not allow choosing fields
+      if (get(this, 'primaryKeyField') && !operatorObject.showFields) {
+        Ember.assign(changeSet, { field: get(this, 'primaryKeyField') });
+      }
+
       this.onUpdateFilter(changeSet);
     }
   }
--- a/packages/reports/addon/components/filter-builders/base.js
+++ b/packages/reports/addon/components/filter-builders/base.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Base class for filter builders.
@@ -7,6 +7,7 @@
 import Ember from 'ember';
 import layout from 'navi-reports/templates/components/filter-builders/base';
 import { readOnly } from '@ember/object/computed';
+import { assign } from '@ember/polyfills';
 
 const { get } = Ember;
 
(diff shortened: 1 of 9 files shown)


packages/reports/addon/components/filter-builders/dimension.js
```diff
@@ -47,16 +49,38 @@ export default Base.extend({
...
+   * @property {?Boolean} showFields - Whether to show the field chooser in the filter builder
+   */
+  showFields: readOnly('filter.operator.showFields'),
+
+  /**
+   * @property {Array} fields - List of fields that a user can choose from
+   */
+  fields: computed('requestFragment.dimension', function() {
+    let fields = get(this, 'requestFragment.dimension.fields');
+    if (fields) {
```
mattgraham1995: `return fields ? fields.map(field => field.name) : ['id', 'desc'];`

packages/reports/tests/integration/components/filter-builders/dimension-test.js
```diff
@@ -0,0 +1,108 @@
...
+    assert.equal(changeSet.operator, 'contains', 'Selected operator is given to action');
+
+    this.set('filter.operator', this.supportedOperators.find(oper => oper.id === changeSet.operator));
+  });
+
+  this.render(
+    hbs`{{filter-builders/dimension filter=filter requestFragment=requestFragment supportedOperators=supportedOperators onUpdateFilter=(action onUpdateFilter)}}`
+  );
+
+  clickTrigger();
```
mattgraham1995: async await syntax?

cythrawll: Doesn't seem to need it. the integration function actions block

packages/reports/addon/components/filter-builders/base.js
```diff
@@ -55,6 +55,11 @@ export default Ember.Component.extend({
         Ember.assign(changeSet, { values: [] });
       }
 
+      //switch field to primary key if operator does not allow choosing fields
+      if (get(this, 'primaryKeyField') && !operatorObject.showFields) {
+        Ember.assign(changeSet, { field: get(this, 'primaryKeyField') });
```
jkusa: import { assign } from '@ember/polyfills';

packages/reports/addon/templates/components/filter-builders/dimension.hbs
```diff
@@ -0,0 +1,40 @@
...
+    selected=filter.field
+    tagName='span'
+    class='filter-builder-dimension__field'
+    dropdownClass='filter-builder-dimension__field-dropdown'
+    triggerClass='filter-builder-dimension__select-trigger'
+    matchTriggerWidth=false
+    onchange=(action 'setField')
+    as | field |
+  }}
+    field: {{field}}
```
jkusa: As discussed in the sync up, can this be just `{{field}}`

cythrawll: Yeah I did that because it bugs me there is no label for the form control. I see how it's weird though, I will respect the council's decision.

PR #384, opened 2019-05-15
Fix: Adjust json-url import

cythrawll:
## Description
Latest json-url changed it's dist install. In the long run we need to change this over to ember-auto-import. But until we merge in 3.x reports upgrade this will have to do for now.

## Proposed Changes
Update json-url and change import to reflect new file layout.


jkusa:
Can we pivot to ember-auto-import:

https://github.com/yahoo/navi/blob/54181e409be1647bd29e94afc54b2e5cfe0050fb/packages/reports/addon/services/model-compression.js#L21

PR #386, opened 2019-05-21
Fix destructure issue

mattgraham1995:
## Description
Some apps on older versions of Ember do not work well with Navi after the 3.9 Ember upgrade.

## Proposed Changes
- Remove problematic destructuring syntax
- In-template styling broke some ember-power-select dropdown instances where options would not appear

## Screenshots
Before:
Screen Shot 2019-05-21 at 4 14 40 PM

After:
Screen Shot 2019-05-21 at 4 15 18 PM


mattgraham1995:
@cythrawll @daniellevinson Please re-review now that the build is passing

PR #387, opened 2019-05-23
Fix: Saving a schedule in the dashboard view never finishes

mattgraham1995:
## Description
Saving a schedule from the dashboard view would never stop saving.

## Proposed Changes
- Use delivery-action-consumer actions to handle changes to the schedule modal like we do in dashboard and report collections

### DIFF
--- a/packages/dashboards/addon/templates/components/navi-dashboard.hbs
+++ b/packages/dashboards/addon/templates/components/navi-dashboard.hbs
@@ -75,9 +75,9 @@
           model=dashboard
           classNames="action schedule"
           disabled=(not dashboard.validations.isTruelyValid)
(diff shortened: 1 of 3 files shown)


packages/dashboards/addon/templates/components/navi-dashboard.hbs
```diff
@@ -75,9 +75,9 @@
           model=dashboard
           classNames="action schedule"
           disabled=(not dashboard.validations.isTruelyValid)
-          onSave=(route-action "saveDeliveryRule")
```
jkusa: Can we clean up the route actions?

mattgraham1995: Removed

PR #388, opened 2019-05-24
Fix tooltip problem

mattgraham1995:
## Description
Tooltips on an element/component that have overflow: hidden will hide part of the tooltips

## Proposed Changes
- Use body as base of the tooltip as mentioned here: https://github.com/sir-dunxalot/ember-tooltips/blob/master/UPGRADING-3.x.md#my-tooltips-appear-clipped-use-within-elements-using-overflow-hidden

## Screenshots
Before 
Screen Shot 2019-05-28 at 11 39 57 AM

After
Screen Shot 2019-05-24 at 4 28 32 PM

### DIFF
--- a/packages/reports/addon/templates/components/dimension-selector.hbs
+++ b/packages/reports/addon/templates/components/dimension-selector.hbs
@@ -19,7 +19,7 @@
     </label>
     {{#unless (eq item.category "Time Grain")}}
       {{#navi-icon "question-circle-o" class="grouped-list__item-info"}}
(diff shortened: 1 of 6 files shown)


cythrawll:
Do you still have the animation issue with attaching to the body? There is an option to change the animation type, could turn off the animation or change to "fade"  if it looks bad.

mattgraham1995:
@cythrawll That's fixed with the latest commit in this PR per your suggestion

cythrawll:
oops I might have been looking at an outdated commit.

PR #410, opened 2019-06-14
Navi app style update

mattgraham1995:
## Description
Style changes in navi added a gray background to dashboards

## Proposed Changes

- Adjust padding so that gray background stretches to left and right edges of window

## Screenshots
Screen Shot 2019-06-17 at 11 47 58 AM
Screen Shot 2019-06-17 at 11 48 03 AM
Screen Shot 2019-06-17 at 11 48 14 AM
Screen Shot 2019-06-17 at 11 48 09 AM

## License

### DIFF
--- a/packages/app/app/styles/views/reports.less
+++ b/packages/app/app/styles/views/reports.less
@@ -1,0 +1,11 @@
+.navi-report {
+  &__breadcrumb,
+  &__header,
+  &__footer {
+    padding: 0 20px;
+  }
+
+  .report-builder {
+    padding: 5px 15px;
+  }
+}
--- a/packages/app/app/styles/views/reports.less
+++ b/packages/app/app/styles/views/reports.less
@@ -6,11 +6,11 @@
   &__breadcrumb,
   &__header,
   &__footer {
-    padding: 0 20px 0 30px;
+    padding: 0 20px 0 25px;
   }
 
   .report-builder {
-    padding: 0 20px;
+    padding: 20px;
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/app/app/styles/views/reports.less
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright 2019, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+.navi-report {
+  &__breadcrumb,
+  &__header,
+  &__footer {
+    padding: 0 20px;
```
balajimanoharan: Can we make this the same as the dashboard?
https://github.com/yahoo/navi/pull/410/files#diff-035417c76de71435bce60d8f3cb07391R12

packages/app/app/styles/views/reports.less
```diff
@@ -0,0 +1,15 @@
...
+  }
+
+  .report-builder {
+    padding: 5px 15px;
```
balajimanoharan: And should this also be the same as the dashboard body?

packages/app/app/styles/views/reports.less
mattgraham1995: I intended to preserve the original padding while extending the gray background, but yeah I can make them more consistent

jkusa:
Can we add equal padding around the report view?

mattgraham1995:
@jkusa You mean like this?
Screen Shot 2019-06-26 at 11 13 49 AM

jkusa:
Is the dashboard index view up to date? Looks like the title needs to move to the left a little

balajimanoharan:
Can we add the gray backgrounds to the index views as well? It would be nice if it was consistent.

PR #411, opened 2019-06-19
Fix: Fix dashboard filter values.

cythrawll:
## Description

Bug where widgets do not get the values of the global dashboard filters added correctly to the request.

## Proposed Changes

- request `addFilter` does not do well when the input is already a data fragment class, so added`addRawFilter` which copies filter values via 'rawValues' which works better in that case and does not break things that are currently using `addFilter`
- big acceptance test

## License

### DIFF
--- a/packages/dashboards/tests/acceptance/dashboard-filter-test.js
+++ b/packages/dashboards/tests/acceptance/dashboard-filter-test.js
@@ -9,7 +9,7 @@ module('Acceptance | Dashboard Filters', function(hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
 
-  test('Test dashbaord filter flow', async function(assert) {
+  test('dashboard filter flow', async function(assert) {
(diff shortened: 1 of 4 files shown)


packages/dashboards/tests/acceptance/dashboard-filter-test.js
```diff
@@ -0,0 +1,91 @@
...
+import { module, test } from 'qunit';
+import config from 'ember-get-config';
+import { setupApplicationTest } from 'ember-qunit';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+
+module('Acceptance | Dashboard Filters', function(hooks) {
+  setupApplicationTest(hooks);
+  setupMirage(hooks);
+
+  test('Test dashbaord filter flow', async function(assert) {
```
mattgraham1995: `dashbaord` typo

PR #413, opened 2019-06-20
Fix: Report run button always runs when clicked

cythrawll:
## Description

Found bug that sometimes the Run button in reports won't run the report, even though you should.  This puts the report in an awkward state that's not easily correctable by the user.

See acceptance test for an example of this bad behavior.  changing a report, reverting, and changing it back to the same report breaks the run button functionality.

## Proposed Changes

- Change behavior of run button so that it will always run the report when clicked.

## License

### DIFF
--- a/packages/reports/tests/acceptance/navi-report-test.js
+++ b/packages/reports/tests/acceptance/navi-report-test.js
@@ -1752,4 +1752,39 @@ module('Acceptance | Navi Report', function(hooks) {
       'When not loading a report, the standard footer buttons are available'
     );
   });
+
+  test('Recreating same report after revert runs', async function(assert) {
+    await visit('/reports/2/view');
+
+    const columns = () => findAll('.table-widget__table-headers .table-header-cell').map(el => el.textContent.trim());
+    assert.deepEqual(columns(), ['Date', 'Property', 'Ad Clicks', 'Nav Clicks'], 'Report loads with expected columns');
+
+    await click(
+      findAll('.checkbox-selector--metric .grouped-list__item')
+        .find(el => el.textContent.trim() == 'Page Views')
+        .querySelector('.grouped-list__item-label')
+    );
+    await click('.navi-report__run-btn');
+    assert.deepEqual(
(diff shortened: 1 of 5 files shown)


packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1752,4 +1752,39 @@ module('Acceptance | Navi Report', function(hooks) {
...
   });
+
+  test('Recreating same report after revert runs', async function(assert) {
+    await visit('/reports/2/view');
+
+    const columns = () => findAll('.table-widget__table-headers .table-header-cell').map(el => el.textContent.trim());
+    assert.deepEqual(columns(), ['Date', 'Property', 'Ad Clicks', 'Nav Clicks'], 'Report loads with expected columns');
+
+    await click(
+      findAll('.checkbox-selector--metric .grouped-list__item')
```
mattgraham1995: We could just use contains to select on the right metric 
`await click($('.checkbox-selector--metric .grouped-list__item:contains(Page Views) .grouped-list__item-label')[0]);`

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1752,4 +1752,39 @@ module('Acceptance | Navi Report', function(hooks) {
...
+      columns(),
+      ['Date', 'Property', 'Ad Clicks', 'Nav Clicks', 'Page Views'],
+      'Report changed and ran successfully'
+    );
+
+    await click('.navi-report__revert-btn');
+    assert.deepEqual(columns(), ['Date', 'Property', 'Ad Clicks', 'Nav Clicks'], 'Report revertted successfully');
+
+    //make same changes and make sure it's runnable
+    await click(
```
mattgraham1995: Same as above

packages/reports/tests/acceptance/navi-report-test.js
cythrawll: I'm totally over `contains`

PR #414, opened 2019-06-24
Update: Prevent changes to the report while it is running

mattgraham1995:
## Description
While a report is running, changing the request by interacting with the report builder was causing a disconnect between the visualization/request of the report and the response data. 

## Proposed Changes

- Prevent use of the report builder components while the report is running
- Disable filter icons and checkboxes for metric and dimension selectors while report is running, but still allow navigation through them
- Completely disable all clicks on the filter collection while report is running
- Disable logical table selector dropdown
- Give disabled components a slightly darker gray background

## Screenshots
Screen Shot 2019-06-24 at 6 36 52 PM

## License

### DIFF
--- a/packages/reports/app/styles/navi-reports/components/report-builder.less
+++ b/packages/reports/app/styles/navi-reports/components/report-builder.less
@@ -25,9 +25,11 @@
 
   &__container--disabled {
     background: @denali-gray-300;
+    cursor: not-allowed;
 
     .ember-power-select-trigger {
       background: none;
+      pointer-events: none;
     }
   }
 
@@ -65,7 +67,11 @@
 
     &-disabled {
       background: @denali-gray-300;
-      pointer-events: none;
+      cursor: not-allowed;
+
+      .filter-collection {
+        pointer-events: none;
+      }
 
       .ember-power-select-trigger {
         background: none;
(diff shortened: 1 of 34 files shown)


packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1705,12 +1717,16 @@ module('Acceptance | Navi Report', function(hooks) {
   });
 
   test('Cancel Report', async function(assert) {
+    assert.expect(30);
     //Slow down mock
```
kevinhinterlong: Remove this?

packages/reports/app/styles/navi-reports/components/report-builder.less
```diff
@@ -23,6 +23,14 @@
     padding: 5px 10px;
   }
 
+  &__container--disabled {
+    background: @denali-gray-300;
```
jkusa: Can we set `cursor: not-allowed;`;

cythrawll: is there a `cursor: banished-to-void-realm;`?

jkusa: And `pointer-events: none;`

mattgraham1995: What do you guys think about showing the selected metrics and dimensions when the report is run? Then when the report finishes running, the metric and dimension selectors will go back to whatever state they were in before running.

PR #417, opened 2019-06-26
Fix: Don't apply global dashboard filters that don't have values

mattgraham1995:
## Description
Removing the last value of a global dashboard filter would result in decorating widget requests with a filter with no values which would result in a bad request error.

## Proposed Changes

- Don't apply filters that don't have at least one value

## License

### DIFF
--- a/packages/dashboards/tests/unit/services/dashboard-data-test.js
+++ b/packages/dashboards/tests/unit/services/dashboard-data-test.js
@@ -63,7 +63,7 @@ module('Unit | Service | dashboard data', function(hooks) {
 
     assert.deepEqual(service.fetchDataForWidgets(1, []), {}, 'no widgets returns empty data object');
 
-    const makeRequest = (data, filters) => ({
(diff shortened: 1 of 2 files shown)


packages/dashboards/tests/unit/services/dashboard-data-test.js
```diff
@@ -222,7 +223,11 @@ module('Unit | Service | dashboard data', function(hooks) {
       }
     });
 
-    const makeFilter = ({ dimension }) => ({ dimension: { name: dimension } });
+    const makeFilter = ({ dimension }) => {
```
jkusa: Do we have any test asserts with this change?

mattgraham1995: Yes the existing test checks that the correct filters are applied to the widget. Now the condition is that the filters have values so we would expect `dim4` to not be added while the others are added as usual

PR #423, opened 2019-07-02
Fix: Tooltips in reports and directory

daniellevinson:
## Description

This fixed a few issues with the tooltips:

❌ Current (second row is an unsaved report):
- **export** expands vertically
- **export** remains visible when not hovering
- **schedule** is never visible
- for unsaved reports, **share** is never visible and button cursor/color styles are not disabled
- overall vertical alignment

old-1

✅ Fixed:

fixed-1

❌ Current:
- line breaks in **export**
- **export** remains visible when not hovering

old-2

✅ Fixed:

fixed-2

## Proposed Changes

- fix parent elements of some tooltips, manually set `targetId` attribute  when necessary.
- setting `delay=0` instead of only `openDelay=0` for `basic-dropdown-hover` seems to fix the issue of export remaining visible.
- css changes

## License

### DIFF
--- a/packages/directory/app/styles/navi-directory/components/dir-asset-row-actions.less
+++ b/packages/directory/app/styles/navi-directory/components/dir-asset-row-actions.less
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 .dir-asset-row-actions {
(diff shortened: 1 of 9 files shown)


packages/directory/app/styles/navi-directory/components/dir-asset-row-actions.less
```diff
@@ -15,7 +15,10 @@
 
     .action {
       margin-right: 15px;
+      margin-bottom: 2px;
       visibility: hidden;
+      height: 17px;
+      line-height: 17px;
```
mattgraham1995: We generally try to keep styles alphabetized

PR #425, opened 2019-07-08
Fix: Support Ember 2.x

cythrawll:
## Description
Fixes issue where charts wouldn't work with old ember versions

## Proposed Changes

- Make sure we use `get` until we drop 2.x support

## License

### DIFF
--- a/packages/core/addon/components/navi-visualizations/bar-chart.js
+++ b/packages/core/addon/components/navi-visualizations/bar-chart.js
@@ -10,7 +10,7 @@
  */
 
 import LineChart from './line-chart';
-import { computed } from '@ember/object';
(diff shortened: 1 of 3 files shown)


packages/core/addon/components/navi-visualizations/bar-chart.js
```diff
@@ -27,6 +27,6 @@ export default LineChart.extend({
    * @property {String} c3ChartType - c3 chart type
    */
   c3ChartType: computed('chartType', function() {
-    return this.chartType;
+    return get(this, 'chartType');
```
daniellevinson: I think it only affects cp's so `this.options` or `this.chartType` would still work

PR #426, opened 2019-07-08
Fix: Travis build runs tests twice

cythrawll:
## Description

travis running tests twice, this doesn't seem to be the intention

## License

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
 install:
   - travis_wait npm install
 script:
-  - lerna run test --scope $PACKAGE -- bower install || true
+  - lerna exec --scope $PACKAGE -- bower install || true
   - travis_wait lerna run test --scope $PACKAGE
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,7 +29,7 @@ before_install:
 install:
   - travis_wait npm install
```
kevinhinterlong: Maybe switch this over to [`npm ci`](https://docs.npmjs.com/cli/ci.html) while we're at it

cythrawll: sure did that, unfortunately`lerna bootstrap` won't use `npm ci` until we upgrade lerna to `3.x` we had problems with upgrading however, we should look into it again.

balajimanoharan: lerna publish in the latest version fails with the current way we publish. Either we would need a tag in npm to use the current publish method or swap our current publish method to not do a canary release.

cythrawll: or we can release a version, that would work too

PR #439, opened 2019-07-17
add a checkbox for the schedule report pop up

Ruiqian1:
Resolves #

## We add a checkbox at the schedule report pop up

-

## 
Screen Shot 2019-07-17 at 9 43 47 AM

## License


cythrawll:
Nice, we have been thinking about doing something similar.

* Add this to the`schedulingRules` hash in the `deliveryRule` model so it gets persisted correctly.
* Put behind a featureFlag so we can hide this option until we have support in the jobs-service for this.

balajimanoharan:
We had an initial design for this feature. The checkboxes would be rendered below the dropdowns. (screenshot below). Can we do something like that?

Screen Shot 2019-07-17 at 10 59 49 AM

Ruiqian1:
@balajimanoharan @cythrawll I have updated the schedule report model and put the checkbox behind a feature flag. The layout has also been adjusted. Please review and let me know if there are any further questions.
Screen Shot 2019-07-18 at 1 41 17 PM

packages/core/addon/models/delivery-rule.js
```diff
@@ -37,6 +37,7 @@ export default DS.Model.extend(Validations, {
   format: DS.attr({ defaultValue: () => {} }),
   recipients: DS.attr({ defaultValue: () => [] }),
   version: DS.attr('number', { defaultValue: '1' }),
+  notifyWhenEmpty: DS.attr('number', { defaultValue: '0' }),
```
mattgraham1995: can we use `boolean` instead of `number`?

cythrawll: Actually this needs to be in scheduling rules, not it's own property.

Ruiqian1: Can you explaining a little more on why we need to put this under schedulingRules instead of at the same level of format, frequency, and recipients?

jkusa:
Can we use a toggle instead of checkbox?

jkusa:
Can we use a name that indicates that the system should not send a notification when there is an empty result?

Ruiqian1:
@jkusa I have changed the checkbox to be a toggle and revised the label. Here is how is looks like now
Screen Shot 2019-07-19 at 3 43 21 PM

Ruiqian1:
@cythrawll In the schedule rule UI, we have frequency, format, as well this boolean flag value indicating whether we want report to be sent if it is empty. I think all those 3 options should be of the equivalent to each other. For this reason, I think it would be find to have this boolean value as a separate attribute just as frequency and format.

packages/reports/addon/templates/components/common-actions/schedule.hbs
```diff
@@ -81,6 +81,17 @@
           {{/power-select}}
         </fieldset>
       </div>
+      {{#if (feature-flag "enableNotifyWhenEmpty")}}
+        <div class="schedule-modal__section-group">
+          {{x-toggle
+            theme="ios"
```
jkusa: Can we remove this to match the other toggles

jkusa: Can we add some margin around this component it seems a little close to the select above

Ruiqian1: cool. I am not an apple fan tbh XD

packages/reports/addon/templates/components/common-actions/schedule.hbs
```diff
@@ -81,6 +81,17 @@
...
+            size="small"
+            class="schedule-modal__toggle schedule-modal__toggle--notifyWhenEmpty"
+            value=localDeliveryRule.notifyWhenEmpty
+            onToggle=(mut localDeliveryRule.notifyWhenEmpty)}}
+          <div class="schedule-modal__label">Do Not Send The Report If It Is Empty</div>
```
jkusa: `Only send if data is present`?

Ruiqian1: sure

packages/core/addon/models/delivery-rule.js
```diff
@@ -37,6 +37,7 @@ export default DS.Model.extend(Validations, {
   format: DS.attr({ defaultValue: () => {} }),
   recipients: DS.attr({ defaultValue: () => [] }),
   version: DS.attr('number', { defaultValue: '1' }),
+  notifyWhenEmpty: DS.attr('boolean', { defaultValue: false }),
```
jkusa: Can we move this to the `schedulingRules` and call it `mustHaveData` with a default of `false`

`schedulingRules` is intended to be where we put advanced settings.

packages/reports/tests/dummy/config/environment.js
```diff
@@ -42,6 +42,7 @@ module.exports = function(environment) {
       FEATURES: {
         enableScheduleReports: true,
         enableMultipleExport: true,
+        enableNotifyWhenEmpty: true,
```
jkusa: `enabledNotifyIfData`

Ruiqian1:
@jkusa I have just changed the variable name from `notifyWhenEmpty` to `mustHaveData`, I also added a 5px margin at the top of the div that contains the toggle. Please let me know if you have further concerns.

Ruiqian1:
@jkusa @cythrawll @balajimanoharan Hi, I have addressed the comments. Please take a look of the PR when you have time. Thanks :)

mattgraham1995:
In the `deliveryRule` model, could we set the default value for `mustHaveData` in `schedulingRules`?

`schedulingRules: DS.attr({ defaultValue: () => {} })` 
=>
`schedulingRules: DS.attr({ defaultValue: () => { 
return featureFlag('enabledNotifyIfData') ? { mustHaveData: false } : {}; })`

packages/reports/addon/templates/components/common-actions/schedule.hbs
```diff
@@ -81,6 +81,16 @@
           {{/power-select}}
         </fieldset>
       </div>
+      {{#if (feature-flag "enabledNotifyIfData")}}
+        <div class="schedule-modal__section-group">
+          {{x-toggle
+            size="small"
+            class="schedule-modal__toggle schedule-modal__toggle--notifyWhenEmpty"
```
balajimanoharan: Can we rename the class name to `schedule-modal__must-have-data-toggle`? 
We don't use camel case for class names.

Ruiqian1: The convention we have for class name is schedule-modal__{type}--{name}. I will change it to `schedule-modal__toggle--must-have-data`

balajimanoharan: actually the convention we follow is called BEM. You can check this out for what BEM stands for.
https://css-tricks.com/bem-101/

Ruiqian1: Oh. That's awesome. Thanks!

balajimanoharan: `must-have-data` is not a modifier. Modifiers are things like `disabled`, `active`, `invalid`, `red`, etc.. Which is why `schedule-modal__must-have-data-toggle` is a better name according to the convention.

Ruiqian1: I see. Let me change that!

Ruiqian1: Just addressed the naming convention concern.

packages/reports/tests/acceptance/schedule-modal-test.js
```diff
@@ -93,6 +97,9 @@ module('Acceptance | Navi Report Schedule Modal', function(hooks) {
     await click('.schedule-modal__dropdown--frequency .ember-power-select-trigger');
     await click($('.ember-power-select-option:contains(Day)')[0]);
 
+    // Set notify when empty to be true
```
balajimanoharan: I think we can remove this comment.

packages/reports/tests/acceptance/schedule-modal-test.js
```diff
@@ -117,6 +124,10 @@ module('Acceptance | Navi Report Schedule Modal', function(hooks) {
       .dom('.schedule-modal__dropdown--frequency .ember-power-select-selected-item')
       .hasText('Day', 'Changes made to the frequency field are kept after clicking save changes');
 
+    assert
+      .dom('.schedule-modal__must-have-data-toggle .x-toggle')
+      .isChecked('the notify when empty toggle should be toggled on after toggling it on');
```
balajimanoharan: `must have data toggle is checked when clicked`

packages/reports/tests/acceptance/schedule-modal-test.js
```diff
@@ -210,6 +221,9 @@ module('Acceptance | Navi Report Schedule Modal', function(hooks) {
     await click('.schedule-modal__dropdown--frequency .ember-power-select-trigger');
     await click($('.ember-power-select-option:contains(Day)')[0]);
 
+    // Set notify when empty to be true
```
balajimanoharan: remove comment, we don't have a `notifyWhenEmpty` property anymore.

packages/reports/tests/acceptance/schedule-modal-test.js
```diff
@@ -244,6 +262,9 @@ module('Acceptance | Navi Report Schedule Modal', function(hooks) {
...
```
balajimanoharan: remove comment.

packages/reports/tests/acceptance/schedule-modal-test.js
```diff
@@ -261,6 +282,10 @@ module('Acceptance | Navi Report Schedule Modal', function(hooks) {
       'Recipients field changes to a new schedule kept but not saved to the store'
     );
 
+    assert
+      .dom('.schedule-modal__must-have-data-toggle .x-toggle')
+      .isNotDisabled('notify when empty changes to an existing schedule are discarded after clicking cancel');
```
balajimanoharan: Did you mean to check `isNotChecked` here instead of `isNotDisabled`?

Ruiqian1: Good catch! Just changed it.

packages/reports/tests/acceptance/schedule-modal-test.js
mattgraham1995: Are we going to update these comments? The component is called `must have data` instead of `notify when empty` now so the comments don't really make sense

Ruiqian1:
@cythrawll Thank you for reviewing this PR. I don't see to have access to merge it. Could you please grant me the access if possible or could you pleases merge the PR?

PR #445, opened 2019-07-18
Update: Move model compression to a generic service

mattgraham1995:
## Description
Compression service is needed for compressing a fragment to support an upcoming feature.
Currently, we have a service for compressing ember data models specifically within the reports package.

## Proposed Changes

- Move model-compression service to navi-core as just simply compression service
- Add generic compress and decompress functions to that service that works for more than models
- Use ember-auto-import to access the codec

## License

### DIFF
--- a/packages/core/addon/services/compression.js
+++ b/packages/core/addon/services/compression.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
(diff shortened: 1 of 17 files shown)


packages/core/addon/services/compression.js
```diff
@@ -8,6 +8,7 @@ import Service, { inject as service } from '@ember/service';
 import { assert } from '@ember/debug';
 import { get, computed } from '@ember/object';
 import { run } from '@ember/runloop';
+import JsonUrl from 'json-url';
```
jkusa: 🎉

PR #457, opened 2019-07-29
Chart stacking order

daniellevinson:
## Description

This adds the ability to control the order of data stacking.

## Proposed Changes

- a new component `chart-series-config` that uses `ember-sortable` to allow reordering of series.
- metric selector is moved from the `dimension` component (`report-view` -> `wrapper` -> `line-chart` -> `dimension`) up to the `wrapper` component so it's rendered above all line chart options.
- wrapper tests file was `wrapper.js` instead of `wrapper-test.js` and didn't even run (how could it with `import { get } from '@meber/object';`) 😺 

## Screenshots

stack-order

## License


packages/core/addon/components/visualization-config/wrapper.js
```diff
@@ -11,10 +11,67 @@
...
+  seriesConfig: computed('series', function() {
+    return get(this, 'series.config');
+  }),
+
+  /**
+   * @property {Boolean} showMetricSelect - whether to display the metric select
+   */
+  showMetricSelect: computed('visualization', 'series', function() {
+    const series = get(this, 'series'),
+      { visualization: { type }, request } = this,
```
cythrawll: make sure this works in ember 2.x

jkusa:
Looks nice, but I think we need to keep the original color/series mapping after reorder. Reorganizing the series colors after the reorder is a bit confusing.

Using something like this could address this issue plus pave the way for custom color selection:
https://c3js.org/samples/data_color.html

packages/core/addon/components/visualization-config/wrapper.js
daniellevinson: Thanks for the reminder to test in 2.x 🙏 
This works since both `visualization` and `request` are not cp's, but `this.metricName` had to be changed `this.get('metricName')`

daniellevinson:
@jkusa we actually had this discussion in one of the stand-ups and agreed it should not be part of the scope of this (already big) PR. It would require a new separate configuration and would complicate things a little.

PR #458, opened 2019-07-30
Fix: Dashboard creation for ember 2.x

cythrawll:
## Description

issue that users using ember 2.x are not able to create new dashboards

## Proposed Changes

- `currentRoute` is a property on `router` in 3.x, not 2.x... use alternative method to get queryParams from transition object in model hooks

## License

### DIFF
--- a/packages/dashboards/addon/routes/dashboards/new.js
+++ b/packages/dashboards/addon/routes/dashboards/new.js
@@ -71,7 +71,6 @@ export default Route.extend({
    * @returns {DS.Model} route model
    */
   _newModel(title = 'Untitled Dashboard') {
-    debugger;
     return get(this, 'user')
       .findOrRegister()
       .then(author =>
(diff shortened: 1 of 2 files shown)


packages/dashboards/addon/routes/dashboards/new.js
```diff
@@ -62,6 +71,7 @@ export default Route.extend({
    * @returns {DS.Model} route model
    */
   _newModel(title = 'Untitled Dashboard') {
+    debugger;
```
kevinhinterlong:
```suggestion
```

cythrawll: Pushed it before you commented, ;) good catch

cythrawll: Pushed it before you commented. Good catch :)

PR #460, opened 2019-08-01
New: Synchronous get method to bard-dimensions

kevinhinterlong:
## Description
The current findById method will return a not resolved promise even if the dimension value is already loaded in keg

## Proposed Changes
Add a `getById` method to `bard-dimensions` service and `keg` adapter

## License


packages/data/tests/unit/adapters/dimensions/keg-test.js
```diff
@@ -115,6 +115,16 @@ module('Unit | Adapters | Dimensions | Keg', function(hooks) {
     });
   });
 
+  test('getById', function(assert) {
+    assert.expect(1);
```
mattgraham1995: Are we testing for the case where the record is not found anywhere?

PR #472, opened 2019-08-07
Publish gh-pages in new travis job

kevinhinterlong:
## Description
The current publish step combines both `npm` and `gh-pages` jobs which can be confusing if something fails/needs to be rerun.

<details>
    <summary>Why the build failed</summary>

The `travis-secrets.tar.gz` extracts into a `.travis-secrets` folder which caused [the build](https://travis-ci.org/yahoo/navi/jobs/568047403#L1900-L1904) to fail since we were referencing `travis-secrets`.

The references are now correct and I've moved `travis-secrets.tar.gz.enc` → `.travis-secrets.tar.gz.enc` to indicate this
</details>

## Proposed Changes
- Keep the `npm` publish independent
- Add a job just for `gh-pages` publishing
- Decrypt travis secrets to `/tmp` to be safe and to have absolute path when connecting to github via ssh
- Update `travis-secrets` -> `.travis-secrets` in scripts 

- *.sh replace `"` -> `'`

_We should probably follow this route for publishing the webservice as well_

## License


scripts/gh-pages-publish.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jkusa: Can this be `#!/usr/bin/env bash`?

scripts/npm-publish.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
jkusa: Same here

PR #481, opened 2019-08-20
Ember c3 upgrade

### LINKED ISSUE #315: Use latest ember-c3 and remove c3 & d3 from bower
Blockers:
- [ ] fix Issue in c3 (NPE)

cythrawll:
Resolves #315 

## Description

Update ember-c3 which takes c3 and d3 out of bower.

## Proposed Changes

- Updates ember c3, which not only gets rid of bower controlled c3 and d3, but allows us to use a newer version of the c3 lib. which means we can get rid of a lot of c3 enhancements that are no longer necessary. (though introduced a new bug that we had to work around for now with unloading the component)

- Translation calculation moved out of the d3 api, so had to create a utility.

- Goal gauge api changed a bit, so small changes to get the same behavior was needed.

- property name changes in ember-c3

## License

### DIFF
--- a/packages/app/bower.json
+++ b/packages/app/bower.json
@@ -3,8 +3,6 @@
   "dependencies": {
     "pretender": "~1.1.0",
     "numeral": "~1.5.3",
-    "d3": "3.5.17",
-    "c3": "0.4.20",
     "bootstrap-datepicker": "1.4.0"
   }
 }
--- a/packages/core/addon/components/navi-vis-c3-chart.js
+++ b/packages/core/addon/components/navi-vis-c3-chart.js
@@ -20,11 +20,11 @@ import { next, scheduleOnce } from '@ember/runloop';
 import { assign } from '@ember/polyfills';
 import { set, getProperties, get, computed } from '@ember/object';
 import C3Chart from 'ember-c3/components/c3-chart';
+import c3 from 'c3';
 import { A } from '@ember/array';
 import { inject as service } from '@ember/service';
 import $ from 'jquery';
 
-/* globals c3: true */
 export default C3Chart.extend({
   /**
    * @property {Service} metricName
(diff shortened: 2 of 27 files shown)


packages/app/bower.json
```diff
@@ -3,8 +3,6 @@
   "dependencies": {
     "pretender": "~1.1.0",
     "numeral": "~1.5.3",
-    "d3": "3.5.17",
-    "c3": "0.4.20",
```
kevinhinterlong: 🎉

packages/core/addon/components/navi-vis-c3-chart.js
```diff
@@ -169,8 +169,8 @@ export default C3Chart.extend({
    * @returns {Void}
    */
   _teardownChart() {
-    get(this, 'chart').destroy();
-    set(this, '_chart', null);
+    //get(this, 'chart').destroy();
```
kevinhinterlong:
```suggestion

```

packages/core/addon/initializers/inject-c3-enhancements.js
```diff
@@ -4,6 +4,7 @@
  */
 import { set } from '@ember/object';
 import { A } from '@ember/array';
+import c3 from 'c3';
```
jkusa: 👍

packages/core/addon/utils/chart.js
```diff
@@ -0,0 +1,30 @@
+/**
```
jkusa: ©️

packages/core/tests/unit/initializers/inject-c3-enhancements-test.js
```diff
@@ -18,7 +19,6 @@ module('Unit | Initializer | inject c3 enhancements', function(hooks) {
 
   test('function overrides', function(assert) {
     initialize(this.application);
-
-    assert.equal(c3.chart.internal.fn.getGaugeLabelHeight(), 100, 'c3 function overrides are present');
+    assert.notOk(c3.chart.internal.fn.isCustomX());
```
jkusa: Can we add an assert msg?

packages/core/addon/utils/chart.js
```diff
@@ -0,0 +1,30 @@
...
+ * given a transform, returns the translation that transform provides
+ * @param {String} transform - transform to calculate
+ * @returns {Object} - Object with x and y properties
+ */
+
+export function getTranslation(transform) {
```
jkusa: Do we have tests for this?

jkusa: Is this method something we could pull in from `d3-interpolate` or is this somewhat different?

packages/core/addon/components/navi-vis-c3-chart.js
jkusa: Is the cleanup logic happening elsewhere now?

jkusa: This was also called on attr update

cythrawll: this absolutely breaks things when you try to destroy from attr update

PR #491, opened 2019-08-30
Update: Enable bar and pie charts for single row multi metrics

daniellevinson:
## Description

This enables bar and pie charts for a single time bucket w\ multiple metrics (a dimension is no longer required)

## Proposed Changes

- allow a series of type `metric` in pie chart model
- update `typeIsValid` in manifests
- dedup `chartBuilders` map creation code
- metric label is visible only for a series of type `dimension`
- updates to vis-config
- update tests

## Screenshots

pie

## License


packages/core/addon/components/visualization-config/wrapper.js
```diff
@@ -22,11 +22,19 @@ export default Component.extend({
 
   classNames: ['visualization-config'],
 
+  /**
+   * @property {String} seriesPath
+   */
+  seriesPath: computed('visualization.type', function() {
+    return get(this, 'visualization.type') === 'pie-chart' ? 'series' : 'axis.y.series';
```
jkusa: Is there a way we can make the wrapper unaware of concrete visualization types?

daniellevinson: So all of these props and methods are here because moving the metric selector above other line/bar options mixed up things a little.
Inside the wrapper we're now rendering:
- metric selector **if series and of type dimension**
- specific chart config by type
- series add/remove **if series and of type dimension**

We _could_ delegate all of the metric selector logic from the wrapper back to `chart-type/dimension` (as it was before) and in line/bar/pie chart do something like 
```
{{#component
  (concat typePrefix seriesType) <-- e.g. visualization-config/chart-type/dimension
  ...
}}
   {{!-- specific line/bar chart options --}}
{{/component}}
```

while `visualization-config/chart-type/dimension` would be something like
```
{{#if showMetricSelect}}
  <div class="metric-select">...</div>
{{/if}}
{{yield}}
{{chart-series-collection   <-- add/remove series
    ...
}}
```

The other option which feels hacky to me would be to keep it in the wrapper and just check for `visualization.metadata.series` or `visualization.metadata.axis.y.series` to determine whether it's a series without relying on chart type.

I'm not sure about either, would like to hear your suggestion and I'll see if I can come up with something more generic.

Also keep in mind that currently `visualization-config/chart-type/date-time` and `visualization-config/chart-type/metric` don't render anything at all.

packages/core/app/styles/navi-core/components/visualization-config/series-chart.less
```diff
@@ -0,0 +1,28 @@
...
+    color: @navi-gray;
+    margin: 5px;
+  }
+
+  &__select {
+    .ember-power-select-status-icon {
+      right: 10px;
+    }
+
+    .ember-power-select-option[aria-current='true'] {
```
kevinhinterlong: Delete this block (goodbye purple)

packages/core/addon/components/navi-visualizations/pie-chart.js
```diff
@@ -261,7 +271,9 @@ export default Component.extend({
     redrawMetricLabel() {
       if (!get(this, 'isDestroyed') && !get(this, 'isDestroying')) {
         this._removeMetricLabel();
-        this._drawMetricLabel();
+        if (get(this, 'seriesType') === 'dimension') {
```
jkusa: Could this be a check on 251 such as:
```js
 if(metricTitle) {
   //insert label
}
```

packages/core/addon/components/visualization-config/series-chart.js
```diff
@@ -37,6 +38,22 @@ export default Component.extend({
     return getRequestMetrics(get(this, 'request'));
   }),
 
+  /**
+   * @property {Object} selectedMetric
+   */
+  selectedMetric: computed('seriesConfig', function() {
```
jkusa: `readOnly`?

packages/core/addon/manifests/bar-chart.js
```diff
@@ -32,6 +32,9 @@ export default ManifestBase.extend({
    * @return {Boolean} - visualization type is valid
    */
   typeIsValid(request) {
-    return this.hasMetric(request) && (this.hasGroupBy(request) || this.hasMultipleTimeBuckets(request));
+    return (
+      this.hasMultipleMetrics(request) ||
```
jkusa: what request do we not want a bar chart?

packages/core/addon/mixins/components/has-chart-builders.js
```diff
@@ -0,0 +1,32 @@
...
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ */
+import Mixin from '@ember/object/mixin';
+import config from 'ember-get-config';
+import { getOwner } from '@ember/application';
+import { computed } from '@ember/object';
+import { camelize } from '@ember/string';
+
+/* global requirejs */
```
jkusa: Can we replace this with:
```js
import require from 'require';

const modules = Object.keys(require.entries);
const builderRegExp = new RegExp(`^${config.modulePrefix}/chart-builders/(.*)`),
const chartBuilderEntries = Object.keys(requirejs.entries).filter(key => builderRegExp.test(key)),

```

packages/core/addon/models/pie-chart.js
```diff
@@ -22,20 +22,46 @@ const Validations = buildValidations(
...
+        }),
+        dependentKeys: ['model._request.dimensions.[]']
+      }
+    )
   },
   {
     //Global Validation Options
-    chartType: DIMENSION_SERIES,
+    chartType: computed('model._request.{dimensions.[],metrics.[],intervals.firstObject.interval}', function() {
+      let request = get(this, 'request');
```
jkusa: const?

packages/core/app/styles/navi-core/components/visualization-config/series-chart.less
jkusa: Do we need this file now, can we just to the standard navi power select styles?

packages/core/addon/manifests/bar-chart.js
daniellevinson: no metrics obviously, or no grouping (one time bucket and no dimensions) otherwise it's a single column (?)
I usually tend to whitelist things rather than blacklist to not have surprises

packages/core/app/styles/navi-core/components/visualization-config/series-chart.less
daniellevinson: I've removed all the power select styles from this file

PR #497, opened 2019-09-06
Demo app cleanup

kevinhinterlong:
## Description
The app would fail sometimes when trying to refresh/share it etc because it couldn't go directly to the url. The data returned was random every time (e.g. the same query), and only basic dimension filters could be applied

## Proposed Changes
- Use `hash` locationType to get around url issues
- Set the seed for `faker` when returning a metric value based on the `[day, dimensions requested, current metric]`, so that we can change `filters`/`havings`/`interval` without changing the data
- Added support for more dimension filters `[in, notin, contains]`
- Added support for havings `[gt, gte, lt, lte, eq, neq, bet, nbet]`

## Screenshots
| 15 days query | 14 days query |
|---------|---------|
| 15 day results        |     14 day results     |

## License

### DIFF
--- a/packages/data/addon/mirage/routes/bard-lite.js
+++ b/packages/data/addon/mirage/routes/bard-lite.js
@@ -14,7 +14,22 @@ const API_DATE_FORMAT = 'YYYY-MM-DD HH:mm:ss.SSS',
   MISSING_INTERVALS = [
     '2018-11-12 00:00:00.000/2018-11-14 00:00:00.000',
     '2018-11-15 00:00:00.000/2018-11-16 00:00:00.000'
-  ];
+  ],
+  HAVING_OPS = {
+    gt: (values, metricValue) => parseFloat(metricValue) > parseFloat(values[0]),
+    gte: (values, metricValue) => parseFloat(metricValue) >= parseFloat(values[0]),
+    lt: (values, metricValue) => parseFloat(metricValue) < parseFloat(values[0]),
+    lte: (values, metricValue) => parseFloat(metricValue) <= parseFloat(values[0]),
+    eq: (values, metricValue) => parseFloat(metricValue) === parseFloat(values[0]),
+    neq: (values, metricValue) => parseFloat(metricValue) !== parseFloat(values[0]),
+    bet: (values, metricValue) => parseFloat(values[0]) < metricValue && metricValue < parseFloat(values[1]),
+    nbet: (values, metricValue) => !(parseFloat(values[0]) < metricValue && metricValue < parseFloat(values[1]))
+  },
+  DIMENSION_OPS = {
+    in: (filterValues, value, field) => filterValues.includes(value[field]),
+    notin: (filterValues, value, field) => !DIMENSION_OPS.in(filterValues, value, field),
+    contains: (filterValues, value, field) => filterValues.some(v => value[field].includes(v))
+  };
 
 /**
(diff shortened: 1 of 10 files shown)


packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -14,7 +14,22 @@ const API_DATE_FORMAT = 'YYYY-MM-DD HH:mm:ss.SSS',
...
-  ];
+  ],
+  HAVING_OPS = {
+    gt: (values, metricValue) => parseFloat(metricValue) > parseFloat(values[0]),
+    gte: (values, metricValue) => parseFloat(metricValue) >= parseFloat(values[0]),
+    lt: (values, metricValue) => parseFloat(metricValue) < parseFloat(values[0]),
+    lte: (values, metricValue) => parseFloat(metricValue) <= parseFloat(values[0]),
+    eq: (values, metricValue) => parseFloat(metricValue) === parseFloat(values[0]),
+    neq: (values, metricValue) => parseFloat(metricValue) !== parseFloat(values[0]),
+    bet: (values, metricValue) => parseFloat(values[0]) < metricValue && metricValue < parseFloat(values[1]),
```
jkusa: Need more `parseFloat` :)

packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -137,10 +167,15 @@ export default function(
         if (currFilter.length > 0) {
           if (currFilter[0] === ',') currFilter = currFilter.substring(1);
 
-          let dimension = currFilter.split('|')[0],
-            values = currFilter.split('[')[1].split(',');
+          const columnOp = currFilter.split('-');
+          const column = columnOp[0].split('|');
+          const op = columnOp[1].split('[');
+          let dimension = column[0],
```
jkusa: `const` all the things

packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -156,27 +191,60 @@ export default function(
...
+    let havings = {};
+    if (request.queryParams.having) {
+      havings = request.queryParams.having.split(']').reduce((havingObj, currHaving) => {
+        if (currHaving.length > 0) {
+          if (currHaving[0] === ',') currHaving = currHaving.substring(1);
+
+          const rowOp = currHaving.split('-');
+          const op = rowOp[1].split('[');
+          const metric = rowOp[0],
+            operation = op[0],
```
jkusa: `const`

packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -156,27 +191,60 @@ export default function(
...
+      }, []);
     });
 
+    let havings = {};
+    if (request.queryParams.having) {
+      havings = request.queryParams.having.split(']').reduce((havingObj, currHaving) => {
+        if (currHaving.length > 0) {
+          if (currHaving[0] === ',') currHaving = currHaving.substring(1);
+
+          const rowOp = currHaving.split('-');
```
jkusa: can we do regex extract?

packages/data/addon/mirage/routes/bard-lite.js
kevinhinterlong: giving up on `const` in order to use regex extraction and make small fixups to `field`and `values`

PR #502, opened 2019-09-10
Fix: Ember 2.x fixes

cythrawll:
## Description

Though tested in a Ember 2.x, the PR changes that were done before the PR merged broke 2.x compatibility.

## Proposed Changes

- Restore Ember 2.x compatibility.

## License

### DIFF
--- a/packages/core/addon/models/fragments/presentation.js
+++ b/packages/core/addon/models/fragments/presentation.js
@@ -5,9 +5,10 @@
 
 import DS from 'ember-data';
 import Fragment from 'ember-data-model-fragments/fragment';
+import { Copyable } from 'ember-copy';
 import { fragmentArray } from 'ember-data-model-fragments/attributes';
 
-export default Fragment.extend({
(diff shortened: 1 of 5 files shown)


packages/core/addon/models/fragments/presentation.js
```diff
@@ -5,9 +5,10 @@
 
 import DS from 'ember-data';
 import Fragment from 'ember-data-model-fragments/fragment';
+import { Copyable } from 'ember-copy';
 import { fragmentArray } from 'ember-data-model-fragments/attributes';
 
-export default Fragment.extend({
+export default Fragment.extend(Copyable, {
```
mattgraham1995: Is this necessary? The Fragment should already have that mixin
https://github.com/lytics/ember-data-model-fragments/blob/master/addon/fragment.js#L70

cythrawll: yeah it's necessary, blew up without it in 2.x app...

PR #507, opened 2019-09-12
Update: Refactor dependencies

jkusa:
# Description

Refactor dependencies

## Proposed Changes

- Move some dev dependencies to normal dependencies
- Remove some unnecessary dependencies
- Move `core` and `data` to dependencies instead of dev dependencies

One thing we might want to do is write back the version numbers that lerna creates during the build. Currently, in git the navi version is `^0.1.0` while we publish with the version `^0.2.0-alpha.47+19a6c1d` for example.

## Screenshots

n/a

## License


packages/directory/package.json
```diff
@@ -25,12 +25,18 @@
...
     "ember-cli-htmlbars": "^3.0.0",
+    "ember-cli-moment-shim": "^3.7.1",
     "ember-data-model-fragments": "~4.0.0",
     "ember-get-config": "^0.2.4",
     "ember-light-table": "^1.13.1",
+    "ember-moment": "^7.8.1",
     "ember-power-select": "^2.3.5",
     "ember-responsive": "^3.0.0-beta.3",
-    "ember-tether": "^1.0.0"
+    "ember-tether": "^1.0.0",
```
jkusa: When rebasing, remove ember-tether

PR #509, opened 2019-09-13
Fix widget nav

cythrawll:
## Description

Dashboards often lose saved state when navigating to other routes

## Proposed Changes

- Move the dashboard rollback to the parent route, that way subroutes like editing widgets can be done without losing dashboard changes.
- When navigating out of parent route, prompt the user if they have unsaved changes.
- Add filter query params to breadcrumb links in widgets

## License

### DIFF
--- a/packages/dashboards/addon/routes/dashboards/dashboard.js
+++ b/packages/dashboards/addon/routes/dashboards/dashboard.js
@@ -80,6 +80,21 @@ export default Route.extend({
     });
   },
 
+  /**
+   * @override
+   * @method resetController - reset query params on exit of route
+   * @param {Object} controller
+   * @param {Boolean} isExiting
+   */
+  deactivate() {
+    this._super(...arguments);
+    const dashboard = this.modelFor(this.routeName);
+    // don't rollback attributes if dashboard was unloaded.
+    if (!dashboard.isEmpty) {
+      dashboard.rollbackAttributes();
+    }
+  },
+
   actions: {
     /**
      * @action didUpdateLayout - updates dashboard's layout property and save it
@@ -166,6 +181,30 @@ export default Route.extend({
       const dashboard = get(this, 'currentDashboard');
       get(dashboard, 'widgets').forEach(widget => widget.rollbackAttributes());
       dashboard.rollbackAttributes();
+    },
+
+    willTransition(transition) {
+      //subroute just continue
+      if (transition.targetName.startsWith('dashboards.dashboard')) {
+        return true;
+      }
+
+      const dashboard = get(this, 'currentDashboard');
+
+      const isDirty =
+        dashboard.get('hasDirtyAttributes') ||
+        dashboard.get('filters.hasDirtyAttributes') ||
+        dashboard.get('presentation.hasDirtyAttributes');
+
+      if (
+        isDirty &&
+        dashboard.get('canUserEdit') &&
+        !confirm('You have unsaved changes, are you sure you want to exit?')
+      ) {
+        transition.abort();
+      } else {
+        return true;
+      }
     }
   }
 });
--- a/packages/dashboards/addon/routes/dashboards/dashboard.js
+++ b/packages/dashboards/addon/routes/dashboards/dashboard.js
@@ -90,7 +90,7 @@ export default Route.extend({
     this._super(...arguments);
     const dashboard = this.modelFor(this.routeName);
     // don't rollback attributes if dashboard was unloaded.
-    if (typeof dashboard.isEmpty !== 'boolean' || !dashboard.isEmpty) {
+    if (dashboard.isEmpty !== true) {
       dashboard.rollbackAttributes();
     }
   },
@@ -183,9 +183,13 @@ export default Route.extend({
       dashboard.rollbackAttributes();
     },
 
+    /**
+     * Prompts user if they are leaving the route with unsaved changes.
+     * @param {Transition} transition
+     */
     willTransition(transition) {
       //subroute just continue
-      if (transition.targetName.startsWith('dashboards.dashboard')) {
+      if (transition.targetName.startsWith(this.routeName)) {
         return true;
       }
 
--- a/packages/dashboards/addon/routes/dashboards/dashboard.js
+++ b/packages/dashboards/addon/routes/dashboards/dashboard.js
@@ -95,6 +95,26 @@ export default Route.extend({
     }
   },
 
+  /**
+   * If traveling to the same route as the current dashboard, cache the query for breadcrumb purposes
+   * @param {Transition} transition
+   */
+  cacheQuery(transition) {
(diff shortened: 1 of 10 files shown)


packages/dashboards/addon/routes/dashboards/dashboard.js
```diff
@@ -166,6 +181,30 @@ export default Route.extend({
       const dashboard = get(this, 'currentDashboard');
       get(dashboard, 'widgets').forEach(widget => widget.rollbackAttributes());
       dashboard.rollbackAttributes();
+    },
+
+    willTransition(transition) {
```
mattgraham1995: jsdoc

packages/dashboards/addon/routes/dashboards/dashboard.js
```diff
@@ -166,6 +181,30 @@ export default Route.extend({
...
+
+      const isDirty =
+        dashboard.get('hasDirtyAttributes') ||
+        dashboard.get('filters.hasDirtyAttributes') ||
+        dashboard.get('presentation.hasDirtyAttributes');
+
+      if (
+        isDirty &&
+        dashboard.get('canUserEdit') &&
+        !confirm('You have unsaved changes, are you sure you want to exit?')
```
mattgraham1995: Is there any way to do a `Save and exit` option?

packages/dashboards/addon/templates/dashboards/dashboard/widgets/widget.hbs
```diff
@@ -9,9 +9,15 @@
     {{/if}}
     {{navi-icon "angle-right"}}
     {{#with (model-for "dashboards.dashboard") as |parentDashboard|}}
-      {{#link-to "dashboards.dashboard" parentDashboard.id class="navi-report-widget__breadcrumb-link"}}
-        {{parentDashboard.title}}
-      {{/link-to}}
+      {{#if parentQueryParams}}
+        {{#link-to "dashboards.dashboard" parentDashboard.id (query-params filters=parentQueryParams.filters) class="navi-report-widget__breadcrumb-link"}}
```
mattgraham1995: Would it work to move the `if` inside the `link-to` helper to just wrap the `query-params`?

packages/dashboards/addon/routes/dashboards/dashboard.js
```diff
@@ -80,6 +80,21 @@ export default Route.extend({
...
+   * @override
+   * @method resetController - reset query params on exit of route
+   * @param {Object} controller
+   * @param {Boolean} isExiting
+   */
+  deactivate() {
+    this._super(...arguments);
+    const dashboard = this.modelFor(this.routeName);
+    // don't rollback attributes if dashboard was unloaded.
+    if (typeof dashboard.isEmpty !== 'boolean' || !dashboard.isEmpty) {
```
jkusa: Would this work:
```js
if (dashboard.isEmpty !== true)
```

packages/dashboards/tests/acceptance/dashboards-test.js
```diff
@@ -395,6 +405,38 @@ module('Acceptance | Dashboards', function(hooks) {
...
+
+    assert.equal(currentURL(), '/dashboards/1/view', 'Taken back to dashboard page');
+
+    assert.deepEqual(
+      widgetsAfter,
+      ['Mobile DAU Goal', 'Mobile DAU Graph', 'Mobile DAU Table', 'Untitled Widget'],
+      '"Untitled Widget" is still on dashboard after navigating to subroute'
+    );
+
+    window.confirm = () => {
```
mattgraham1995: Do we have a test for rolling back the dashboard's unsaved changes through this confirmation window?

packages/dashboards/addon/routes/dashboards/dashboard.js
jkusa: Would this work:
```js
if (dashboard.isEmpty !== true)
```

packages/dashboards/addon/routes/dashboards/dashboard.js
```diff
@@ -166,6 +181,30 @@ export default Route.extend({
       const dashboard = get(this, 'currentDashboard');
       get(dashboard, 'widgets').forEach(widget => widget.rollbackAttributes());
       dashboard.rollbackAttributes();
+    },
+
+    willTransition(transition) {
+      //subroute just continue
+      if (transition.targetName.startsWith('dashboards.dashboard')) {
```
jkusa: Can we do:
```js
if (transition.targetName.startsWith(this.routeName)) {
```

packages/dashboards/addon/routes/dashboards/dashboard.js
cythrawll: Do we want to build a more complicated custom confirmation dialog?

packages/dashboards/addon/templates/dashboards/dashboard/widgets/widget.hbs
cythrawll: no, ember 2.x bad.

packages/dashboards/addon/routes/dashboards/dashboard/widgets/widget.js
```diff
@@ -7,6 +7,27 @@ import { getOwner } from '@ember/application';
...
+    const fromDashboardId = fromRoute ? fromRoute.params.dashboard_id : null;
+    const toRoute = transition.to && transition.to.find(info => info.paramNames.includes('dashboard_id'));
+    const toDashboardId = toRoute ? toRoute.params.dashboard_id : null;
+
+    if (
+      transition.from &&
+      transition.from.name.startsWith('dashboards.dashboard') &&
+      fromDashboardId === toDashboardId
+    ) {
+      controller.set('parentQueryParams', transition.from.queryParams);
```
jkusa: Instead of handing query params in sub routes, could we cache them in the route that has the query params?

packages/dashboards/tests/acceptance/dashboards-test.js
cythrawll: it's covered in a dashboard-filters test

packages/dashboards/addon/routes/dashboards/dashboard/widgets/widget.js
cythrawll: Unfortunately no, that lives in the view route which is not a child of the widget route. The information doesn't need to be cached as it's already given in the transition.  Also the management of cache here would be troublesome. The only reason it's put in the controller is so that the the template has access to it.

jkusa: I was thinking the conditional clearing the query params would be similar to the way you did conditional rollback of the model. This way any child/sibling route would not have to be knowledgeable about the `dashboard.view` route params and have all the `link-to`s correct

PR #514, opened 2019-09-17
Update: Add widget to dashboard

### LINKED ISSUE #485: New Dashboard Widget Flow

daniellevinson:
Resolves #485 

## Description

When adding a report to a dashboard, instead of transitioning to the report view and having to click "add to dashboard", this will add the new widget and stay in the dashboard view.

## Proposed Changes

- add the new widget and stay in dashboard view
- reuse actions of `report-to-widget` mixin
- switch to controller actions instead of route-actions

## Screenshots

Sep-17-2019 15-37-48

## License


packages/dashboards/addon/templates/components/navi-dashboard.hbs
```diff
@@ -100,6 +100,7 @@
         classNames="add-widget primary action"
         reports=(get (get-user) "reports")
         dashboard=dashboard
+        onAddWidgetToDashboard=(action onAddWidgetToDashboard)
```
jkusa: Should this be `onSelectedReport`?

daniellevinson: It's used as the click action handler of the "Add to dashboard" button, so `onAddWidgetToDashboard` is more specific and appropriate imo. The `onchange` of the select only mutates the `selectedReport` property internally.

jkusa: To me `onAddWidgetToDashboard` sounds like the widget was already added to the dashboard, when really the only event that happened was that a report was selected **to be added**.

`onSelect`?

daniellevinson: Oh I see. `onSelect` sounds like an action handler for the select's `onchange`, but I don't have a strong objection here. Let's go with your first suggestion `onSelectedReport`

daniellevinson: updated

PR #516, opened 2019-09-20
Fix: Clone depth

cythrawll:
Resolves #

## Description

builds failing because it can't find the tagged release because the clone depth is too shallow.

## Proposed Changes

- https://docs.travis-ci.com/user/customizing-the-build/#git-clone-depth

## License

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ jdk:
   - openjdk8
 
 sudo: true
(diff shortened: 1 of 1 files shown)


cythrawll:
Yeah I think this makes sense, as default depth 50.  and publish started breaking once the refcount got up to 50.

PR #522, opened 2019-09-23
Fix: Remove and update outdated packages.

cythrawll:
## Description

Some packages were pulling in deprecated build time dependencies that also were bringing in insecure packages. Needed to update and remove those.

## Proposed Changes

- `ember-debounced-properties` is no longer maintained, so replaced it with equivalent code.
- `ember-pluralize` is no longer maintained so replaced it with `ember-inflector`.
- update `ember-font-awesome`
- add eslint rule to remove build time warning.
- added ember/jquery for packages that need it.

## License


packages/core/addon/components/series-selector.js
```diff
@@ -39,9 +39,11 @@ export default Component.extend(DebouncedPropertiesMixin, {
   classNames: ['series-selector'],
 
   /**
-   * @property {Array} debouncedProperties - list of properties to debounce
+   * @property {observer} - Observes searchTerm
    */
-  debouncedProperties: computed(() => ['searchTerm']),
+  searchTermDidChange: observer('searchTerm', function() {
```
jkusa: @colemanProjects please delete this :)

PR #530, opened 2019-09-26
Fix: Remove selector button color

kevinhinterlong:
## Description
The remove buttons for ember-power-select have different styles from our remove button

## Proposed Changes
Copy ember-power-select, now they get darker on hover

## Screenshots
Before:
different colors
After:
same color

## License

### DIFF
--- a/packages/reports/app/styles/navi-reports/vendor/ember-tag-input.less
+++ b/packages/reports/app/styles/navi-reports/vendor/ember-tag-input.less
@@ -19,7 +19,7 @@
     padding: 0;
 
     &-remove {
-      opacity: 0.5;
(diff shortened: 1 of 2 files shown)


packages/reports/app/styles/navi-reports/vendor/ember-tag-input.less
```diff
@@ -19,7 +19,7 @@
     padding: 0;
 
     &-remove {
-      opacity: 0.5;
+      .navi-pill-remove-button;
       cursor: pointer;
```
jkusa: Can remove this? Looks like `.ember-power-select-multiple-remove-btn` has this.

PR #534, opened 2019-09-30
Update: Remove series selector

colemanProjects:
Resolves #

## Description
Removes the series selector. C3 already allows the selection of series

## Proposed Changes
Remove series selector and related components
-

## License


packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1238,26 +1238,25 @@ module('Acceptance | Navi Report', function(hooks) {
   });
 
   test('Updating chart series', async function(assert) {
-    assert.expect(3);
+    assert.expect(4);
```
balajimanoharan: Do we need this test anymore? It is now just testing the c3 feature, rather than something we built.

colemanProjects: That's true. I don't mind either way. Could help as we upgrade c3 though to make sure a series selector feature exists? Though if the team thinks we should remove, I'd be happy to

mattgraham1995: I feel like it's good to keep it so we make sure the functionality is there in Navi.

jkusa:
Does this PR remove the max series limit?

jkusa:
@colemanProjects any update on this?

PR #535, opened 2019-09-30
Fix: Allow save when mustHaveData is toggled (fixes #532)

### LINKED ISSUE #532: Can't save delivery rule when only updating mustHaveData toggle
**Describe the bug**
We added the feature to only send reports via email if there is data. The schedule modal did not allow users to save a delivery rule if toggling the mustHaveData toggle was the only change.

**To Reproduce**
1. Go to '/reports'
2. Click on 'schedule modal'
3. Toggle "mustHaveData" toggle
4. No option to save delivery rule

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

colemanProjects:
Resolves #532

## Description
We added the feature to only send reports via email if there is data. The schedule modal did not allow users to save a delivery rule if toggling the mustHaveData toggle was the only change.

## Proposed Changes
Make schedulingRules an ember data model fragment so that the model passed to the schedule modal actions can actually listen for each rule in the schedulingRules object.

There's a corresponding change to the schedule.js and hbs common actions that depend on this navi-core update
-

## License

### DIFF
--- a/packages/core/addon/models/delivery-rule.js
+++ b/packages/core/addon/models/delivery-rule.js
@@ -5,7 +5,7 @@
 
 import DS from 'ember-data';
 import { validator, buildValidations } from 'ember-cp-validations';
-import { featureFlag } from 'navi-core/helpers/feature-flag';
+import { fragment } from 'ember-data-model-fragments/attributes';
 
 const Validations = buildValidations({
   frequency: [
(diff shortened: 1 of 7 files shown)


packages/core/addon/models/delivery-rule.js
```diff
@@ -34,7 +35,11 @@ export default DS.Model.extend(Validations, {
   updatedOn: DS.attr('moment'),
   deliveryType: DS.attr('string', { defaultValue: 'report' }),
   frequency: DS.attr('string', { defaultValue: 'week' }),
-  schedulingRules: DS.attr({ defaultValue: () => {return featureFlag('enabledNotifyIfData') ? {mustHaveData: false} : {} }}),
+  schedulingRules: fragment('fragments/scheduling-rules', {
+    defaultValue: () => {
```
balajimanoharan: Do we need defaultValue here, if we define defaults in the fragment?

packages/reports/addon/templates/components/common-actions/schedule.hbs
```diff
@@ -87,7 +87,7 @@
             size="small"
             class="schedule-modal__must-have-data-toggle"
             value=localDeliveryRule.schedulingRules.mustHaveData
-            onToggle=(mut localDeliveryRule.schedulingRules.mustHaveData)}}
+            onToggle=(action "toggleMustHaveData")}}
```
balajimanoharan: Can we use `toggle` instead? https://github.com/DockYard/ember-composable-helpers#toggle

packages/core/addon/models/delivery-rule.js
colemanProjects: we discussed this. Removed it from the fragment definition

PR #537, opened 2019-10-01
Update: Namespace loading-message component

### LINKED ISSUE #38: Loader Component
- [x] Need to move `navi-loader` to navi-core
- [ ] Need to rename `loading-message` to `navi-loading-message` and move to `navi-core`

have generic styles for the components so that they aren't overridden everywhere they are used.

balajimanoharan:
Resolves #38 

## Description
- Namespace loading message component

## Proposed Changes
- rename component to `navi-loading-message`
- Did not move to core as specified in the issue since this component is only used in reports

## Screenshots
NA

## License

### DIFF
--- a/packages/core/app/styles/navi-core/common/spinner.less
+++ b/packages/core/app/styles/navi-core/common/spinner.less
@@ -11,13 +11,3 @@
   justify-content: center;
   width: 100%;
 }
-
-.loading-message {
-  .spinner {
-    background-color: transparent;
-    display: inline;
-    height: 30px;
-    margin-right: 5px;
-    width: 150px;
-  }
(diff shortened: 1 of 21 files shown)


packages/core/app/styles/navi-core/common/spinner.less
```diff
@@ -12,7 +12,7 @@
   width: 100%;
 }
 
-.loading-message {
+.navi-loading-message {
```
jkusa: Can this be in `navi-loading-message.less`?

balajimanoharan: done!

PR #540, opened 2019-10-02
Update: Lock down navi dependency versions

jkusa:
Resolves #

## Description

Lock down navi dependency versions

## Proposed Changes

- Set versions to `0.2.0` as that is what we are currently publishing
- Change navi dependencies from floating the minor to a static version

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "navi",
-  "version": "0.1.0",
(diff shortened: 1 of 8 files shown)


cythrawll:
why

jkusa:
> why

@cythrawll can you expand on your question :)

cythrawll:
talked offline

PR #544, opened 2019-10-03
Fix: Dashboard Filters were not being updated by query params

mattgraham1995:
Resolves #

## Description
Dashboard filters are not adding or updating.
It seems the `queryParams` property was moved in the `transition` object in the model hook. It now resides behind `to` and `from` rather than being a top level property.

## Proposed Changes
- Access `to.queryParams`

## License

### DIFF
--- a/packages/dashboards/addon/routes/dashboards/dashboard/view.js
+++ b/packages/dashboards/addon/routes/dashboards/dashboard/view.js
@@ -60,7 +60,7 @@ export default Route.extend({
    */
   async model(_, transition) {
     const dashboard = this.modelFor('dashboards.dashboard');
-    const filterQueryParams = get(transition, 'queryParams.filters');
+    const filterQueryParams = get(transition, 'to.queryParams.filters');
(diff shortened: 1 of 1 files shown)


packages/dashboards/addon/routes/dashboards/dashboard/view.js
```diff
@@ -60,7 +60,7 @@ export default Route.extend({
    */
   async model(_, transition) {
     const dashboard = this.modelFor('dashboards.dashboard');
-    const filterQueryParams = get(transition, 'queryParams.filters');
+    const filterQueryParams = get(transition, 'to.queryParams.filters');
```
jkusa:
```js
const filterQueryParams = transition.to.queryParams.filters;
```

mattgraham1995: I'm going to upgrade all the navi packages to 3.13 next. Hopefully the codemods will get rid of the rest of the `get` and `set` statements

PR #560, opened 2019-10-09
Update Core to Ember 3.12

mattgraham1995:
Blocked by #559 

## Description
Update Core to Ember.12
## Proposed Changes

-

## License

### DIFF
--- a/packages/core/tests/integration/components/navi-visualizations/table-test.js
+++ b/packages/core/tests/integration/components/navi-visualizations/table-test.js
@@ -300,7 +300,7 @@ module('Integration | Component | table', function(hooks) {
 
     let options = merge({}, Options, { showTotals: { subtotal: 'os' } });
 
-    Model[0].response.rows = ROWS.slice(0, 4);
+    set(Model, 'firstObject.response.rows', ROWS.slice(0, 4));
     this.set('model', Model);
     this.set('options', options);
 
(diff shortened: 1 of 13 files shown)


packages/core/tests/integration/components/navi-visualizations/table-test.js
```diff
@@ -337,7 +337,7 @@ module('Integration | Component | table', function(hooks) {
 
     let options = merge({}, Options, { showTotals: { subtotal: 'dateTime' } });
 
-    Model[0].response.rows = ROWS.slice(0, 4);
+    set(Model, 'firstObject.response.rows', ROWS.slice(0, 4));
```
jkusa: What was the issue here?

mattgraham1995: I was getting an error when the tests ran saying that because that property is being observed, it should be set with `Ember.set`

packages/core/.eslintrc.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
     browser: true
   },
   rules: {
-    'multiline-comment-style': ['error', 'starred-block']
+    'multiline-comment-style': ['error', 'starred-block'],
```
kevinhinterlong: Needed?

kevinhinterlong: Maybe come back to `.eslintrc.js` later to make sure they're all the same/strip out unneeded globals

PR #562, opened 2019-10-10
Add blueprints to add default dependencies

jkusa:
Resolves #

## Description

Add blueprints to add default dependencies

## Proposed Changes

Needs #561

## Screenshots

n/a 

## License

### DIFF
--- a/packages/core/blueprints/navi-core/index.js
+++ b/packages/core/blueprints/navi-core/index.js
@@ -1,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  normalizeEntityName() {}, // no-op since we're just adding dependencies
+
+  afterInstall() {
+    return this.addAddonsToProject({
+      packages: [
+        { name: 'navi-notifications' }, // default notification provider
+        { name: 'ember-font-awesome' } // default icon provider
+      ]
+    });
+  }
+};
--- a/packages/reports/blueprints/navi-reports/index.js
+++ b/packages/reports/blueprints/navi-reports/index.js
@@ -1,0 +1,11 @@
+'use strict';
+
+module.exports = {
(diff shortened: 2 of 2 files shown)


packages/reports/blueprints/navi-reports/index.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  normalizeEntityName() {}, // no-op since we're just adding dependencies
+
+  afterInstall() {
+    return this.addAddonsToProject({
+      packages: [{ name: 'navi-core' }]
```
mattgraham1995: Is this to force the core blueprint to run when running reports?

packages/core/blueprints/navi-core/index.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  normalizeEntityName() {}, // no-op since we're just adding dependencies
+
+  afterInstall() {
+    return this.addAddonsToProject({
```
kevinhinterlong: Do addons typically print any messages here? Not necessary but could be nice

packages/reports/blueprints/navi-reports/index.js
jkusa: This allows you to simply do `ember install navi-reports` and get navi-core.

packages/core/blueprints/navi-core/index.js
jkusa: `addAddonsToProject` will print the addons it adds. Did you have something else in mind?

kevinhinterlong: Nope just wanted to check it was printed somewhere

PR #563, opened 2019-10-10
Update: Use ember-set-helper instead of the mut helper

mattgraham1995:
## Description
`mut` is a bit ambiguous in how it actually works under the hood. Read more about the issues with `mut` and how to live without it here https://www.pzuraq.com/on-mut-and-2-way-binding/
Essentially, the common usage of `mut` is not the behavior intended when it was created. The set helper is more explicit in what it is doing. 

## Proposed Changes

- Use `ember-set-helper` to be more explicit about how we set values in our actions

## License

### DIFF
--- a/packages/core/addon/templates/components/number-format-dropdown.hbs
+++ b/packages/core/addon/templates/components/number-format-dropdown.hbs
@@ -6,6 +6,6 @@ terms. --}}
   {{/dd.trigger}}
 
   {{#dd.content class="number-format-dropdown__container"}}
-    {{number-format-selector format=format onUpdateFormat=(action (mut format))}}
+    {{number-format-selector format=format onUpdateFormat=(action (set this "format" _))}}
   {{/dd.content}}
 {{/basic-dropdown}}
--- a/packages/reports/addon/templates/components/dropdown-date-picker.hbs
+++ b/packages/reports/addon/templates/components/dropdown-date-picker.hbs
@@ -8,7 +8,7 @@
     {{navi-date-picker
       date=_selectedDate
       dateTimePeriod="day"
-      onUpdate=(action (mut _selectedDate))
+      onUpdate=(action (set this "_selectedDate" _))
     }}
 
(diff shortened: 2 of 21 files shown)


packages/core/addon/templates/components/number-format-dropdown.hbs
```diff
@@ -6,6 +6,6 @@ terms. --}}
   {{/dd.trigger}}
 
   {{#dd.content class="number-format-dropdown__container"}}
-    {{number-format-selector format=format onUpdateFormat=(action (mut format))}}
+    {{number-format-selector format=format onUpdateFormat=(action (set this "format" _))}}
```
kevinhinterlong: Can we make these
```suggestion
    {{number-format-selector format=format onUpdateFormat=(action (set this.format _))}}
```

mattgraham1995: I also prefer that syntax, but it seems it doesn't work right now.
https://github.com/pzuraq/ember-set-helper/issues/8

kevinhinterlong: Oh :(

mattgraham1995: The first example in the docs using that sort of 2 argument syntax, but the function definition seems to explicitly require 3 arguments. This will have to do for now until that issue is fixed.

packages/reports/addon/templates/components/dropdown-date-picker.hbs
```diff
@@ -8,7 +8,7 @@
     {{navi-date-picker
       date=_selectedDate
       dateTimePeriod="day"
-      onUpdate=(action (mut _selectedDate))
```
pzuraq: Note: You don't need to wrap `(set)` in an `action`, it returns an action, essentially 😄

kevinhinterlong: Thanks! I was curious about that

PR #567, opened 2019-10-14
Update: Enable swagger for navi-webservice

kevinhinterlong:
Partially addresses some [1.0 issues](https://github.com/yahoo/navi/projects/2), by adding swagger docs and building a jar

## Description
- Our webservice should have some documentation around how to use it

## Proposed Changes
- Enables elide-standalone swagger documentation
- Now bundles swagger-ui for easy to consume api
- Builds an easy to use "fat jar" with everything included to run the webservice

---
Try running this and check out the docs
```shell script
./gradlew run
```

and also for a deployable jar
```shell script
./gradlew shadowJar #see packages/webservice/app/build/libs/app-0.2.0-all.jar
```

## License


packages/webservice/app/src/main/kotlin/com/yahoo/navi/ws/app/Settings.kt
```diff
@@ -56,18 +68,58 @@ open class Settings : ElideStandaloneSettings {
...
+     * Runs swagger on <host>/swagger/doc/api
+     */
+    override fun enableSwagger(): Map<String, Swagger> {
+        val dictionary = EntityDictionary(HashMap())
+
+        dictionary.bindEntity(Dashboard::class.java)
+        dictionary.bindEntity(DashboardWidget::class.java)
+        dictionary.bindEntity(Report::class.java)
+        dictionary.bindEntity(User::class.java)
+        val info = Info().title("Navi webservice").version("0.2.0")
```
cythrawll: We already have this entity dictionary in getElideSettings. I'd rather figure out how to use that, rather than attempt to manage building entity dictionaries in different ways.

kevinhinterlong: It looks like that entity dictionary somehow isn't loaded with all the entities so I used the elide util to search for the entities and then add them

PR #568, opened 2019-10-23
Fix: Dashboard filter alt primarykey support

cythrawll:
## Description

Found dashboards break when you use a dimension with alternative primary key.

## Proposed Changes

- Add support for looking up primaryKey field from metadata when filter is added.
- Change dimension display to pass in idField to formatter rather than filter field.

This will make it consistent with how we display dimension values in dropdowns and tag selectors.

## License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ lerna-debug.log
 *.iml
 .idea/
 .vscode
+jsconfig.json
--- a/packages/dashboards/addon/components/dashboard-filters-collapsed-filter.js
+++ b/packages/dashboards/addon/components/dashboard-filters-collapsed-filter.js
@@ -8,6 +8,7 @@ import layout from '../templates/components/dashboard-filters-collapsed-filter';
 import { computed, get, getWithDefault } from '@ember/object';
 import { A as arr } from '@ember/array';
 import { isEmpty } from '@ember/utils';
+import { inject as service } from '@ember/service';
 
 const SUPPORTED_OPERATORS = {
   in: {
@@ -30,6 +31,7 @@ const SUPPORTED_OPERATORS = {
 export default Component.extend({
   layout,
   classNames: ['dashboard-filters-collapsed-filter'],
+  bardMetadata: service(),
 
   /**
    * The computed filter dimension display name.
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ lerna-debug.log
 *.iml
 .idea/
 .vscode
+jsconfig.json
```
kevinhinterlong: Hmm?

cythrawll: vs code configuration file, didn't want to commit it.

kevinhinterlong: They've already been committed, didn't want to push any more changes?

packages/dashboards/addon/components/dashboard-filters-collapsed-filter.js
```diff
@@ -71,6 +73,17 @@ export default Component.extend({
     return getWithDefault(SUPPORTED_OPERATORS, `${op}.longName`, op);
   }),
 
+  /**
+   * Which id field to use as ID display.
+   */
+  filterValueFieldId: computed('filter.{dimension.name,field}', function() {
+    const meta = this.bardMetadata.getById('dimension', get(this, 'filter.dimension.name'));
```
jkusa: Do we need to use the `get` fn in this method?

packages/dashboards/tests/acceptance/dashboard-filter-test.js
```diff
@@ -104,6 +104,42 @@ module('Acceptance | Dashboard Filters', function(hooks) {
...
+    );
+    assert.equal(dataRequests.length, 3, 'three data requests were made (one for each widget)');
+
+    await click('.dashboard-filters__expand-button');
+
+    assert
+      .dom('.dashboard-filters-collapsed-filter__values')
+      .matchesText(/[a-z ]+\(1\)/i, 'filter collapse display should show id and not key');
+
+    return new Promise(resolve => setTimeout(resolve, 5000));
```
jkusa: What is this guy for?

packages/dashboards/addon/components/dashboard-filters-collapsed-filter.js
cythrawll: don't have to, but it does make dealing with null properties much simpler.

packages/dashboards/tests/acceptance/dashboard-filter-test.js
cythrawll: oops that's left over from writing the test. i'll remove

.gitignore
cythrawll: right it has personal setup in it now.

PR #577, opened 2019-10-31
Update: Add support for multiple fact datasources.

cythrawll:
## Description

While we configure fact datasources in navi as an array, the adapter only uses the first defined.
This will allow the service or users of the service to pick which datasource they want to use.

## Proposed Changes

- Add a host method that decides which host to use in the adapter.
- Use `dataSourceName` to choose which datasource to use when using the fact service.

## License

### DIFF
--- a/packages/data/addon/adapters/bard-facts.js
+++ b/packages/data/addon/adapters/bard-facts.js
@@ -1,12 +1,12 @@
 /**
- * Copyright 2018, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Description: The adapter for the bard-facts model.
  */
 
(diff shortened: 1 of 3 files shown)


packages/data/addon/adapters/bard-facts.js
```diff
@@ -174,6 +172,27 @@ export default EmberObject.extend({
...
+   * @param {String} name - name of host to get
+   * @returns {String} - uri of fact datasource to use
+   */
+  _getHost(name) {
+    if (name) {
+      const host = config.navi.dataSources.find(dataSource => dataSource.name === name);
+      if (host && host.uri) {
+        return host.uri;
+      }
+      warn(`Fact host for ${name} requested but none was found in configuration. Falling back to default`, {
```
mattgraham1995: It might be worth adding a test case to make sure the default is used when a non-existent host is passed in to this function.

PR #583, opened 2019-11-04
Bump elide

blitzmohit:
## Description
Bump elide version in webservice package and added/modified necessary packages for the build to work.
## Proposed Changes
- Bump elide version in webservice package
- Publish sources jar for webservice models

## License

### DIFF
--- a/packages/webservice/models/build.gradle.kts
+++ b/packages/webservice/models/build.gradle.kts
@@ -20,13 +20,13 @@ java {
     }
 }
 
-tasks.create<Jar>("sourcesJar") {
+val sourcesJar by tasks. registering(Jar::class) {
     dependsOn(JavaPlugin.CLASSES_TASK_NAME)
     archiveClassifier.set("sources")
     from(sourceSets.main.get().allSource)
 }
 
-artifacts.add("archives", tasks["sourcesJar"])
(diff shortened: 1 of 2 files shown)


packages/webservice/models/build.gradle.kts
```diff
@@ -20,6 +20,14 @@ java {
     }
 }
 
+tasks.create<Jar>("sourcesJar") {
```
kevinhinterlong: Can we change this to the format from [app/build.gradle.kts](https://github.com/yahoo/navi/blob/master/packages/webservice/app/build.gradle.kts#L38)

I think
```suggestion
val sourcesJar by tasks. registering(Jar::class) {
```

packages/webservice/models/build.gradle.kts
```diff
@@ -20,6 +20,14 @@ java {
...
+    dependsOn(JavaPlugin.CLASSES_TASK_NAME)
+    archiveClassifier.set("sources")
+    from(sourceSets.main.get().allSource)
+}
+
+artifacts.add("archives", tasks["sourcesJar"])
```
kevinhinterlong: Then here we should be able to use
```suggestion
artifacts.add("archives", sourcesJar)
```

PR #588, opened 2019-11-09
Update: Update look of sort icon in custom reports (fixes #489)

### LINKED ISSUE #489: Improve Custom Report Sort Icon
**Is your feature request related to a problem? Please describe.**
Customers have reported some confusion regarding how the sort icon behaves. Since it's consisted of two separate arrows, customers think they can press either one and sort either ascending or descending. But in reality, pressing that icon cycles through the different sorting modes.

We could select a different icon to appear next to the sorted columns that makes it more clear to the user what clicking it does.

ldkge:
Resolves #489 

## Description
Update how the sorting icon works in custom reports. Instead of showing both up and down arrows directions, only show the active sorting arrow.

## Proposed Changes

- Show only active icon when sorting

## Screenshots

- Before:
Screen Shot 2019-11-09 at 5 36 29 PM

- After:
Screen Shot 2019-11-09 at 5 45 14 PM

## License

### DIFF
--- a/packages/core/addon/components/navi-table-sort-icon.js
+++ b/packages/core/addon/components/navi-table-sort-icon.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Usage:
--- a/packages/core/addon/components/navi-table-sort-icon.js
+++ b/packages/core/addon/components/navi-table-sort-icon.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019, Yahoo Holdings Inc.
+ * Copyright 2017, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Usage:
--- a/packages/core/addon/templates/components/navi-table-sort-icon.hbs
+++ b/packages/core/addon/templates/components/navi-table-sort-icon.hbs
@@ -1,3 +1,3 @@
-{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
 {{navi-icon "sort-asc" class="navi-table-sort-icon__caret sort-icon__caret-up"}}
 {{navi-icon "sort-desc" class="navi-table-sort-icon__caret sort-icon__caret-down"}}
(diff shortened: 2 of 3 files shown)


packages/core/app/styles/navi-core/components/navi-table-sort-icon.less
```diff
@@ -7,14 +7,20 @@
   display: flex;
   flex-flow: column;
   margin-left: 5px;
+  margin-top: -2px;
```
mattgraham1995: Is this to vertically center the icon? If so, can we just use flex box instead since the `.navi-table-sort-icon` is already a flex container.

packages/core/addon/templates/components/navi-table-sort-icon.hbs
```diff
@@ -1,3 +1,2 @@
 {{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
```
mattgraham1995: Be sure to update the year on the copyright for each file you change

packages/core/app/styles/navi-core/components/navi-table-sort-icon.less
```diff
@@ -7,14 +7,20 @@
...
   &__caret {
     height: 0;
     color: @navi-lighter-gray;
     font-size: @font-size-base;
   }
 
-  &--asc .sort-icon__caret-up,
+  &--asc .sort-icon__caret-down {
+    color: @navi-gray;
+    transform: scaleY(-1);
```
mattgraham1995: I'd prefer if we used an `if` in the template to decide which icon to use instead of applying a transform to flip the caret icon. I think it will be easier for someone in the future to know how it works and also the element class will not have `fa caret-down` in it when it is actually facing upwards.

packages/core/app/styles/navi-core/components/navi-table-sort-icon.less
ldkge: The problem is that the icon itself is not centered, the down arrow is hanging below the center and the up arrow above, centering it won't actually put the arrow in the center.

packages/core/app/styles/navi-core/components/navi-table-sort-icon.less
cythrawll: I don't mind this, but we should change the name of the class to just .sort-icon__caret--asc and .sort-icon__caret--desc

mattgraham1995: The element will still have the font awesome class name in it though, so when you look at the dom, the ascending icon will still have `fa-caret-down` or whatever the icon name is on it even though it is pointing upwards.

balajimanoharan: I agree that would be confusing, if you are just looking at the dom. Can we switch the icon dynamically?

packages/core/addon/components/navi-table-sort-icon.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2019, Yahoo Holdings Inc.
```
balajimanoharan: Can we revert? No changes made in this file.

packages/core/addon/templates/components/navi-table-sort-icon.hbs
```diff
@@ -1,3 +1,3 @@
-{{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
```
balajimanoharan: Same here. Can we revert?

ldkge:
Yes, it should be.

mattgraham1995:
Didn't we talk about removing the `navi-sort-icon` component since we're not displaying both ascending and descending icons simultaneously now?

mattgraham1995:
Closing this as I'll be creating a component from the sort icon used in the request preview.

PR #589, opened 2019-11-11
Rename bard facts model

balajimanoharan:
**BREAKING CHANGE**
-  Rename the bard-facts model to something more appropriate 
- Move the fetchNext and fetchPrevious methods from model to service

## License

### DIFF
--- a/packages/data/addon/models/navi-facts.js
+++ b/packages/data/addon/models/navi-facts.js
@@ -7,37 +7,37 @@
 
 import EmberObject from '@ember/object';
 
-export default EmberObject.extend({
+export default class NaviFacts extends EmberObject {
   /**
    * @property {Object} request - the request object
    */
-  request: undefined,
+  request = undefined;
 
   /**
    * @property {Object} response - response for request
    */
-  response: undefined,
+  response = undefined;
 
   /**
    * @property {Service} _factsService - instance of the facts service passed in on create
    */
-  _factService: undefined,
+  _factService = undefined;
 
   /**
-   * @method next 
+   * @method next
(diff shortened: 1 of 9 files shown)


kevinhinterlong:
Rename to bard to fili as well?

packages/data/addon/models/navi-facts.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright 2017, Yahoo Holdings Inc.
```
mattgraham1995: 2019

packages/data/addon/models/navi-facts.js
```diff
@@ -0,0 +1,43 @@
...
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * Description: A model that holds the response from a Bard query.
+ */
+
+import EmberObject from '@ember/object';
+
+export default EmberObject.extend({
```
jkusa: Can this be an es6 class?

balajimanoharan:
> Rename bard-facts to fili-facts?

@kevinhinterlong the idea is to use the same model for any datasource.`navi-facts` is therefore apt for this reason.

kevinhinterlong:
> > Rename bard-facts to fili-facts?
> 
> @kevinhinterlong the idea is to use the same model for any datasource.`navi-facts` is therefore apt for this reason.

Yea, just pointing out that the linked ticket says

> We have some services named `bard-facts`, `bard-dimensions`, `bard-metadata`. Bard has been renamed to fili, so this could also be renamed appropriately.

so I wasn't sure if the rename of bard -> fili was missed

PR #593, opened 2019-11-15
Update: Use same class for checkbox and dimension/metric labels

daniellevinson:
## Description

Use the same class name for time grain checkboxes and dimension/metric labels - helpful for functional/acceptance tests.

## Proposed Changes

Change `.grouped-list__item-checkbox-label` to `.grouped-list__item-label` and adjust css accordingly.

## License

### DIFF
--- a/packages/reports/app/styles/navi-reports/components/grouped-list.less
+++ b/packages/reports/app/styles/navi-reports/components/grouped-list.less
@@ -60,49 +60,49 @@
       &-container {
         position: relative;
       }
+    }
 
-      &-label {
-        &:before {
-          border-radius: 1px;
-          box-shadow: 0 0 0 2px rgba(54, 151, 242, 0.5);
-          content: '';
-          display: inline-block;
-          height: 8px;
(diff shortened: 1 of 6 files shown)


packages/reports/app/styles/navi-reports/components/grouped-list.less
```diff
@@ -60,49 +60,49 @@
...
+      height: 8px;
+      left: 6px;
+      position: absolute;
+      top: 4px;
+      transform: rotate(45deg);
+      width: 5px;
     }
 
-    &-label {
+    span&-label {
```
kevinhinterlong: Space?
```suggestion
    span &-label {
```

daniellevinson: No, this is supposed to be `span.grouped-list__item-label` as opposed to `label.grouped-list__item-label`

kevinhinterlong: Ah gotcha just looked off to me

PR #600, opened 2019-11-26
Fix: Webservice integration issues.

cythrawll:
## Description

Found a few issues when integrating models

## Proposed Changes

- Add a @JvmField to PermissionExpressions.expressions so that Java code can access the field.
- Need to add a @get field so join table is correct for favorited dashboards
- Absolutely link img assets.

## License


packages/app/app/templates/components/top-bar.hbs
```diff
@@ -1,7 +1,7 @@
-{{!-- Copyright 2017, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{!-- Copyright 2019s, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
```
jkusa: `2019`

PR #602, opened 2019-12-04
Fix: Dashboard fixes

cythrawll:
## Description

Dashboard widgets stopped scrolling

## Proposed Changes

- flex-basis is causing the table widget to be rendered in a overflowed hidden content.
- Add passive hints to table scroll events.

## License

### DIFF
--- a/packages/core/addon/components/table-renderer-vertical-collection.js
+++ b/packages/core/addon/components/table-renderer-vertical-collection.js
@@ -95,11 +95,30 @@ export default Component.extend({
    * @returns {void}
    */
   _registerTableScroll() {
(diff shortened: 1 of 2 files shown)


packages/core/addon/components/table-renderer-vertical-collection.js
```diff
@@ -95,11 +95,30 @@ export default Component.extend({
    * @returns {void}
    */
   _registerTableScroll() {
+    //detect passive feature
+    let supportsPassive = false;
```
jkusa: Can this test be done once outside the component definition?

PR #611, opened 2019-12-07
Update: Collapsible report filters

daniellevinson:
## Description

Collapsible report filters 💥 

## Proposed Changes

It only looks like a lot of files, here's a walkthrough:

- add a new `isFiltersCollapsed` prop to the `report/report` controller which is passed down to `report-view` and `report-builder` and then down to `filter-collection`.
- The reason it's in the parent component is to be able to trigger `resizeVisualization` in `report-view` in order to resize the chart on expand/collapse.
- pass down `setFiltersCollapsed` action to toggle the collapsed state, expand when clicking collapsed (new `click` action in `filter-collection.js`) and expand when adding filters (on `TOGGLE_DIM_FILTER`, `TOGGLE_METRIC_FILTER`, `TOGGLE_PARAMETERIZED_METRIC_FILTER` actions).
- `filter-collection` template has a new `{{#if isCollapsed}}` which renders the same `filter-builders/...` components only with `isCollapsed=true`.
- `filter-builders/...` components also get a new `{{#if isCollapsed}}` which renders a new `filter-builders/collapsed` component.
- `filter-builders/collapsed` renders plain text and the existing `filter-values/...` components with `isCollapsed=true`.
- `filter-values/...` components also get a new `{{#if isCollapsed}}` and render texts.
- a new `filter-values/invalid-value` component for invalid filter values.

## Screenshots

Screen Shot 2019-12-06 at 10 57 46 PM

Gifs coming.

## License


packages/reports/addon/components/filter-builders/collapsed.js
```diff
@@ -0,0 +1,23 @@
...
+ *       displayName=displayName
+ *       filter=filter
+ *       request=request
+ *       field=field
+ *   }}
+ */
+import Component from '@ember/component';
+import layout from 'navi-reports/templates/components/filter-builders/collapsed';
+
+export default Component.extend({
```
kevinhinterlong: Could be a native class
https://github.com/yahoo/navi/blob/289a5225f311da729865a689618852fa17acc1d6/packages/reports/addon/components/navi-request-column-config/base.js#L15-L23

packages/reports/addon/components/filter-values/invalid-value.js
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright 2019, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * Usage:
+ *   {{filter-values/invalid-value}}
+ */
+import Component from '@ember/component';
+import layout from '../../templates/components/filter-values/invalid-value';
+
+export default Component.extend({
```
kevinhinterlong: Native class syntax

packages/reports/addon/components/report-builder.js
```diff
@@ -39,5 +39,21 @@ export default Component.extend({
...
+  }),
+
+  /**
+   * actions
+   */
+  actions: {
+    /**
+     * @action setFiltersCollapsed
+     */
+    setFiltersCollapsed(collapsed) {
```
daniellevinson: I think I'll change to an optional method in the template here

packages/dashboards/addon/templates/dashboards/dashboard/widgets/widget.hbs
```diff
@@ -109,6 +109,8 @@
   <div class="navi-report-widget__body">
     {{#report-builder
       report=model
+      isFiltersCollapsed=isFiltersCollapsed
+      setFiltersCollapsed=(action (set this "isFiltersCollapsed" _))
```
mattgraham1995: I think the `action` wrapper is unnecessary with `ember-set-helper`

mattgraham1995:
Did you consider making the existing dashboard filter components generic and using those? It seems we're doing the same thing twice in two different ways.

daniellevinson:
@mattgraham1995 would it make more sense to make the reports filters generic and re-use in dashboards? some apps may not install dashboards. 
I did take a look at dashboard filters and they seemed different in the logic and in how the filters are constructed, I figured this was why you didn't re-use the ones from reports. Feel free to share what you had in mind when working on this.
We could make the collapsed component generic...

packages/reports/tests/unit/consumers/request/filter-test.js
```diff
@@ -138,7 +140,28 @@ module('Unit | Consumer | request filter', function(hooks) {
...
-    consumer.send(RequestActions.TOGGLE_DIM_FILTER, { currentModel }, 'mockDim');
+    consumer.send(RequestActions.TOGGLE_DIM_FILTER, { currentModel }, expandFiltersMock, 'mockDim');
+  });
+
+  test('TOGGLE_DIM_FILTER - remove dimension', function(assert) {
+    assert.expect(2);
+
+    const dimension = 'mockDim';
+
+    const expandFiltersMock = () => assert.ok(true, 'the expand filters action is not called');
```
kevinhinterlong: force it to fail
```suggestion
    const expandFiltersMock = () => assert.ok(false, 'the expand filters action is not called');
```

packages/reports/tests/unit/consumers/request/filter-test.js
```diff
@@ -203,7 +228,36 @@ module('Unit | Consumer | request filter', function(hooks) {
...
       currentModel = { request: { having: A() } };
 
-    consumer.send(RequestActions.TOGGLE_METRIC_FILTER, { currentModel }, 'mockMetric');
+    consumer.send(RequestActions.TOGGLE_METRIC_FILTER, { currentModel }, expandFiltersMock, 'mockMetric');
+  });
+
+  test('TOGGLE_METRIC_FILTER - remove metric', function(assert) {
+    assert.expect(2);
+
+    const expandFiltersMock = () => assert.ok(true, 'the expand filters action is not called');
```
kevinhinterlong: same
```suggestion
    const expandFiltersMock = () => assert.ok(false, 'the expand filters action is not called');
```

packages/reports/tests/unit/consumers/request/filter-test.js
```diff
@@ -389,13 +445,21 @@ module('Unit | Consumer | request filter', function(hooks) {
...
+      'mockMetric',
+      parameters
+    );
   });
 
   test('TOGGLE_PARAMETERIZED_METRIC_FILTER - remove metric', function(assert) {
     assert.expect(3);
 
-    let parameters = { foo: 'bar' },
+    const expandFiltersMock = () => assert.ok(true, 'the expand filters action is not called');
```
kevinhinterlong:
```suggestion
    const expandFiltersMock = () => assert.ok(false, 'the expand filters action is not called');
```

packages/reports/tests/integration/components/filter-collection-test.js
```diff
@@ -49,19 +49,41 @@ module('Integration | Component | filter collection', function(hooks) {
...
+            isCollapsed=isCollapsed
+            expandFilters=expandFilters
         }}`);
   });
 
-  test('it renders', function(assert) {
+  test('it renders', async function(assert) {
     assert.expect(4);
 
+    this.set('expandFilters', () => assert.ok(true, 'expandFilters is not called on click when not collapsed'));
```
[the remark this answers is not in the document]
daniellevinson: The number of assertions would fail but yeah that's better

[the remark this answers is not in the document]
kevinhinterlong: Yea just don't want someone to skim over it and just bump the assert.expect number

packages/reports/tests/integration/components/filter-values/multi-value-input-test.js
```diff
@@ -27,6 +28,14 @@ module('Integration | Component | filter values/multi value input', function(hoo
     );
   });
 
+  test('collapsed', function(assert) {
+    assert.expect(1);
+
+    this.set('isCollapsed', true);
+
+    assert.dom().hasText('1000', 'Selected values are rendered correctly when collapsed');
```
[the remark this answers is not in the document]
daniellevinson: the collapsed class is passed in but I can check the input doesn't exist

mattgraham1995:
> would it make more sense to make the reports filters generic and re-use in dashboards? some apps may not install dashboards.

This is what I meant. Either one can be made generic so long as we're using the same component/logic in the other. 

The filters attached to a report or a dashboard should be the same filter fragments. A lot of the dashboard filter components just depend on the dashboard having a `filters` property but I don't see why we couldn't just make that a generic model rather than dashboard and use it in reports. There are probably some areas where overrides will be needed, but we can at least set up a base to extend from.

daniellevinson:
Agree @mattgraham1995, thinking if we should mess up this PR or do a separate "refactoring" one.

daniellevinson:
New scary commit af6d05b reuses the entire filters collection in dashboards. Please review 😨 

dashfilters

packages/dashboards/tests/integration/components/dashboard-filters-test.js
```diff
@@ -1,80 +1,218 @@
...
+
+  test('remove a filter', async function(assert) {
+    assert.expect(1);
+
+    this.set('onRemoveFilter', filter => {
+      assert.equal(filter, filterFragment1, 'When clicking remove icon, remove action is sent with selected filter');
+    });
+
+    await click('.dashboard-filters__expand-button');
+    await click(findAll('.filter-collection__remove')[0]);
```
mattgraham1995:
```suggestion
    await click('.filter-collection__remove');
```

packages/reports/addon/components/filter-values/null-input.js
```diff
@@ -20,12 +22,16 @@ export default Component.extend({
...
+    const { isCollapsed, onUpdateFilter } = this;
+
     /*
      * Since this operator doesn't require values, pass empty string
      * here empty string denoted as "" is the same as 'null' in druid
      */
-    this.onUpdateFilter({
-      values: ['""']
-    });
+    if (!isCollapsed && onUpdateFilter && !isEqual(get(this, 'filter.values'), ['""'])) {
```
mattgraham1995:
```suggestion
    if (!isCollapsed && onUpdateFilter && get(this, 'filter.values')[0] !== '""') {
```

packages/reports/addon/consumers/request/filter.js
```diff
@@ -51,14 +51,18 @@ export default ActionConsumer.extend({
...
      */
-    [RequestActions.TOGGLE_DIM_FILTER]: function(route, dimension) {
+    [RequestActions.TOGGLE_DIM_FILTER]: function(route, expandFilters, dimension) {
       let filteredDimensions = get(route, 'currentModel.request.filters'),
         filter = filteredDimensions.findBy('dimension', dimension);
 
       //do not add filter if it already exists
       if (!filter) {
+        if (typeof expandFilters === 'function') {
+          expandFilters();
```
mattgraham1995: Is this expand filters function a visual thing? If that's the case, we should probably keep that out of the consumers as they are generally used for updating models and fragments rather than component properties. Perhaps a better option would be to queue an action that expands the filter collection followed by the consumer action.

packages/reports/addon/controllers/reports/report.js
```diff
@@ -19,6 +19,11 @@ export default Controller.extend({
    */
   showSaveAs: false,
 
+  /**
+   * @property {Boolean} isFiltersCollapsed
+   */
+  isFiltersCollapsed: false,
```
mattgraham1995: Any reason why this needs to be at the route level instead of in the `report-builder` component?

packages/reports/addon/templates/components/filter-values/dimension-select.hbs
```diff
@@ -1,21 +1,33 @@
...
-    <span title={{dimension}}>{{dimension}}</span>
-  {{/with}}
-{{/power-select-multiple}}
+{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{#if this.isCollapsed}}
+  {{#if (is-fulfilled this.selectedDimensions)}}
+    {{#if this.filter.validations.attrs.rawValues.isInvalid}}
+      <FilterValues::InvalidValue/>
+    {{else}}
+      {{#each (await this.selectedDimensions) as |value|}}
```
mattgraham1995: do we need to `await this.selectedDimensions` when `is-fulfilled this.selectedDimensions` is a condition for reaching this line?

packages/reports/addon/templates/components/filter-values/dimension-select.hbs
```diff
@@ -1,21 +1,33 @@
...
+    @search={{action "searchDimensionValues"}}
+    @closeOnSelect={{false}}
+    @placeholder={{concat this.filter.subject.longName " Values"}}
+    @loadingMessage="Loading..."
+    @dropdownClass="filter-values--dimension-select__dropdown"
+    @triggerClass={{concat "filter-values--dimension-select__trigger" (if this.filter.validations.attrs.rawValues.isInvalid " filter-values--dimension-select--error")}}
+    class="filter-values--dimension-select"
+    as |item|
+  >
+    {{#with (format-dimension item) as |dimension|}}
```
mattgraham1995: Might as well change `with` to `let` while we're updating this component.

packages/reports/addon/templates/components/filter-values/range-input.hbs
```diff
@@ -1,18 +1,26 @@
...
 
-{{input
-    value=(readonly filter.values.lastObject)
-    placeholder=highPlaceholder
-    classNames="filter-values--range-input__input"
-    classNameBindings="filter.validations.attrs.values.isInvalid:filter-values--range-input__input--error"
-    key-up=(action "setHighValue")
-}}
\ No newline at end of file
+  {{input
```
mattgraham1995: Angle brackets? 
`<Input @value=.../>`

packages/reports/addon/templates/components/filter-values/value-input.hbs
```diff
@@ -1,8 +1,16 @@
...
-}}
+{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{#if this.isCollapsed}}
+  {{#if this.filter.validations.attrs.values.isInvalid}}
+    <FilterValues::InvalidValue/>
+  {{else}}
+    {{this.filter.values.firstObject}}
+  {{/if}}
+{{else}}
+  {{input
```
mattgraham1995: Same here

packages/reports/addon/templates/components/metric-selector.hbs
```diff
@@ -39,7 +39,7 @@
             @request={{this.request}}
             @onAddParameterizedMetric={{update-report-action "ADD_METRIC_WITH_PARAM"}}
             @onRemoveParameterizedMetric={{update-report-action "REMOVE_METRIC_WITH_PARAM"}}
-            @onToggleParameterizedMetricFilter={{update-report-action "TOGGLE_PARAMETERIZED_METRIC_FILTER"}}
+            @onToggleParameterizedMetricFilter={{update-report-action "TOGGLE_PARAMETERIZED_METRIC_FILTER" this.expandFilters}}
```
mattgraham1995: As mentioned above, we should just queue this `expandFilters` action rather than calling it within the consumer

packages/reports/app/styles/navi-reports/components/filter-collection.less
```diff
@@ -146,4 +146,13 @@
       color: @navi-gray-600;
     }
   }
+
+  &--collapsed {
+    cursor: pointer;
+
+    &-item:not(:last-child):after {
+      content: ';';
```
mattgraham1995: Maybe `.` would be a better separator. It fits better with the plain text style display of the collapsed filters. 
We could also add some whitespace between the last filter value and the next filter builder.

packages/reports/tests/integration/components/filter-builders/dimension-test.js
```diff
@@ -62,12 +62,56 @@ module('Integration | Component | filter-builders/dimension', function(hooks) {
...
     );
     this.owner.register('component:mock/another-values-component', Component.extend());
   });
 
+  test('collapsed', async function(assert) {
+    assert.expect(2);
+
+    await render(hbs`<FilterBuilders::Dimension
+      @filter={{this.filter}}
+      @supportedOperators={{this.upportedOperators}}
```
mattgraham1995: typo `this.upportedOperators`

packages/dashboards/addon/components/dashboard-filters.js
```diff
@@ -18,5 +22,21 @@ export default Component.extend({
...
+  /**
+   * @property {Boolean} isAddingMode
+   */
+  isAddingMode: false,
+
+  /**
+   * @property {Object} filters
+   */
+  filters: computed('dashboard.filters', function() {
+    const {
```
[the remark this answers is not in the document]
daniellevinson: `<ReportBuilder>` expects an object with a `filters` key, could do `return { filters: this.dashboard.filters };` if it's more readable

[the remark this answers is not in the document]
mattgraham1995: Yeah that works. Makes this property more compact.

packages/dashboards/tests/acceptance/dashboard-filter-test.js
```diff
@@ -160,7 +160,7 @@ module('Acceptance | Dashboard Filters', function(hooks) {
       {
         dimension: 'Property',
         operator: 'not equals',
-        rawValues: ['2', '3', '4']
```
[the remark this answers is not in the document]
daniellevinson: because of this `find`
https://github.com/yahoo/navi/blob/master/packages/reports/addon/components/filter-values/dimension-select.js#L78-L81
I'll make an issue

packages/reports/addon/components/filter-values/null-input.js
daniellevinson: wanted to cover a non-array case

packages/reports/addon/consumers/request/filter.js
daniellevinson: yeah, I didn't like it in the consumers as well, even though checking for the dimension existence is already there. I'll move it somewhere else

packages/reports/addon/controllers/reports/report.js
daniellevinson: it's passed down to `report-view` as well to trigger the chart resize when changing the collapsed state

packages/reports/addon/templates/components/filter-values/dimension-select.hbs
daniellevinson: yep, `this.selectedDimensions` is still a promise

packages/reports/addon/templates/components/report-builder.hbs
```diff
@@ -17,35 +17,52 @@
...
+        (if this.isFiltersCollapsed " report-builder__container--filters--collapsed")
+      }}>
       <span class="report-builder__container-header">
-        <h3 class="report-builder__container-header__title">Filters</h3>
-        <span class="report-builder__container-header__message">
-          Click the icon {{navi-icon "filter" class="report-builder__container-header__message--icon"}} next to a dimension/metric to add a filter.
+        <span
+          class="report-builder__container-header__filters-toggle"
+          role="button"
+          onclick={{action (optional this.onUpdateCollapsed) (not this.isFiltersCollapsed)}}>
```
kevinhinterlong: I think this also works and we should try to use it going forward
```suggestion
          {{on "click" (fn (optional this.onUpdateCollapsed) (not this.isFiltersCollapsed))}}>
```

see [`on` rfc](https://github.com/emberjs/rfcs/blob/master/text/0471-on-modifier.md) and [`fn` rfc](https://github.com/emberjs/rfcs/blob/master/text/0470-fn-helper.md)

kevinhinterlong:
Can we also enable `enableDashboardFilters` for the demo app?
https://github.com/yahoo/navi/blob/82eaecee49da0ec1109babb7b5c62ef691372fa4/packages/app/config/environment.js#L28-L31
```diff
      FEATURES: {
        enableMultipleExport: false,
        enableTableEditing: true,
+       enableDashboardFilters: true
      }
```

packages/reports/addon/components/report-builder.js
```diff
@@ -39,5 +40,55 @@ export default Component.extend({
...
     return metadataService.all('table').sortBy('longName');
-  })
+  }),
+
+  actions: {
+    /**
+     * @action expandFilters
+     * @param {Function} shouldExpand
+     */
+    expandFilters(shouldExpand) {
```
mattgraham1995: It doesn't look like `expandFilters` is being invoked from a template anywhere. Does it need to be an action?

packages/reports/addon/components/filter-values/invalid-value.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Usage:
+ *   <FilterValues::InvalidValue/>
+ */
+import Component from '@ember/component';
+import { layout as templateLayout, tagName, classNames } from '@ember-decorators/component';
+import layout from '../../templates/components/filter-values/invalid-value';
+
+@classNames('filter-values--selected-error')
+@tagName('span')
```
jkusa: Forward-looking, can this be a tagless component? Moving to glimmer components we should be able to simply remove this file

packages/reports/addon/components/report-builder.js
```diff
@@ -39,5 +40,55 @@ export default Component.extend({
...
+      this.send('expandFilters', () => arr(this.request.filters).findBy('dimension', dimension));
+    },
+
+    /**
+     * @action onToggleMetricFilter
+     * @param {Object} metric
+     */
+    onToggleMetricFilter(metric) {
+      this.send('expandFilters', () =>
+        arr(this.request.having).find(having => get(having, 'metric.metric.name') === get(metric, 'name'))
```
mattgraham1995: We can remove the `arr` wrapper or change this to use `findBy`

`arr(this.request.having).findBy('metric.metric.name', get(metric, 'name'))`

packages/reports/addon/components/report-builder.js
```diff
@@ -39,5 +40,55 @@ export default Component.extend({
...
+    },
+
+    /**
+     * @action onToggleParameterizedMetricFilter
+     * @param {Object} metric
+     * @param {Object} parameters
+     */
+    onToggleParameterizedMetricFilter(metric, parameters) {
+      this.send('expandFilters', () =>
+        arr(this.request.having).find(
```
mattgraham1995: Same as above

packages/dashboards/addon/components/dashboard-filters.js
```diff
@@ -18,5 +22,17 @@ export default Component.extend({
...
+  /**
+   * @property {Boolean} isAddingMode
+   */
+  isAddingMode: false,
+
+  /**
+   * @property {Object} filters
+   */
+  filters: computed('dashboard.filters', function() {
+    return { filters: this.dashboard.filters };
```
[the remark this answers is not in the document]
daniellevinson: `<FilterCollection>` expects `{ filters: [...], intervals: [...], having: [...] }` in the `request` argument

packages/reports/addon/templates/components/filter-builders/base.hbs
```diff
@@ -1,24 +1,33 @@
...
-    filter=filter
-    request=request
-    onUpdateFilter=onUpdateFilter
-  }}
-</span>
+{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{#if this.isCollapsed}}
+  <FilterBuilders::Collapsed
+    @displayName={{this.displayName}}
+    @filter={{this.filter}}
```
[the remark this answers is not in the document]
daniellevinson: it's a cp here but I've changed in other templates

PR #613, opened 2019-12-10
Update: Disable column reordering inside navi widget

kevinhinterlong:
## Description
When using a table widget on a dashboard, dragging the name would attempt to reorder the column, but would also move the widget itself

## Proposed Changes
- Update ember-sortable
- Disable dragging when inside of a widget

## Screenshots
before after

## License

### DIFF
--- a/packages/core/addon/components/navi-visualizations/table.js
+++ b/packages/core/addon/components/navi-visualizations/table.js
@@ -74,6 +74,10 @@ export default Component.extend({
    */
   rowsInResponse: alias('model.firstObject.response.rows.length'),
 
+  get isDraggingDisabled() {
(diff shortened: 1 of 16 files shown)


daniellevinson:
also need to change in `table-renderer-vertical-collection`

packages/core/addon/components/navi-visualizations/table.js
```diff
@@ -74,6 +74,10 @@ export default Component.extend({
    */
   rowsInResponse: alias('model.firstObject.response.rows.length'),
 
+  get isDraggingDisabled() {
```
mattgraham1995: jsdoc

PR #619, opened 2019-12-18
Navi core deprecation warnings

### LINKED ISSUE #398: Fix deprecation warnings in core
Use ember-cli-deprecation-workflow to narrow down the source of the warnings

mattgraham1995:
Resolves #398 

## Description
Solves most of the deprecation warnings in the navi-core package. A few are silenced in the dummy app because they are caused by dependencies with no quick work arounds for now.

## Proposed Changes

- Remove observers where possible by using lifecycle hooks to detect changes to a given property and recompute accordingly
- Update ember data to match ember version for all packages
- Remove usage of `sendAction` internally
- Remove usage of `this.$` or `Ember.$` unless needed then replace with `import $ from jquery`
- Remove computed on properties that need to be set
- Replace volatile computed properties with get functions
- Avoid moment deprecation warning by checking if input is a moment in the format-date helper 
- Update tests to avoid overwriting computed properties on instantiation

## Screenshots
Unresolved deprecation warnings are marked with a comment on the cause
Screen Shot 2019-12-18 at 1 47 08 PM

## License


packages/core/tests/unit/components/pick-container-test.js
```diff
@@ -70,13 +79,18 @@ module('Unit | Component | pick container', function(hooks) {
...
     component.send('stageChanges', 2);
     assert.equal(component.get('_editableSelection'), 2, 'Staged changes are visible when selection remains the same');
 
     run(() => {
+      /**
+       * Removal of observer in component now depends on a lifecycle hook that isn't called
+       * without a component render so we force the hook to run
+       */
       component.set('selection', 3);
+      component.didReceiveAttrs();
```
kevinhinterlong: `updateSelection`?

Do these need to be in `run()`

packages/core/tests/unit/models/deliverable-item-test.js
```diff
@@ -15,11 +15,15 @@ module('Unit | Model | deliverable item', function(hooks) {
...
+    })
+  );;
 
     assert.equal(model.get('modelId'), tempId, 'getId returns the tempId of the model when id is absent');
-
-    model.set('id', id);
-
-    assert.equal(model.get('modelId'), id, 'getId returns the id when present');
+  
+    assert.equal(model2.get('modelId'), id, 'getId returns the id when present');
```
kevinhinterlong: This changes the behavior of the test, is there no easy way to mock `id`?

packages/core/tests/integration/components/navi-loader-test.js
```diff
@@ -13,14 +13,11 @@ module('Integration | Component | navi loader', function(hooks) {
           containerClass='test-container'
       }}`);
 
-    assert.ok(this.$('.navi-loader__container').is(':visible'), 'The loader container is rendered');
+    assert.dom('.navi-loader__container').exists('The loader container is rendered');
```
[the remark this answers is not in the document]
mattgraham1995: `isVisible` won't always pass. It checks for a positive value for the height and width of the element. For some reason the testing environment will render the loader container with a height of 0 sometimes. I think confirming the loader is in the dom is sufficient.

packages/core/tests/integration/components/navi-vis-c3-chart-test.js
```diff
@@ -68,15 +69,9 @@ module('Integration | Component | navi vis c3 chart', function(hooks) {
...
-      assert.ok(
-        this.$('.c3-chart-line:eq(0)').is('.chart-series-0'),
-        'Each chart series has a corresponding class applied'
-      );
-
-      assert.ok(
-        this.$('.c3-chart-line:eq(1)').is('.chart-series-1'),
-        'Each chart series has a corresponding class applied'
-      );
+      assert.ok($('.c3-chart-line:eq(0)').is('.chart-series-0'), 'Each chart series has a corresponding class applied');
```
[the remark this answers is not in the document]
mattgraham1995: hasClass requires a match on the entire list of classes, but for some reason it's appending whitespace to the result so it never matches even if I had whitespace to the expected string. There's a regex option, but it seems to be forcing an anchor and failing. I think it's easier to just continue using jquery for now.

packages/core/tests/unit/models/deliverable-item-test.js
mattgraham1995: I don't remember why this changed. It seems to pass now though so I'll change it back.

packages/core/tests/integration/components/navi-visualizations/table-test.js
```diff
@@ -407,12 +408,12 @@ module('Integration | Component | table', function(hooks) {
     await render(TEMPLATE);
 
     assert.ok(
-      this.$('.table-header-cell:contains(Platform Revenue) .navi-table-sort-icon--desc').is(':visible'),
+      $('.table-header-cell:contains(Platform Revenue) .navi-table-sort-icon--desc').is(':visible'),
       'The right sort metric is recognized from the alias'
     );
```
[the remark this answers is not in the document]
mattgraham1995: Screen Shot 2019-12-26 at 4 56 42 PM

Getting errors for some reason when passing an element to assert.dom

packages/core/tests/unit/components/pick-container-test.js
```diff
@@ -45,7 +55,7 @@ module('Unit | Component | pick container', function(hooks) {
     );
 
     run(() => {
-      component.set('selection', 3);
+      UpdateSelection(component, 3);
     });
```
kevinhinterlong: Does this need to be in `run`?

packages/core/tests/unit/models/deliverable-item-test.js
kevinhinterlong: Still kinda changes the behavior of the test, originally we
1. Made model with tempId, check it is tempId
2. Set the actual id
3. Check it is now actual id

Now we just check that if we construct a model with a temp id and an id it takes the id

mattgraham1995: In my previous comment I meant I would revert this line to the previous behavior.

PR #623, opened 2020-01-08
Update: Support for multi datasource metadata

### LINKED ISSUE #615: Inconsistent methods to fetch uri from config
Related to Addon: **Navi Data**

We have multiple ways within the addons to fetch the uri's from the config.
The metadata adapter looks at the first object in the array
https://github.com/yahoo/navi/blob/master/packages/data/addon/adapters/bard-metadata.js#L11

The facts adapter looks for a specific name if available to find the uri https://github.com/yahoo/navi/blob/master/packages/data/addon/adapters/bard-facts.js#L184

We should standardize the way the uri is fetched from the config.

cythrawll:
Resolves #615

## Description 

Need to support multiple datasupport.  Allows app to load metadata from multiple datasources and annotate each record in the keg from which datasource the metadata was retreived from.

Also unifies how the adapters identify which host to use.

## Proposed Changes

- Add host selection support for each adapter.
- change metadata service and serializer to load more than one datasource and annotate each record with 'source' pointing to where the metadata came from.

## License


packages/data/tests/unit/utils/adapter-test.js
```diff
@@ -0,0 +1,51 @@
...
+  test('configHost gets correct source object depending on configuration', function(assert) {
+    assert.equal(
+      configHost({ dataSourceName: 'dummy' }),
+      'https://data.naviapp.io',
+      'dummy host is correctly configured'
+    );
+    assert.equal(
+      configHost({ dataSourceName: 'blockhead' }),
+      'https://data2.naviapp.com',
+      'blockhead is correclty configured'
```
mattgraham1995: typo

packages/data/addon/services/bard-metadata.js
```diff
@@ -74,16 +80,18 @@ export default Service.extend({
...
             //create metadata model objects and load into keg
-            this._loadMetadataForType('table', metadata.tables);
-            this._loadMetadataForType('dimension', metadata.dimensions);
-            this._loadMetadataForType('metric', metadata.metrics);
+            this._loadMetadataForType('table', metadata.tables, options.dataSourceName);
+            this._loadMetadataForType('dimension', metadata.dimensions, options.dataSourceName);
+            this._loadMetadataForType('metric', metadata.metrics, options.dataSourceName);
 
             this.set('metadataLoaded', true);
+            this.loadedDataSources.push(options.dataSourceName);
```
jkusa: Can we add a `namespace` option to the other accessor public methods (`all`, `getById`, etc)?

packages/data/addon/services/keg.js
```diff
@@ -77,21 +78,22 @@ export default class KegService extends Service {
     const factory = this._getFactoryForType(options.modelFactory || type);
     const recordKeg = this._getRecordKegForType(type);
     const idIndex = this._getIdIndexForType(type);
+    const namespace = options.namespace || getDefaultDataSourceName();
```
jkusa: Instead of teaching the key about data sources (right now it is a general service), can we have a generic default namespace name and gave the keg clients do the defaulting to the `getDefaultDataSourceName()`

packages/data/tests/unit/services/bard-metadata-test.js
```diff
@@ -65,6 +65,76 @@ module('Unit - Service - Bard Metadata', function(hooks) {
     assert.ok(Service.metadataLoaded, 'metadataLoaded property is set to true after data is loaded');
   });
 
+  test('loadMetadata from multiple sources', async function(assert) {
+    metadataRoutes.bind(Server)(1);
+    await Service.loadMetadata({ dataSourceName: 'dummy' });
+    await Service.loadMetadata({ dataSourceName: 'blockhead' });
+    let keg = Service._keg;
```
jkusa: Can we test this with the public methods?

packages/data/addon/services/bard-metadata.js
```diff
@@ -38,6 +38,11 @@ export default Service.extend({
    */
   metadataLoaded: false,
 
+  /**
+   * @property {Array} loadedDataSources - list of data sources in which meta data has already been loaded
+   */
+  loadedDataSources: null,
```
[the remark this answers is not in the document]
cythrawll: Is this comment on the right line?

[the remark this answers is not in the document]
jkusa: It isn't :) This comment was for line 39. This line (44) seems like a replacement for 39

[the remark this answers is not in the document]
cythrawll: we can get rid of it eventually right now that property is used widely as part of the public API. Best we can do is mark it as deprecated.

packages/data/addon/services/bard-metadata.js
cythrawll: Does all need it? I can add it to the others for sure.

packages/data/addon/services/bard-metadata.js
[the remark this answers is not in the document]
jkusa: @cythrawll Ah, where do we use it? I couldn't find any references outside this service.

If it is used, can we computed it off `loadedDataSources`? Right now it looks like a single data source would change this to true instead of indicating all metadata is loaded.

packages/data/addon/services/keg.js
```diff
@@ -77,21 +82,22 @@ export default class KegService extends Service {
...
+      const existingRecord = this.getById(type, id, namespace);
 
       if (existingRecord) {
         setProperties(existingRecord, rawRecords[i]);
         returnedRecords.pushObject(existingRecord);
       } else {
         let newRecord = factory.create(Object.assign({}, owner.ownerInjection(), rawRecords[i]));
 
-        idIndex[id] = newRecord;
+        idIndex[`${namespace}.${id}`] = newRecord;
```
kevinhinterlong: would it be worth it to make keg have `namespace > type > id`?

cythrawll: I think that would be a big refactor, not sure if it's worth it right now.

packages/data/addon/services/bard-metadata.js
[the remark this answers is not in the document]
cythrawll: I could have sworn we used in the apps itself, seems like i'm wrong. Yoink it goes.

packages/data/addon/services/bard-metadata.js
```diff
@@ -129,7 +121,7 @@ export default Service.extend({
    */
   all(type, namespace) {
     assert('Type must be table, metric or dimension', A(['table', 'dimension', 'metric']).includes(type));
-    assert('Metadata must be loaded before the operation can be performed', get(this, 'metadataLoaded'));
+    assert('Metadata must be loaded before the operation can be performed', this.loadedDataSources.length > 0);
```
jkusa: Could this be
```suggestion
    assert('Metadata must be loaded before the operation can be performed', this.loadedDataSources.includes(namespace));
```

PR #625, opened 2020-01-08
New: Added search addon (refs #624)

### LINKED ISSUE #624: Add Navi Search
Add Search functionality to Navi

ldkge:
Resolves #624

## Description
Add structure for Navi Search feature

## Proposed Changes

- Added folder structure and placeholder components for navi-search

## License


packages/search/addon/components/navi-search-bar.js
```diff
@@ -0,0 +1,6 @@
+import Component from '@ember/component';
+import layout from '../templates/components/navi-search-bar';
+
+export default Component.extend({
```
balajimanoharan: Can we octaneify everything?

ldkge: It's octanified  now!

packages/search/addon/components/navi-search-bar.js
```diff
@@ -0,0 +1,3 @@
+import Component from '@glimmer/component';
+
+export default class NaviSearchBarComponent extends Component {}
```
kevinhinterlong: Make sure we can still override these in the app, you may or may not have to use [`setComponentTemplate`](https://github.com/emberjs/rfcs/blob/master/text/0481-component-templates-co-location.md#low-level-primitives) on the component in order to be able to override it

ldkge: I think we won't need to worry about this, the docs say 

> The low-level setComponentTemplate and getComponentTemplate APIs are not intended to be called by end-users directly, even though they will be public and part of our semver stability guarantee. Instead, they are meant to be used primarily in the output emitted by build tools. At build time, any template files found in {app,addon}/components will be inlined into the component's JavaScript file and removed from the build output.

If the subclassed component has a template with the same name it should pick this one up instead of the inherited template.

packages/search/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "doc": "doc",
+    "test": "tests"
+  },
+  "scripts": {
+    "build": "ember build",
+    "lint:hbs": "ember-template-lint .",
+    "lint:js": "eslint .",
+    "start": "ember serve",
+    "test": "ember test",
+    "test:all": "ember try:each"
```
balajimanoharan: Can we update these scripts to be similar to the other repos?

.travis.yml
```diff
@@ -28,6 +28,7 @@ env:
     - PACKAGE=navi-directory
     - PACKAGE=navi-notifications
     - PACKAGE=navi-reports
+    - PACKAGE=navi-search
```
balajimanoharan: Can we not publish search until it is ready?

packages/search/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "navi-search",
+  "version": "0.2.0",
+  "description": "Search for navi",
```
kevinhinterlong:
```suggestion
  "description": "Search for navi",
  "private": true,
```

Make it private so that it builds but doesn't publish

packages/search/addon/components/navi-search-bar.js
kevinhinterlong: I ended up having issues with template co-location/glimmer components when using it in an addon since glimmer components don't have a `layout` field like classic ember components. Although since these have no implementation I guess they have to be extended anyways so maybe not an issue

.travis.yml
kevinhinterlong: Keep this, but make it private in `package.json`

ldkge: will just marking it as private in the package json work or is there another option in the travis.yml that needs to be updated?

PR #627, opened 2020-01-10
Fix: New widget is blank if saved before running

daniellevinson:
## Description

Adding a new widget and saving before running is resulting in a blank widget. 
An invalid widget visualization config is not dealt with in `dashboard/view`.

## Proposed Changes

Wait for `runReport` to resolve before saving - the visualization config will be rebuilt in the widget view before redirecting to the dashboard.
Also fix invalid config of reports fixtures in app (breaks when adding to dashboard).

## License


packages/dashboards/tests/acceptance/dashboards-test.js
```diff
@@ -383,6 +383,12 @@ module('Acceptance | Dashboards', function(hooks) {
...
 
+    assert.deepEqual(
+      findAll('.navi-report-widget__footer .btn').map(e => e.textContent.trim()),
+      ['Run'],
+      'Before running a new widget, only the "Run" button is shown ("Save" and "Revert" are hidden)'
+    );
+
     // Fill out request
     await click($('.checkbox-selector--metric .grouped-list__item:contains(Total Clicks) .grouped-list__add-icon')[0]);
     await click('.navi-report-widget__run-btn');
```
kevinhinterlong: Copy
```js
    assert.deepEqual(
      findAll('.navi-report-widget__footer .btn').map(e => e.textContent.trim()),
      ['Run'],
      'Before running a new widget, only the "Run" button is shown ("Save" and "Revert" are hidden)'
    );
```
but check for expected buttons after running

packages/app/tests/acceptance/print-views-test.js
```diff
@@ -15,7 +15,7 @@ module('Acceptance | print views', function(hooks) {
     assert.dom('.navi-report__title').hasText('Kart Wins By Character', 'The report title shows up');
     assert.deepEqual(
       [...document.querySelectorAll('.line-chart-widget .c3-legend-item')].map(el => el.textContent),
```
kevinhinterlong: Pretty sure I wrote this, should probably just use `findAll`
```suggestion
      findAll('.line-chart-widget .c3-legend-item').map(el => el.textContent),
```

packages/dashboards/tests/acceptance/dashboards-test.js
daniellevinson: only the "save" should be added ("revert" is for already saved widgets) and we're clicking it right below

packages/dashboards/addon/controllers/dashboards/dashboard/widgets/widget.js
```diff
@@ -1,12 +1,17 @@
...
  */
 import ReportController from 'navi-reports/controllers/reports/report';
 import { computed } from '@ember/object';
 import { inject as controller } from '@ember/controller';
 
 export default ReportController.extend({
+  /**
+   * @property {Boolean} showSaveWidget - whether the save button is rendered (hidden for new widgets before first run)
+   */
+  showSaveWidget: false,
```
mattgraham1995: `showSaveButton`?

packages/dashboards/addon/templates/dashboards/dashboard/widgets/widget.hbs
```diff
@@ -131,7 +131,7 @@
         Run
       </button>
 
-      {{#if (and model.dashboard.canUserEdit model.hasDirtyAttributes)}}
+      {{#if (and model.dashboard.canUserEdit model.hasDirtyAttributes (or showSaveWidget (not model.isNew)))}}
```
mattgraham1995: It looks like clicking `Save` already fires off the `runReport` action. Maybe there's a way to have the `saveWidget` route action wait for the `runReport` action to resolve first.

daniellevinson: Thanks for that, `transitionTo` does resolve after the `model` resolves so all we missed was a `return`. 🤦‍♂ 
It still however doesn't fix adding existing new reports with bad vis config since we don't go through the view route.

packages/reports/addon/routes/reports/report.js
```diff
@@ -126,7 +126,7 @@ export default Route.extend({
      * transition to view subroute if runReport is not handled in subroutes
      */
     runReport(report) {
-      this.transitionTo(`${this.routeName}.view`, get(report, 'tempId') || get(report, 'id'));
+      return this.transitionTo(`${this.routeName}.view`, get(report, 'tempId') || get(report, 'id'));
```
mattgraham1995: Don't forget the ©️ for this file

PR #631, opened 2020-01-13
Fix: Metric filter input

daniellevinson:
## Description

Fix the input for metric filters.

## Proposed Changes

For one-way binding we should use `<input>` with the `value` property and the `input` event, i.e.:
<strike>`<Input @value={{...}} @key-up={{action "setValue"}} />`</strike>
`<input value={{...}} {{on "input" (action this.setValue value="target.value")}}>`.

## License

### DIFF
--- a/packages/reports/addon/templates/components/filter-values/range-input.hbs
+++ b/packages/reports/addon/templates/components/filter-values/range-input.hbs
@@ -20,7 +20,7 @@
       @value={{@filter.values.lastObject}}
       @placeholder={{this.highPlaceholder}}
       class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
-      {{on "input" (action "setHighValue" value="target.value")}}
+      {{on "input" (action this.setHighValue value="target.value")}}
     />
   </div>
 {{/if}}
--- a/packages/reports/addon/templates/components/filter-values/range-input.hbs
+++ b/packages/reports/addon/templates/components/filter-values/range-input.hbs
@@ -1,4 +1,4 @@
-{{!-- Copyright 2019, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+{{!-- Copyright 2020, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
 {{#if @isCollapsed}}
   {{#if @filter.validations.attrs.values.isInvalid}}
     <FilterValues::InvalidValue/>
@@ -6,19 +6,21 @@
     {{concat @filter.values.firstObject " and " @filter.values.lastObject}}
   {{/if}}
 {{else}}
-  <Input
-    @value={{readonly @filter.values.firstObject}}
-    @placeholder={{this.lowPlaceholder}}
-    class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
-    @key-up={{action "setLowValue"}}
-  />
+  <div class="filter-values--range-input">
+    <Input
+      @value={{@filter.values.firstObject}}
+      @placeholder={{this.lowPlaceholder}}
+      class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
+      {{on "input" (action "setLowValue" value="target.value")}}
+    />
 
-  <span class="filter-values--range-input__separator">and</span>
+    <span class="filter-values--range-input__separator">and</span>
 
-  <Input
-    @value={{readonly @filter.values.lastObject}}
-    @placeholder={{this.highPlaceholder}}
-    class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
-    @key-up={{action "setHighValue"}}
-  />
+    <Input
+      @value={{@filter.values.lastObject}}
+      @placeholder={{this.highPlaceholder}}
+      class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
+      {{on "input" (action "setHighValue" value="target.value")}}
+    />
+  </div>
 {{/if}}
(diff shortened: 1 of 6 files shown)


packages/reports/addon/templates/components/filter-values/range-input.hbs
```diff
@@ -1,24 +1,26 @@
...
     {{concat @filter.values.firstObject " and " @filter.values.lastObject}}
   {{/if}}
 {{else}}
-  <Input
-    @value={{readonly @filter.values.firstObject}}
-    @placeholder={{this.lowPlaceholder}}
-    class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
-    @key-up={{action "setLowValue"}}
-  />
+  <div class="filter-values--range-input">
```
kevinhinterlong:
```suggestion
  <div class="filter-values--range-input" ...attributes>
```

packages/reports/addon/templates/components/filter-values/range-input.hbs
```diff
@@ -1,24 +1,26 @@
...
+    <Input
+      @value={{@filter.values.firstObject}}
```
kevinhinterlong: Did readonly cause an issue?

packages/reports/addon/templates/components/filter-values/range-input.hbs
```diff
@@ -1,24 +1,26 @@
...
+      @placeholder={{this.lowPlaceholder}}
+      class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
+      {{on "input" (action "setLowValue" value="target.value")}}
```
kevinhinterlong:
```suggestion
      {{on "input" (action this.setLowValue value="target.value")}}
```

packages/reports/addon/templates/components/filter-values/range-input.hbs
```diff
@@ -1,24 +1,26 @@
...
-    @value={{readonly @filter.values.lastObject}}
-    @placeholder={{this.highPlaceholder}}
-    class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
-    @key-up={{action "setHighValue"}}
-  />
+    <Input
+      @value={{@filter.values.lastObject}}
+      @placeholder={{this.highPlaceholder}}
+      class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
+      {{on "input" (action "setHighValue" value="target.value")}}
```
kevinhinterlong:
```suggestion
      {{on "input" (action this.setHighValue value="target.value")}}
```

packages/reports/addon/templates/components/filter-values/value-input.hbs
```diff
@@ -7,9 +7,9 @@
   {{/if}}
 {{else}}
   <Input
-    @value={{readonly @filter.values.firstObject}}
+    @value={{@filter.values.firstObject}}
     @placeholder="Value"
     class="filter-values--value-input {{if @filter.validations.attrs.values.isInvalid "filter-values--value-input--error"}}"
-    @key-up={{action "setValue"}}
+    {{on "input" (action "setValue" value="target.value")}}
```
kevinhinterlong:
```suggestion
    {{on "input" (action this.setValue value="target.value")}}
```

packages/reports/addon/templates/components/filter-values/range-input.hbs
daniellevinson: does it matter when using the action helper?

kevinhinterlong: not really, I just prefer to directly pass the action instead of look it up

packages/reports/addon/templates/components/filter-values/range-input.hbs
daniellevinson: The `input` event makes it only one-way binding instead of two so I figured it's not necessary.

packages/reports/addon/templates/components/filter-values/range-input.hbs
```diff
@@ -1,24 +1,24 @@
...
-    class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
-    @key-up={{action "setHighValue"}}
-  />
-{{/if}}
\ No newline at end of file
+    <input
+      value={{@filter.values.lastObject}}
+      placeholder={{this.highPlaceholder}}
+      class="filter-values--range-input__input {{if @filter.validations.attrs.values.isInvalid "filter-values--range-input__input--error"}}"
+      {{on "input" (action this.setHighValue value="target.value")}}>
```
mattgraham1995: I've never seen this `value="..."` syntax with the action helper before. Is there any way to get the same functionality using the new `fn` helper instead?

kevinhinterlong: You can't accomplish the same thing just via the `fn` helper, but we could use `pipe` to read in the event, map it to `target.value`, and then pass it to `setHighValue`
```suggestion
      {{on "input" (pipe (get _ "target.value") this.setHighValue)}}>
```

daniellevinson: saw it in https://github.com/emberjs/ember.js/blob/d96d9aad52e010da977d813f479494397ee53ec6/packages/%40ember/-internals/glimmer/lib/helpers/action.ts#L86-L113 and https://github.com/emberjs/ember.js/blob/d96d9aad52e010da977d813f479494397ee53ec6/packages/%40ember/-internals/glimmer/lib/helpers/mut.ts#L81

mattgraham1995: Your second example uses the `fn` helper so can we try using `fn` instead?

daniellevinson: why? :)

mattgraham1995:
```suggestion
      {{on "input" this.setHighValue}}>
```

Then in the `setHighValue` and `setLowValue` actions the first argument should be the input event. You can get `target.value` from there.

mattgraham1995: https://emberjs.github.io/rfcs/0470-fn-helper.html

Ember is moving away from overusage of the action helper

PR #632, opened 2020-01-14
Update: Demo app logo url

kevinhinterlong:
## Description
no logo

## Proposed Changes
yes logo

## License


balajimanoharan:
We need to update this as well.
https://github.com/yahoo/navi/blob/master/packages/app/app/templates/components/top-bar.hbs#L4

PR #633, opened 2020-01-15
Fix: Send credentials with requests

kevinhinterlong:
## Description
The #619 PR removed the fetch adapter, which caused Cookies to not be included in requests and resulted in 401s when fetching resources

## Proposed Changes
Send credentials to persistence api host

## License

### DIFF
--- a/packages/core/addon/adapters/base-json-adapter.js
+++ b/packages/core/addon/adapters/base-json-adapter.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2020, Yahoo Holdings Inc.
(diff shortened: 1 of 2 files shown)


packages/core/addon/adapters/base-json-adapter.js
```diff
@@ -31,7 +31,10 @@ export default DS.JSONAPIAdapter.extend({
    */
   ajaxOptions() {
     let hash = this._super(...arguments);
-    hash.credentials = 'include';
+    hash.xhrFields = {
```
mattgraham1995: Let's add a unit test for this

mattgraham1995: Also ©️

PR #639, opened 2020-01-28
New: Added Report Search Provider

ldkge:
#625  Description
Adds new navi_report_search_provider for navi_search

## Proposed Changes

- Adds navi_base_search_provider the provides a template for all search providers
- Implements navi_report_search_provider that handles searching for reports in an elide webservice

## License


packages/core/app/mirage/routes/report.js
```diff
@@ -8,17 +8,39 @@ export default function() {
    * reports/ - GET endpoint to fetch many reports
    */
   this.get('/reports', function({ reports }, request) {
+    let reportsObject = null;
```
colemanProjects: should this be `null` or `undefined`. Could we just do `let reportsObject;`?

packages/core/app/mirage/routes/report.js
```diff
@@ -8,17 +8,39 @@ export default function() {
...
+    let queryFilter = request.queryParams['filter[reports]'];
 
     // Allow filtering
     if (idFilter) {
       let ids = idFilter.split(',');
-      reports = reports.find(ids);
+      reportsObject = reports.find(ids);
+    } else if (queryFilter) {
+      let filterParameters = queryFilter
+        .split(';')[0]
```
colemanProjects: it's hard for me to tell first off what this transformation is doing. Would it be helpful to have a separate function that does the transformation. Then we could have a javadoc style comment at the top that says e.g.`Transforms filters in the format of xyz into abc`

packages/core/app/mirage/routes/report.js
```diff
@@ -8,17 +8,39 @@ export default function() {
...
+        .replace(/\(/g, '')
+        .replace(/\)/g, '')
+        .replace(/\*/g, '')
+        .split(',')
+        .map(el => el.split('==')[1]);
+      let author = queryFilter.includes('author')
+        ? queryFilter
```
colemanProjects: maybe a similar function could be invoked here? `getQueryAuthor`. It could return `undefined` or `null` whichever we think is best. Then there wouldn't be a need for the turnery here

packages/core/app/mirage/routes/report.js
```diff
@@ -8,17 +8,39 @@ export default function() {
...
+            .split(';')[1]
+            .replace(/\*/g, '')
+            .split('==')[1]
+        : null;
+      reportsObject = reports.all().filter(function(d) {
```
colemanProjects: could we have a more specific variable name instead of `d`? maybe `filteredReports`

packages/core/app/mirage/routes/report.js
```diff
@@ -8,17 +8,39 @@ export default function() {
...
+        // Return all reports that include the filter parameters and are written by the specified author (if exists)
+        return filterParameters.every(arr => JSON.stringify(d).match(new RegExp(arr, 'i'))) && author != null
```
colemanProjects: `arr` can we maybe name it `filterParam`?

colemanProjects: we talked about refactoring the tunery here

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,29 @@
+import Service from '@ember/service';
+import { getOwner } from '@ember/application';
+import config from 'ember-get-config';
+
+/* global requirejs */
+
+export default class NaviSearchProviderService extends Service {
+  /**
+   * @method getService
```
colemanProjects: maybe `getProvider`

colemanProjects: or `getSearchProviderByName`

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  /**
+   * @method all
+   * @returns {Array} an array of available search providers
+   */
+  all() {
+    const searchProvidersRegex = new RegExp(`^(?:${config.modulePrefix}/)?services/navi-search/([a-z-]*)$`),
+      searchProviderServices = Object.keys(requirejs.entries).filter(
+        key => searchProvidersRegex.test(key) && !key.includes('base')
```
colemanProjects: should we just call out the entire navi-base-search-provider to exclude here.

packages/search/addon/services/navi-search-provider.js
colemanProjects: or `getById`

colemanProjects: `getByName`

colemanProjects: when others use this service they'd do something like: 
`searchProviderService: service.inject()`

`searchProviderService.all()`
or
`searchProviderService.getById()`

ldkge: I think we can do `getByName` since in the search provider we have a name property `name = 'reportSearchProvider';`

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,69 @@ import moment from 'moment';
...
+    } else if (queryFilter) {
+      let filterParameters = getFilterParams(queryFilter);
+      let author = getQueryAuthor(queryFilter);
+      reportsObject = reports.all().filter(function(report) {
+        // Return all reports that include the filter parameters and are written by the specified author (if exists)
+        const matchesFilterParameter = filterParameters.every(filterParameter =>
+          JSON.stringify(report).match(new RegExp(filterParameter, 'i'))
+        );
+        const mathesAuthorIfExists = author != null ? report.author.id.match(new RegExp(author, 'i')) : true;
+        return matchesFilterParameter && mathesAuthorIfExists;
```
colemanProjects: typo `mathesAuthorIfExists` --> `matchesAuthorIfExists`.

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,69 @@ import moment from 'moment';
...
+    let queryFilter = request.queryParams['filter[reports]'];
 
     // Allow filtering
     if (idFilter) {
       let ids = idFilter.split(',');
-      reports = reports.find(ids);
+      reportsObject = reports.find(ids);
+    } else if (queryFilter) {
+      let filterParameters = getFilterParams(queryFilter);
+      let author = getQueryAuthor(queryFilter);
```
colemanProjects: 👍

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,69 @@ import moment from 'moment';
...
+      reportsObject = reports.all().filter(function(report) {
+        // Return all reports that include the filter parameters and are written by the specified author (if exists)
```
colemanProjects: do we need this comment?

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,29 @@
+import Service from '@ember/service';
```
colemanProjects: copyright?

colemanProjects: and maybe a description of what the search provider service is?

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  /**
+   * @method all
+   * @returns {Array} an array of available search providers
+   */
+  all() {
+    const searchProvidersRegex = new RegExp(`^(?:${config.modulePrefix}/)?services/navi-search/([a-z-]*)$`),
+      searchProviderServices = Object.keys(requirejs.entries).filter(
+        key => searchProvidersRegex.test(key) && !key.includes('navi-base-search-provider')
```
colemanProjects: can the variable `key` be more specific? maybe `requirejsFileName`

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,29 @@
...
+      ),
+      searchProviderArray = searchProviderServices.map(key => this.getProvider(searchProvidersRegex.exec(key)[1]));
```
colemanProjects: can the variable `key` be more specific here too? Maybe `providerFileName`

packages/search/addon/services/navi-search/navi-base-search-provider.js
```diff
@@ -0,0 +1,28 @@
+import Service, { inject as service } from '@ember/service';
```
colemanProjects: copyright & description?

packages/search/addon/services/navi-search/navi-base-search-provider.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default class NaviBaseSearchProviderService extends Service {
+  /**
+   * @property {Ember.Service} store
+   */
+  @service store;
+
+  /**
+   * @property name
```
colemanProjects: maybe we could have the type `{String}` as well in all of these properties?

packages/search/addon/services/navi-search/navi-base-search-provider.js
```diff
@@ -0,0 +1,28 @@
...
+   */
+  niceName = undefined;
+
+  /**
+   * @property associatedComponent
+   */
+  associatedComponent = undefined;
+
+  /**
+   * @method search
```
colemanProjects: what should search return? An array? void?

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,44 @@
+import NaviBaseSearchProviderService from './navi-base-search-provider';
+
+export default class NaviReportSearchProviderService extends NaviBaseSearchProviderService {
+  /**
+   * @property name
```
colemanProjects: can we add `@override`?

packages/search/config/environment.js
```diff
@@ -1,5 +1,35 @@
 'use strict';
 
 module.exports = function(/* environment, appConfig */) {
-  return {};
+  return {
+    moment: {
```
colemanProjects: do we need this navi config

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+// import { settled } from '@ember/test-helpers';
```
colemanProjects: do we need this comment?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+module('Unit | Service | navi-report-search-provider', function(hooks) {
+  setupTest(hooks);
+  setupMirage(hooks);
+
+  hooks.beforeEach(async function() {
+    // Load metadata needed for request fragment
+    await this.owner.lookup('service:bard-metadata').loadMetadata();
+  });
+
+  // Replace this with your real tests.
```
colemanProjects: do we need this comment?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    const results = await service.search({ title: 'Hyrule', request: 'clicks' }, 'navi_user');
+    const author = await results.get('firstObject.author.id');
+    assert.ok(results.get('firstObject').title.includes('Hyrule'));
+    assert.ok(
+      JSON.stringify(results.get('firstObject').request)
+        .toLowerCase()
+        .includes('click')
+    );
+    assert.ok(author.includes('navi_user'));
```
colemanProjects: could we have a text description of what the test is testing? like 
https://github.com/yahoo/navi/blob/master/packages/directory/tests/unit/services/directories-test.js#L33

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    assert.ok(service);
+  });
+
+  test('search by user', async function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
```
colemanProjects: could we do this in the `beforeEach` function?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+    assert.ok(
+      JSON.stringify(results.get('firstObject').request)
+        .toLowerCase()
+        .includes('click')
+    );
+    assert.ok(author.includes('navi_user'));
+  });
+
+  test('search any user', async function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
```
colemanProjects: could we do this in the `beforeEach` function?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+  test('it exists', function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    assert.ok(service);
+  });
+
+  test('search by user', async function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    const results = await service.search({ title: 'Hyrule', request: 'clicks' }, 'navi_user');
+    const author = await results.get('firstObject.author.id');
+    assert.ok(results.get('firstObject').title.includes('Hyrule'));
```
colemanProjects: could we have a text description of what the test is testing? like
https://github.com/yahoo/navi/blob/master/packages/directory/tests/unit/services/directories-test.js#L33

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+    assert.ok(
+      JSON.stringify(results.get('firstObject').request)
+        .toLowerCase()
+        .includes('click')
```
colemanProjects: could we have a text description of what the test is testing? like
https://github.com/yahoo/navi/blob/master/packages/directory/tests/unit/services/directories-test.js#L33

And, could we make sure it includes `clicks` instead of `click`?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+    await this.owner.lookup('service:bard-metadata').loadMetadata();
+  });
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    assert.ok(service);
+  });
+
+  test('search by user', async function(assert) {
```
colemanProjects: what if the search is empty?  and the user is not?

colemanProjects: what if both are empty

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    const results = await service.search({ title: 'Hyrule', request: 'clicks' }, 'navi_user');
```
colemanProjects: what if you pass in a key that is not in the request object?

colemanProjects: or duplicate keys? or if you pass in something that's not an object? like undefined, or the integer `1`

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,44 @@
...
+    assert.ok(results.get('firstObject').title.includes('Hyrule'));
+    assert.ok(
+      JSON.stringify(results.get('firstObject').request)
+        .toLowerCase()
+        .includes('click')
+    );
+    assert.ok(author.includes('navi_user'));
+  });
+
+  test('search any user', async function(assert) {
```
colemanProjects: what if the user doesn't match but the request does? maybe we could test that

packages/search/tests/unit/services/navi-search-provider-test.js
```diff
@@ -0,0 +1,17 @@
...
+
+  test('it exists', function(assert) {
+    let service = this.owner.lookup('service:navi-search-provider');
+    assert.ok(service);
+  });
+
+  test('get all search providers', function(assert) {
+    let service = this.owner.lookup('service:navi-search-provider');
+    let availableSearchProviders = service.all();
+    assert.ok(availableSearchProviders.length > 0);
```
colemanProjects: maybe this should assert on the exact number? 
`availableSearchProviders.length  == 1`
that way if we add a new one it'll fail. We'd have to update the tests to cause it to pass? Let me know what you think

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,68 @@ import moment from 'moment';
...
+ * to a list of all the OR parameters, ie., [H, Revenue]
+ */
+const getFilterParams = function(queryFilter) {
+  return queryFilter
+    .split(';')[0]
+    .replace(/\(/g, '')
+    .replace(/\)/g, '')
+    .replace(/\*/g, '')
+    .split(',')
+    .map(el => el.split('==')[1]);
```
mattgraham1995: Rather than doing all the replaces, can we just use a RegEx?
`return queryFilter.split(';')[0].match(/\*(.*?)\*/g).map(val => val.replace('*', ''));`

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,68 @@ import moment from 'moment';
...
+ * @returns Author
+ * @description Parse filter parameters in the form of "(title==*H*,request==*Revenue*)";author==*ramvish*
+ * to get the author, ie., ramvish
+ */
+const getQueryAuthor = function(queryFilter) {
+  if (queryFilter.includes('author')) {
+    return queryFilter
+      .split(';')[1]
+      .replace(/\*/g, '')
+      .split('==')[1];
```
mattgraham1995: `return queryFielder.split(';')[1].match(/\*(.*?)\*/)[0].replace('*', '');`

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,68 @@ import moment from 'moment';
...
-      reports = reports.find(ids);
+      reportsObject = reports.find(ids);
+    } else if (queryFilter) {
+      let filterParameters = getFilterParams(queryFilter);
+      let author = getQueryAuthor(queryFilter);
+      reportsObject = reports.all().filter(function(report) {
+        const matchesFilterParameter = filterParameters.every(filterParameter =>
+          JSON.stringify(report).match(new RegExp(filterParameter, 'i'))
+        );
+        const matchesAuthorIfExists = author != null ? report.author.id.match(new RegExp(author, 'i')) : true;
```
mattgraham1995:
```suggestion
        const matchesAuthorIfExists = author && report.author.id.match(new RegExp(author, 'i')) || true;
```

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,68 @@ import moment from 'moment';
...
     if (idFilter) {
       let ids = idFilter.split(',');
-      reports = reports.find(ids);
+      reportsObject = reports.find(ids);
+    } else if (queryFilter) {
+      let filterParameters = getFilterParams(queryFilter);
+      let author = getQueryAuthor(queryFilter);
+      reportsObject = reports.all().filter(function(report) {
+        const matchesFilterParameter = filterParameters.every(filterParameter =>
+          JSON.stringify(report).match(new RegExp(filterParameter, 'i'))
```
mattgraham1995: Would be better to Stringify the report outside of the `every` callback since the result will be the same with each call.
Would be best to keep track of the property the filter value is being applied on and stringifying that property in the report to search for. 

If a filter contains `title==*H*` then we should do `JSON.stringify(report.title)`

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * @method all
+   * @returns {Array} array of available search providers
+   */
+  all() {
+    const searchProvidersRegex = new RegExp(`^(?:${config.modulePrefix}/)?services/navi-search/([a-z-]*)$`),
+      searchProviderServices = Object.keys(requirejs.entries).filter(
+        requirejsFileName =>
+          searchProvidersRegex.test(requirejsFileName) && !requirejsFileName.includes('navi-base-search-provider')
```
mattgraham1995: We could just exclude `navi-base-search-provider` in the regex

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,39 @@
...
+      ),
+      searchProviderArray = searchProviderServices.map(providerFileName =>
+        this.getProvider(searchProvidersRegex.exec(providerFileName)[1])
+      );
+    return searchProviderArray;
```
mattgraham1995: We don't need the declaration of this variable. We can just return directly. 
```suggestion
    return searchProviderServices.map(providerFileName =>
      this.getProvider(searchProvidersRegex.exec(providerFileName)[1])
    );
```

packages/search/addon/services/navi-search/navi-base-search-provider.js
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * @property {String} associatedComponent
+   */
+  associatedComponent = undefined;
+
+  /**
+   * @method search
+   * @returns {Array} array of search results
+   */
+  search() {}
```
mattgraham1995: Should we add an assert statement in here to make sure `search` is never run on the base search provider?

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,47 @@
...
+  associatedComponent = 'navi-report-search-result';
+
+  /**
+   * @method search
+   * @override
+   * @param {Object} searchParams
+   * @param {Object} author
+   * @returns {Promise} promise
+   */
+  search(searchParams, author) {
```
mattgraham1995: Can this function be run with just an author property? What if searchParams is an empty object?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,49 @@
...
+    await this.owner.lookup('service:bard-metadata').loadMetadata();
+    this.service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+  });
+
+  test('it exists', function(assert) {
+    const service = this.owner.lookup('service:navi-search/navi-report-search-provider');
+    assert.ok(service);
+  });
+
+  test('search by user', async function(assert) {
```
mattgraham1995: We should have tests to confirm that the query string is being constructed correctly. These tests are okay but they rely on the mirage endpoint so it's not truly a unit test.

packages/search/tests/unit/services/navi-search-provider-test.js
mattgraham1995: `assert.equal(availableSearchProviders.length, 1, "Put some assert message here so it's easier to debug if it fails");`

packages/core/app/mirage/routes/report.js
ldkge: That would always return true right? I want to return true only if author is null, otherwise I should return whatever the regex returns.

mattgraham1995: No the `&&` operator and `||` operators work differently for non-bool types.

Screen Shot 2020-02-03 at 2 41 20 PM

packages/core/app/mirage/routes/report.js
```diff
@@ -3,22 +3,68 @@ import moment from 'moment';
...
 const TIMESTAMP_FORMAT = 'YYYY-MM-DD HH:mm:ss';
 
+/**
+ * @method getFilterParams
+ * @param {String} queryFilter
+ * @returns {Array} Filter parameters
+ * @description Parse filter parameters in the form of "(title==*H*,request==*Revenue*)";author==*ramvish*
+ * to a list of all the OR parameters, ie., [H, Revenue]
+ */
+const getFilterParams = function(queryFilter) {
```
balajimanoharan: Would it be better to use a parsing library instead? https://github.com/vodori/rsql-parser=

packages/search/addon/services/navi-search-provider.js
balajimanoharan: or just move it up one level?

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,38 @@
...
+   * @returns {Array} array of available search providers
+   */
+  all() {
+    const searchProvidersRegex = new RegExp(`^(?:${config.modulePrefix}/)?services/navi-search/([a-z-]*)$`),
+      searchProviderServices = Object.keys(requirejs.entries).filter(
+        requirejsFileName =>
+          searchProvidersRegex.test(requirejsFileName) && !requirejsFileName.includes('navi-base-search-provider')
+      );
+    return searchProviderServices.map(providerFileName =>
+      this.getProvider(searchProvidersRegex.exec(providerFileName)[1])
```
balajimanoharan: Why do we need to do the `exec`?  Aren't you already filtering them on line 31?

packages/search/addon/services/navi-search/navi-base-search-provider.js
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * @property {String} name
+   */
+  name = undefined;
+
+  /**
+   * @property {String} niceName
+   */
+  niceName = undefined;
```
balajimanoharan: How is this going to be different from name?

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,47 @@
+import NaviBaseSearchProviderService from './navi-base-search-provider';
```
balajimanoharan: ©️

packages/search/addon/services/navi-search/navi-report-search-provider.js
ldkge: it now can, it will return the reports that are authored by the specified user, added a test for it as well.

packages/search/addon/services/navi-search-provider.js
ldkge: moved it up one level

packages/search/addon/services/navi-search/navi-base-search-provider.js
ldkge: it's the same name but written not in camel case.

packages/search/addon/services/navi-search-provider.js
mattgraham1995: Then we can remove the `!requirejsFileName.includes(...` check

packages/search/addon/services/navi-search-provider.js
ldkge: Screen Shot 2020-02-04 at 10 57 06 AM

ldkge: with exec we're pretty much just getting the name of the provider not the whole filepath

packages/search/addon/services/navi-search/navi-base-search-provider.js
mattgraham1995: Is it necessary to store another property for that or can we just use a string formatter of some sort whenever the nice name is needed?

packages/core/app/mirage/routes/report.js
```diff
@@ -1,24 +1,39 @@
...
 import RESPONSE_CODES from '../enums/response-codes';
+import { getFilterParams, getQueryAuthor } from 'navi-core/utils/rsql-utils';
 const TIMESTAMP_FORMAT = 'YYYY-MM-DD HH:mm:ss';
 
 export default function() {
   /**
    * reports/ - GET endpoint to fetch many reports
    */
   this.get('/reports', function({ reports }, request) {
+    let reportsObject;
```
balajimanoharan: reportsObject or reportObjects?

ldkge:
Here's a Navi Search RFC, it should help with the review process: https://github.com/yahoo/navi/issues/645

packages/search/addon/components/navi-search-result/report.hbs
```diff
@@ -0,0 +1,13 @@
+<div {{did-insert this.search @query}}>
+    {{#if searchResult.content}}
+    {{debugger}}
```
colemanProjects: can we remove the `debugger`

packages/search/addon/components/navi-search-result/report.hbs
```diff
@@ -0,0 +1,12 @@
+<div {{did-insert this.search @query}}>
```
balajimanoharan: ©️

packages/search/addon/components/navi-search-result/report.hbs
```diff
@@ -0,0 +1,12 @@
...
+    {{#if searchResult.content}}
+    <h1>{{this.title}}</h1>
+    <div class="search-result">
```
balajimanoharan: Can we combine with `ul` below?

packages/search/addon/services/navi-base-search-provider.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import Service from '@ember/service';
+import { assert } from '@ember/debug';
+
+export default class NaviBaseSearchProviderService extends Service {
+  /**
+   * @property {String} associatedComponent
+   */
+  associatedComponent = undefined;
```
balajimanoharan: We don't need this anymore right?

packages/search/addon/components/navi-search-result/report.hbs
balajimanoharan: Can we remove this div and move the `did-insert` into an init method in the component?

packages/search/addon/components/navi-search-result/report.js
```diff
@@ -0,0 +1,17 @@
+import Component from '@glimmer/component';
```
balajimanoharan: ©️

packages/search/addon/services/navi-report-search-provider.js
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * @property {Ember.Service} user
+   */
+  @service user;
+
+  /**
+   * @property associatedComponent
+   */
+  associatedComponent = 'navi-report-search-result';
```
balajimanoharan: cleanup?

packages/search/addon/components/navi-search-result/report.js
```diff
@@ -0,0 +1,17 @@
...
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+export default class Report extends Component {
+  @service('navi-report-search-provider') reportSearchProvider;
+
+  @tracked searchResult;
+
+  title = 'Reports & Dashboards';
```
colemanProjects: TODO: file a ticket to pull dashboards along with the reports

packages/search/addon/services/navi-report-search-provider.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return parsedQuery;
+  }
+
+  /**
+   * @method _constructSearchQuery
+   * @private
+   * @param {Object} searchParams
+   * @param {String} author
+   * @returns {Object} search query object
```
balajimanoharan: Can we add a description to the jsdoc?

packages/search/addon/services/navi-report-search-provider.js
```diff
@@ -0,0 +1,91 @@
...
+   */
+  _constructSearchQuery(searchParams, author) {
```
balajimanoharan: Don't we already do this in our fili adapter?

packages/search/addon/services/navi-search-result.js
```diff
@@ -0,0 +1,26 @@
...
+
+import Service from '@ember/service';
+import config from 'ember-get-config';
+
+/* global requirejs */
+
+export default class NaviSearchResultService extends Service {
+  /**
+   * @method all
+   * @returns {Array} array of available search providers
```
balajimanoharan: needs update?

packages/search/addon/services/navi-search-result.js
```diff
@@ -0,0 +1,26 @@
...
+   */
+  all() {
+    const resultComponentRegex = new RegExp(`^(?:${config.modulePrefix}/)?components/navi-search-result/([a-z-]*)$`),
+      auxResultComponentRegex = new RegExp(`^(?:${config.modulePrefix}/)?components/([a-z-/]*)$`),
```
balajimanoharan: What is this for? this wasn't part of the rfc.

packages/search/addon/components/navi-search-result/report.js
```diff
@@ -0,0 +1,17 @@
...
+
+  @action
+  async search(element, query) {
+    this.searchResult = await this.reportSearchProvider.search(query);
```
colemanProjects: should we either have a javadoc explaining that this sets the search result property, or define the searchResult property as a property of the component?

packages/search/addon/services/navi-search-result.js
balajimanoharan: Can we add a description?

packages/search/addon/services/navi-base-search-provider.js
colemanProjects: should this have a "searchResult" property?

packages/search/addon/services/navi-report-search-provider.js
```diff
@@ -0,0 +1,91 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * This service is used to search for reports stored in the persistence layer.
```
colemanProjects: `reports & dashboards`?

packages/search/addon/services/navi-report-search-provider.js
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * @method search
+   * @override
+   * @param {String} query
+   * @param {String} author
+   * @returns {Promise} promise with search query results
+   */
+  search(query) {
+    const parsedQuery = this._parseQueryString(query);
+    return this.store.query('report', this._constructSearchQuery(parsedQuery.searchParams, parsedQuery.author));
```
colemanProjects: the params say that you can pass in an author, but you can't.

packages/search/addon/services/navi-report-search-provider.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    return query;
+  }
+
+  /**
+   * @method search
+   * @override
+   * @param {String} query
+   * @param {String} author
```
colemanProjects: maybe we should remove this?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
+import Service from '@ember/service';
```
colemanProjects: maybe we can add a test for the base search provider. to make sure it throws the error when you try to search?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
...
+    );
+  });
+
+  test('it exists', function(assert) {
+    const service = this.owner.lookup('service:navi-report-search-provider');
+    assert.ok(service);
+  });
+
+  test('construct full search query', function(assert) {
+    assert.deepEqual(this.service._constructSearchQuery({ title: 'Hyrule', request: 'clicks' }, 'navi_user'), {
```
colemanProjects: could we add text to the end explaining what the expected result is? `Report constructs the correct query for the api` or something like that

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
...
+      filter: { reports: '(title==*Hyrule*,request==*clicks*);author==*navi_user*' }
+    });
+  });
+
+  test('construct only query parameters search query', function(assert) {
+    assert.deepEqual(this.service._constructSearchQuery({ title: 'Hyrule', request: 'clicks' }), {
```
colemanProjects: same here, can we add text

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
...
+      filter: { reports: '(title==*Hyrule*,request==*clicks*)' }
+    });
+  });
+
+  test('construct only author search query', function(assert) {
+    assert.deepEqual(this.service._constructSearchQuery(null, 'navi_user'), {
+      filter: { reports: 'author==*navi_user*' }
```
colemanProjects: and here?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
...
+    const results = await this.service.search('Hyrule');
+    const author = await results.get('firstObject.author.id');
+    assert.ok(
+      results.get('firstObject').title.includes('Hyrule'),
+      'The service returns a report that includes the requested title.'
+    );
+    assert.ok(
+      JSON.stringify(results.get('firstObject').request)
+        .toLowerCase()
+        .includes('click'),
```
colemanProjects: where do we ask for `click`?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
...
+      'The service returns a report that includes the requested request parameter.'
+    );
+    assert.ok(author.includes('navi_user'), 'The service returns a report from the requested user.');
+  });
+
+  test('search with no results for search parameters', async function(assert) {
+    const results = await this.service.search('something');
+    assert.equal(results.content.length, 0, 'No results are being returned');
```
colemanProjects: maybe `No results when there's no match`?

packages/search/tests/unit/services/navi-report-search-provider-test.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  test('search with empty parameters', async function(assert) {
+    const results = await this.service.search();
+    assert.equal(results.content.length, 0, 'No results are being returned');
```
colemanProjects: `No results for an empty query`?

packages/search/tests/integration/components/navi-search-result/report-test.js
```diff
@@ -0,0 +1,34 @@
...
+    const store = this.owner.lookup('service:store'),
+      mockAuthor = store.createRecord('user', { id: 'navi_user' });
+    this.owner.register(
+      'service:user',
+      Service.extend({
+        getUser: () => mockAuthor
+      })
+    );
+  });
+
```
colemanProjects: a few other ideas for things to test (probably integration tests): 
- special characters in the search string (commas, etc.., since it's user input)
- if something weird comes back from the ws, like an empty object comes back without a title.

packages/search/addon/components/navi-search-result/report.hbs
ldkge: What do we mean by combine?

packages/search/addon/services/navi-search-result.js
ldkge: this regex matches the name of the component together with the folder so for example from `dummy/components/navi-search-result/report` it will return `navi-search-result/report`

ldkge: Will add that to the RFC as well.

packages/search/addon/services/navi-report-search-provider.js
ldkge: No, the adapter doesn't support filtering

packages/search/addon/components/navi-search-result/report.hbs
balajimanoharan: move the class to the `ul` and remove the `div`.

packages/core/app/mirage/routes/dashboard.js
```diff
@@ -43,17 +45,45 @@ export default function() {
...
   this.get('/dashboards', ({ dashboards }, request) => {
+    let dashboardObject;
     let idFilter = request.queryParams['filter[dashboards.id]'];
+    let queryFilter = request.queryParams['filter[dashboards]'];
 
     // Allow filtering
     if (idFilter) {
       let ids = idFilter.split(',');
-      dashboards = dashboards.find(ids);
+      dashboardObject = dashboards.find(ids);
```
colemanProjects: could we abstract this function out to `utils` and then pull it in and use it here and in reports?

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,61 @@
...
+    );
+  }
+
+  /**
+   * @method search
+   * @param {String} query
+   * @returns {Array} array of objects that contain the search results,
+   * the name of the result component as well as result ordering information
+   * @description uses all the discovered search providers to fetch search results
```
colemanProjects: perfect 👍

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,111 @@
...
+    const dashboardParsedQuery = this._parseQueryString(query, 'dashboard');
+    const reportPromise = this.store.query(
+      'report',
+      this._constructSearchQuery(reportParsedQuery.searchParams, reportParsedQuery.author, 'reports')
+    );
+    const dashboardPromise = this.store.query(
+      'dashboard',
+      this._constructSearchQuery(dashboardParsedQuery.searchParams, dashboardParsedQuery.author, 'dashboards')
+    );
+    return Promise.all([reportPromise, dashboardPromise]).then(function(values) {
```
colemanProjects: 👍

packages/search/tests/unit/services/navi-base-search-provider-test.js
```diff
@@ -0,0 +1,23 @@
...
+    assert.ok(service);
+  });
+
+  test('cannot call search from base class', function(assert) {
+    const service = this.owner.lookup('service:navi-base-search-provider');
+    try {
+      service.search();
+    } catch (error) {
+      assert.ok(
+        error.message.includes('Search method must be called from a subclass'),
```
colemanProjects: 👍

packages/search/addon/components/navi-search-result/report.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+
+import Component from '@glimmer/component';
+
+export default class Report extends Component {
```
mattgraham1995: It could just be my preference, but I like to give class names more context within their names. We have lots of objects throughout navi that represent a `Report` in one place or another. Could we call this `class ReportSearchResult` instead?

packages/search/addon/components/navi-search-result/report.js
```diff
@@ -0,0 +1,14 @@
...
+  /**
+   * @property {String} title
+   * @description Title of search result
+   */
+  title = 'Reports & Dashboards';
```
mattgraham1995: So will dashboards be included in report search results? Maybe this shouldn't be called the Report search result then. I think we decided a while ago to call the base class for a report/dashboard an `asset`, although we refer to it as `item` in some other places.

packages/search/addon/services/navi-base-search-provider.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import Service from '@ember/service';
+import { assert } from '@ember/debug';
+
+export default class NaviBaseSearchProviderService extends Service {
+  /**
+   * @property {String} displayComponentName
+   */
+  displayComponentName = undefined;
```
mattgraham1995:
```suggestion
  displayComponentName;
```

packages/search/addon/services/navi-search-provider.js
mattgraham1995: We should probably have our jsdoc styles defined more clearly within navi. We usually put the description in the same line as the method name with a dash separating them rather than having its own tag and line.

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,61 @@
...
+   */
+  async search(query) {
+    const searchProviders = this.all();
+    let results = [];
+    for (const provider of searchProviders) {
+      const data = await provider.search(query);
+      if (data.length) {
+        results.push({
+          component: provider.displayComponentName,
+          data: data
```
mattgraham1995:
```suggestion
          data
```

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,111 @@
...
+   */
+  @service user;
+
+  /**
+   * @property {String} displayComponentName
+   */
+  displayComponentName = 'navi-search-result/report';
+
+  /**
+   * @method _parseReportString
```
mattgraham1995: jsdoc method name doesn't match method

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    return parsedQuery;
+  }
+
+  /**
+   * @method _constructSearchQuery
+   * @private
+   * @param {Object} searchParams
+   * @param {String} author
```
mattgraham1995: Need a param line for type

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,111 @@
...
+   * @returns {Object} search query object
+   * @description Constructs the query filter parameters adhering to the RSQL standard
+   */
+  _constructSearchQuery(searchParams, author, type) {
+    let query = { filter: { [type]: '' } };
+
+    if (searchParams) {
+      for (let p in searchParams) {
```
mattgraham1995: Instead of adding each character one by one could we map on the keys in searchParams, construct an array of filter strings for each param and call join on the array with a comma, then wrap that joined string in parentheses?

mattgraham1995:
```js
const paramsFilter = `${Object.keys(searchParams).map(p => `${p}==*${searchParams[p]}*`).join(',')}`; // comma separated list of param filters
const paramsFilterString = filters ? `(${filters})` : ''; //wrap in parentheses if param filter present
let authorFilterString = '';
if (author) {
  authorFilterString = paramsFilterString ? `;author==*${author}*` : `author==*${author}*`; //add semicolon if param filters present
}
const filterString = `${paramsFilterString}${authorFilterString}`;
query.filter[type] = filterString;
```

packages/search/addon/components/navi-search-result/report.js
ldkge: I'd rather keep it as report since it already lives under the `navi-search-result` folder. When you use it you will refer that as `<NaviSearchResult::Report />` so there is contenxt.

mattgraham1995: That would still be the case if you change the class name as that is determined by the file name. What is the advantage to keeping the class just named `Report`?

packages/search/addon/components/navi-search-result/report.js
ldkge: I think calling it `Report` is more descriptive that `Asset` or `Item`. It's supposed to show just a list of all the results that are fetched. 
Screen Shot 2020-02-19 at 2 27 45 PM

packages/core/addon/utils/rsql-utils.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * Utilities for parsing RSQL.
+ */
+
+import Mirage from 'ember-cli-mirage';
```
mattgraham1995: We should import Response directly here too

packages/search/addon/components/navi-search-result/report.js
mattgraham1995: Report references an actual model though. This search result returns dashboards as well as reports, not just reports so the name of reports is misleading. If you want to have it include both reports and dashboards, it should have a different name as we may need a search result that only returns reports where the name `navi-search-result/report` would be more appropriate.

packages/search/addon/services/navi-search/navi-report-search-provider.js
```diff
@@ -0,0 +1,105 @@
...
+      paramsFilterString = paramsFilter ? `(${paramsFilter})` : ''; //wrap in parentheses if param filter present
+    }
+
+    let authorFilterString = '';
+    if (author) {
+      authorFilterString = paramsFilterString ? `;author==*${author}*` : `author==*${author}*`; //add semicolon if param filters present
+    }
+
+    const filterString = `${paramsFilterString}${authorFilterString}`;
+    query.filter[type] = filterString;
```
mattgraham1995: My mistake, we don't actually need the filterString variable.
```suggestion
    query.filter[type] = `${paramsFilterString}${authorFilterString}`;
```

packages/search/package.json
```diff
@@ -21,11 +21,18 @@
     "test": "export COVERAGE=true; ember test --test-port 0"
   },
   "dependencies": {
+    "ember-ajax": "^5.0.0",
```
mattgraham1995: What do we need ajax for? I thought we were trying to move away from ajax and towards the fetch api.

packages/search/addon/components/navi-search-result/report-and-dashboard.hbs
```diff
@@ -0,0 +1,7 @@
+{{!-- Copyright 2020, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+<h1>{{this.title}}</h1>
+<ul class="search-result">
```
mattgraham1995:
```suggestion
<ul class="navi-search-results">
```

packages/search/addon/components/navi-search-result/report-and-dashboard.hbs
```diff
@@ -0,0 +1,7 @@
...
+    {{#each @data as |result|}}
+    <li>{{result.title}}</li>
```
mattgraham1995: Can we give a class to the `li` elements?

packages/search/tests/integration/components/navi-search-result/report-and-dashboard-test.js
```diff
@@ -0,0 +1,44 @@
...
+    assert.equal(this.element.textContent.trim(), 'Reports & Dashboards');
+  });
+
+  test('displays results', async function(assert) {
+    const result = await this.service.search('Revenue');
+    set(this, 'result', result.find(element => element.component === 'navi-search-result/report-and-dashboard'));
+
+    await render(hbs`<NaviSearchResult::ReportAndDashboard @data={{this.result.data}}/>`);
+
+    assert.dom('li').exists('Showing search results');
```
mattgraham1995: Could we assert against the content of the `li` rather than just making sure it's there?

packages/search/tests/unit/services/navi-search-provider-test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  test('it exists', function(assert) {
+    let service = this.owner.lookup('service:navi-search-provider');
+    assert.ok(service);
+  });
+
+  test('get all search providers', function(assert) {
+    let availableSearchProviders = this.service.all();
+    assert.equal(availableSearchProviders.length, 1, 'Discovered 1 search provider.');
```
mattgraham1995: Can we check the properties of the returned search providers rather than just testing the length? `availableSearchProviders` could be `['foo']` and it would satisfy this test case.

packages/search/tests/unit/services/navi-search-provider-test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  test('search all providers', async function(assert) {
+    let results = await this.service.search('Revenue');
+    assert.ok(results.length > 0, 'Returns multiple results');
```
mattgraham1995: same for results

packages/search/tests/unit/services/navi-search-provider-test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  test('search with no results', async function(assert) {
+    let results = await this.service.search('something');
+    assert.ok(results.length == 0, 'Returns no results');
```
mattgraham1995:
```suggestion
    assert.equal(results.length, 0, 'Returns no results');
```

packages/search/tests/unit/services/navi-search-provider-test.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  test('search with special characters', async function(assert) {
+    let results = await this.service.search('lala!@#$%^&*');
+    assert.ok(results.length == 0, 'Returns no results');
```
mattgraham1995:
```suggestion
    assert.equal(results.length, 0, 'Returns no results');
```

packages/search/tests/unit/services/navi-search/navi-report-and-dashboard-search-provider-test.js
```diff
@@ -0,0 +1,105 @@
...
+        filter: { dashboards: 'author==*navi_user*' }
+      },
+      'Report constructs the correct query for the api with author.'
+    );
+  });
+
+  test('search by user search returns reports and dashboards', async function(assert) {
+    const results = await this.service.search('Revenue');
+    const author = await results[0].author;
+    assert.ok(results[0].title.includes('Revenue'), 'The service returns a report that includes the requested title.');
```
mattgraham1995: Can we assert against all results rather than just the first one? We want to make sure that all results have `Revenue` in their title field and have author `ciela`

packages/core/app/utils/rsql-utils.js
```diff
@@ -0,0 +1 @@
+export { default } from 'navi-core/utils/rsql-utils';
```
balajimanoharan: Isnt this just for mirage, should it be part of core? Can we just make it a mirage util?

packages/data/addon/models/metadata/table.js
```diff
@@ -58,14 +58,15 @@ let Model = EmberObject.extend({
...
         'timeGrains',
-        timeGrains.map(timeGrain => {
-          let timeGrainPayload = assign({}, timeGrain, { source: this.source }),
-            owner = getOwner(this);
-          return owner.factoryFor('model:metadata/time-grain').create(timeGrainPayload);
-        })
-        .sort((l, r) => {
-          return (timeGrainSorting[l.name] || Infinity) - (timeGrainSorting[r.name] || Infinity);
-        })
+        timeGrains
```
balajimanoharan: CAn we revert? This file is not changed.

packages/search/addon/components/navi-search-result/report-and-dashboard.hbs
balajimanoharan: also indent is off? on line 4

packages/search/addon/components/navi-search-result/report-and-dashboard.js
```diff
@@ -0,0 +1,13 @@
...
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+
+import Component from '@glimmer/component';
+
+export default class ReportAndDashboardSearchResult extends Component {
+  /**
+   * @property {String} title – Title of search result
+   */
+  title = 'Reports & Dashboards';
```
balajimanoharan: Can we move this to the template? Maybe convert component to template-only.

packages/search/addon/services/navi-base-search-provider.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import Service from '@ember/service';
+import { assert } from '@ember/debug';
+
+export default class NaviBaseSearchProviderService extends Service {
+  /**
+   * @property {String} displayComponentName
+   */
+  displayComponentName;
```
balajimanoharan: Can we make this a `get displayComponent..`? And assert that its defined in the subclass like search?

packages/search/addon/services/navi-search-provider.js
balajimanoharan:
```js
searchProviderServices = Object.keys(requirejs.entries)
  .filter(requirejsFileName => searchProvidersRegex.test(requirejsFileName))
  .map(fileName => searchProvidersRegex.exec(fileName)[1]));
```

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,59 @@
...
+  getProvider(name) {
+    return getOwner(this).lookup(`service:navi-search/${name}`);
+  }
+
+  /**
+   * @method all – Discovers all the search provider services
+   * under the folder services/navi-search/
+   * @returns {Array} array of available search provider services
+   */
+  all() {
```
balajimanoharan: Is this method meant to be used outside this service? If not, should this be private?

packages/search/addon/services/navi-search/navi-report-and-dashboard-search-provider.js
```diff
@@ -0,0 +1,104 @@
...
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * This service is used to search for reports and dashboards stored in the persistence layer.
+ */
+
+import { inject as service } from '@ember/service';
+import NaviBaseSearchProviderService from '../navi-base-search-provider';
+
+export default class NaviReportAndDashboardSearchProviderService extends NaviBaseSearchProviderService {
```
balajimanoharan: Would it be better to separate the search providers for reports and dashboards? Whats the benefit from using just one?

packages/core/app/utils/rsql-utils.js
ldkge: if so it should live under `addon/mirage/utils`?

packages/search/addon/components/navi-search-result/report-and-dashboard.hbs
```diff
@@ -0,0 +1,7 @@
+{{!-- Copyright 2020, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+<h1>{{this.title}}</h1>
+<ul class="navi-search-results">
+    {{#each @data as |result|}}
```
balajimanoharan: indent is off?

packages/search/addon/components/navi-search-result/report-and-dashboard.hbs
```diff
@@ -0,0 +1,7 @@
+{{!-- Copyright 2020, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
+<h1>{{this.title}}</h1>
+<ul class="navi-search-results">
```
balajimanoharan: Can we follow the `BEM` styles for this component?

packages/data/addon/models/metadata/table.js
ldkge: The commit formatting command is reverting it back to this every time :/

packages/search/package.json
```diff
@@ -21,30 +21,41 @@
     "test": "export COVERAGE=true; ember test --test-port 0"
   },
   "dependencies": {
+    "ember-auto-import": "^1.5.3",
     "ember-cli-babel": "^7.13.0",
-    "ember-cli-htmlbars": "^4.2.0"
+    "ember-cli-htmlbars": "^4.2.0",
+    "ember-cli-mirage": "^1.1.6",
+    "ember-data": "~3.12.0",
+    "ember-uuid": "^2.1.0",
```
balajimanoharan: Why is this needed?

packages/search/package.json
```diff
@@ -21,30 +21,41 @@
...
+    "navi-core": "0.2.0"
   },
   "devDependencies": {
-    "@ember/optional-features": "^1.1.0",
+    "@ember/jquery": "^1.1.0",
```
balajimanoharan: Is this needed?

packages/search/package.json
```diff
@@ -21,30 +21,41 @@
...
+    "@ember/optional-features": "^1.3.0",
+    "@ember/render-modifiers": "^1.0.2",
```
balajimanoharan: and this?

packages/search/tests/dummy/app/services/navi-search/navi-sample-search-provider.js
```diff
@@ -0,0 +1,26 @@
+import NaviBaseSearchProviderService from '../navi-base-search-provider';
```
balajimanoharan: ©️

packages/search/.eslintrc.js
```diff
@@ -16,6 +16,10 @@ module.exports = {
   rules: {
     'ember/no-jquery': 'error'
   },
+  globals: {
+    c3: true,
+    d3: true
```
balajimanoharan: is this needed for this repo?

packages/search/tests/unit/services/navi-search-provider-test.js
```diff
@@ -0,0 +1,41 @@
...
+module('Unit | Service | navi-search-provider', function(hooks) {
+  setupTest(hooks);
+
+  let Service;
+
+  hooks.beforeEach(async function() {
+    Service = this.owner.lookup('service:navi-search-provider');
+  });
+
+  test('it exists', function(assert) {
```
balajimanoharan: Can we remove this test?

packages/search/addon/services/navi-search-provider.js
```diff
@@ -0,0 +1,57 @@
...
+   * under the folder services/navi-search/
+   * @private
+   * @returns {Array} array of available search provider services
+   */
+  _all() {
+    const searchProvidersRegex = new RegExp(`^(?:${config.modulePrefix}/)?services/navi-search/([a-z-]*)$`),
+      searchProviderServices = Object.keys(requirejs.entries)
+        .filter(requirejsFileName => searchProvidersRegex.test(requirejsFileName))
+        .map(fileName => searchProvidersRegex.exec(fileName)[1]);
+    return searchProviderServices.map(providerName => this.getProvider(providerName));
```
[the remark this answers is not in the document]
ldkge: I tried it but there's an issue with the `this`, it's not the right one that way.

packages/search/tests/dummy/app/services/navi-search/navi-sample-search-provider.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * This service is used to discover all the available search providers.
```
balajimanoharan: needs to be updated.

PR #640, opened 2020-01-28
Bulk import paste for multi-value-input

kevinhinterlong:
## Description
When we have a dimension with `storageStrategy === 'none'`, we fall back to multi-value-input which means we don't support dimension bulk import on (pasting a list of ids)

https://github.com/yahoo/navi/blob/842719c9c2b074aa1d04d6982cc1a0fca4c4e210/packages/reports/addon/components/filter-builders/dimension.js#L31-L34

## Proposed Changes
- Add a simple dimension importer for multi-value-input
  - detect if the users pastes something in with a `','`
  - let user choose whole string value or split values into array
- clean up some styling on dimension importing
- fix key code from `118` to `188` for commas 👀 

## Screenshots
Updated styles:
normal dimension bulk importer
New simple importer:
simple dimension bulk importer

## License

### DIFF
--- a/packages/reports/addon/components/dimension-bulk-import-simple.js
+++ b/packages/reports/addon/components/dimension-bulk-import-simple.js
@@ -9,23 +9,13 @@
  *   />
  */
 import Component from '@ember/component';
-import { action } from '@ember/object';
+import { computed } from '@ember/object';
 import layout from '../templates/components/dimension-bulk-import-simple';
 import { layout as templateLayout, tagName } from '@ember-decorators/component';
 
 @tagName('')
 @templateLayout(layout)
 class DimensionBulkImportSimpleComponent extends Component {
-  /**
-   * @property {Array<String>} unsplitValue - empty/single value array of rawInput
-   */
-  unsplitValue = [];
-
-  /**
-   * @property {Array<String>} splitValues - array of values split by commas from rawInput
-   */
-  splitValues = [];
-
   /**
    * Filters the input list to a unique list of nonempty values
    * @param {Array<String>} list - the list to filter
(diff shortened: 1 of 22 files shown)


packages/reports/addon/components/dimension-bulk-import-simple.js
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Filters the input list to a unique list of nonempty values
+   * @param {Array<String>} list - the list to filter
+   */
+  listFilter(list) {
+    const filter = list
+      .map(v => v.trim()) // remove surrounding space
+      .filter(v => v.length > 0); // remove empty values
+    return [...new Set(filter)]; // only unique falues
```
cythrawll: values

packages/reports/tests/integration/components/dimension-bulk-import-test.js
```diff
@@ -237,7 +237,7 @@ module('Integration | Component | Dimension Bulk Import', function(hooks) {
 
     //Remove First Valid Pill, item removed depends on the ordering
     run(() => {
-      $('.items-list:first .item:first button').click();
+      $('.items-list:first .item:first .remove-pill').click();
```
cythrawll: Do we need to use jquery here?

cythrawll: both. Jquery is fine when we need to do content selectors, but otherwise we don't need them. And we should be using the event helpers rather than jquery event triggers.

packages/reports/addon/components/dimension-bulk-import-simple.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * @method didReceiveAttrs
+   * @override
+   */
+  didReceiveAttrs() {
```
jkusa: Since this hook isn't in the glimmer API, can we use getters instead?

kevinhinterlong: It's effectively the same as using `{{did-update}}` render modifier so I switched it to that since this isn't actually a glimmer component so there's no `tracking` on the `rawInput` and thus no re-rendering

packages/reports/addon/components/navi-tag-input.js
```diff
@@ -29,18 +30,80 @@ export default TagInput.extend({
...
+
   /**
    * Ignore comma so that only 'enter' key adds new tags
    *
    * @param {InputEvent} e - The keydown input event
    * @override https://github.com/calvinlough/ember-tag-input/blob/v1.2.2/addon/components/tag-input.js#L74-L106
    */
   _onInputKeyDown(e) {
-    if (e.which === 118 || e.target.value === '') {
+    if (e.which === 188) {
```
mattgraham1995: Could make a const for 188 called `COMMA_KEY_CODE` instead of using a comment below this

packages/reports/tests/integration/components/dimension-bulk-import-simple-test.js
```diff
@@ -0,0 +1,79 @@
...
+    assert.expect(2);
+
+    const split = ['some weird', 'spacing', 'going', 'on'];
+    this.set('rawInput', ' some weird,  spacing,going ,on,,, ,,,');
+    assert.deepEqual(getSplitValues(), split, 'The split values will trim away weird inner spacing');
+    this.set('onSelectValues', values =>
+      assert.deepEqual(values, split, 'The selected values are just the split string array')
+    );
+    await click($('.btn-primary:contains(split)')[0]);
+  });
```
mattgraham1995: What happens if they just paste commas? For example `val1,val2,,,,`

kevinhinterlong: This component doesn't handle the pasting, so it just gets `val1,val2,,,,` as the input and like above it will
1. Split by commas
2. trim each value
3. remove empty values

so the result would be `['val1', 'val2']`

mattgraham1995: Could we add a test case for this just to make sure our bases are covered?

PR #649, opened 2020-02-07
Better filter support for bard-{dimensions,facts}

kevinhinterlong:
## Description
`adapters:bard-facts` provided support for passing multiple filters such as, but  `service:bard-dimensions` supported only a single filter
```js
[
  {
    dimension: 'dim',
    field: 'id',
    operator: 'in',
    values: [1, 2, 3]
  },
  {
    dimension: 'dim',
    field: 'desc',
    operator: 'contains',
    values: ['wow']
  }
]
```
> `dim.id` is any of `1, 2, 3` AND `dim.desc` contains `wow`

## Proposed Changes
- we now accept the same format of filters which are ANDed together where the values in a single filter are ORed
- provided a serializer/deserializer for fili's `filter` query param (since the `values` in `dim|field-op[values]` supports a csv string)
  - This provide support for commas and quotes in values
  - This will break existing reports where a filter value contains a comma since all values are `"wrapped"`
    >  `{ dimension: 'dim', field: 'id', operator: 'contains', values: ['1,2,3', '4'] }` -> 
    > 
    > OLD `dim|id-contains[1,2,3,4]`
    > 
    > NEW `dim|id-contains["1,2,3","4"]`

## Potentially Breaking
- Reports containing commas may need to be migrated
- Mirage may need to be updated with filter parser

## License

### DIFF
--- a/packages/core/addon/models/bard-request/fragments/filter.js
+++ b/packages/core/addon/models/bard-request/fragments/filter.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018, Yahoo Holdings Inc.
+ * Copyright 2020, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 
@@ -54,11 +54,11 @@ export default Fragment.extend(Validations, {
         return DS.PromiseArray.create({ promise });
       } else {
         let dimensionName = get(this, 'dimension.name'),
-          values = arr(get(this, 'rawValues')).join(','),
+          values = arr(get(this, 'rawValues')),
           dimensionService = get(this, 'dimensionService');
 
         return DS.PromiseArray.create({
-          promise: dimensionService.find(dimensionName, { values }).then(values => arr(values))
+          promise: dimensionService.find(dimensionName, [{ values }]).then(values => arr(values))
         });
       }
     },
--- a/packages/core/addon/models/bard-request/fragments/filter.js
+++ b/packages/core/addon/models/bard-request/fragments/filter.js
@@ -58,7 +58,7 @@ export default Fragment.extend(Validations, {
           dimensionService = get(this, 'dimensionService');
 
         return DS.PromiseArray.create({
-          promise: dimensionService.find(dimensionName, [{ values }]).then(values => arr(values))
+          promise: dimensionService.find(dimensionName, [{ values }]).then(arr)
         });
       }
     },
--- a/packages/data/addon/adapters/bard-facts.js
+++ b/packages/data/addon/adapters/bard-facts.js
@@ -29,8 +29,7 @@ export function serializeFilters(filters) {
   return filters
     .map(filter => {
       const { dimension, field, operator, values } = filter;
-      const serializedValues = array(values)
-        .toArray()
+      const serializedValues = values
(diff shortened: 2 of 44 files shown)


packages/data/addon/adapters/bard-facts.js
```diff
@@ -16,6 +16,30 @@ import { configHost } from '../utils/adapter';
...
+ * @param {String} filter.field - the dimension field to filter
+ * @param {String} filter.operator - the type of filter operator
+ * @param {Array<String|number>} filter.values - the values to pass to the operator
+ */
+export function serializeFilters(filters) {
+  return filters
+    .map(filter => {
+      const { dimension, field, operator, values } = filter;
+      const serializedValues = array(values)
+        .toArray()
```
jkusa: Why do we need to go from wrapping the array in an `Ember.Array` then convert into a native array?

packages/data/package.json
```diff
@@ -23,10 +23,12 @@
   },
   "dependencies": {
     "ember-ajax": "^5.0.0",
+    "ember-auto-import": "^1.2.19",
     "ember-cli-babel": "^7.11.1",
     "ember-cli-htmlbars": "^4.0.0",
     "ember-get-config": "0.2.4",
-    "ember-lodash": "^4.19.5"
+    "ember-lodash": "^4.19.5",
+    "papaparse": "^5.1.1"
```
jkusa: Should we add this to the host `devDependencies`?

kevinhinterlong: It is pulled in via `ember-auto-import` so they both must be in `dependencies` otherwise [it won't work](https://github.com/ef4/ember-auto-import#usage-from-addons), unfortunately

packages/core/addon/models/bard-request/fragments/filter.js
```diff
@@ -54,11 +54,11 @@ export default Fragment.extend(Validations, {
         return DS.PromiseArray.create({ promise });
       } else {
         let dimensionName = get(this, 'dimension.name'),
-          values = arr(get(this, 'rawValues')).join(','),
+          values = arr(get(this, 'rawValues')),
           dimensionService = get(this, 'dimensionService');
 
         return DS.PromiseArray.create({
-          promise: dimensionService.find(dimensionName, { values }).then(values => arr(values))
+          promise: dimensionService.find(dimensionName, [{ values }]).then(values => arr(values))
```
mattgraham1995:
```suggestion
          promise: dimensionService.find(dimensionName, [{ values }]).then(arr)
```

packages/data/addon/mirage/routes/bard-lite-parsers.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Parses a single filter query param
+ * @param {String} serializedFilter - a filter in the format of `dim|field-op[values]`
+ * @returns {Object} parsed filter
+ */
+export function parseSingleFilter(serializedFilter) {
+  const [, dimension, field, operator, valuesString] = serializedFilter.match(/(.*)\|(.*)-(.*)\[(.*)\]/);
+
+  // field = field === 'desc' ? 'description' : field;
```
mattgraham1995:
```suggestion
```

PR #657, opened 2020-02-12
Update: Include ember-intl config in dummy apps

kevinhinterlong:
## Description
`ember-intl` makes a _lot_ of noise during tests

## Proposed Changes
stop it

## License


cythrawll:
mood

PR #660, opened 2020-02-13
Update: Parse filters support multiple filters

kevinhinterlong:
## Description
The parser would incorrectly parse `dim|id-in[value,value 2],dim2|id-in[this]` as
```js
[
  { dimension: 'dim', field: 'id', operator: 'in', values: ['value', 'value 2'] },
  { dimension: ',dim2', field: 'id', operator: 'in', values: ['this'] }
]
```
**,dim2** and not **dim2**

## Proposed Changes
- Fix how we split the input
- added tests

## License


packages/data/addon/mirage/routes/bard-lite-parsers.js
```diff
@@ -34,8 +34,11 @@ export function parseSingleFilter(serializedFilter) {
  * @returns {Array} filters
  */
 export function parseFilters(filtersParam = '') {
-  const filterStrings = filtersParam
-    .split(']')
+  if (!filtersParam || filtersParam === '') {
```
jkusa:
```suggestion
  if (!filtersParam) {
```

packages/data/addon/mirage/routes/bard-lite-parsers.js
```diff
@@ -34,8 +34,11 @@ export function parseSingleFilter(serializedFilter) {
  * @returns {Array} filters
  */
 export function parseFilters(filtersParam = '') {
```
jkusa: Do we want the default?

kevinhinterlong:
```suggestion
export function parseFilters(filtersParam) {
```
No longer needed, 👍

packages/data/addon/mirage/routes/bard-lite-parsers.js
```diff
@@ -33,9 +33,12 @@ export function parseSingleFilter(serializedFilter) {
...
  * @returns {Array} filters
  */
-export function parseFilters(filtersParam = '') {
-  const filterStrings = filtersParam
-    .split(']')
+export function parseFilters(filtersParam) {
+  if (!filtersParam) {
+    return [];
+  }
+  const filterStrings = (filtersParam + ',')
```
balajimanoharan: template string 
```suggestion
  const filterStrings = `${filters},`
```

PR #671, opened 2020-02-14
Update: Tooltip position behavior near edge

kevinhinterlong:
## Description
If the tooltip gets too close to the right edge of the chart it will just stay snapped to the right side. This can be an issue because it will cover the data the user wants to look at

## Proposed Changes
- Mirror the tooltip over the vertical bar to the left side of the data instead of blocking the right side of the data

### Before:
before
### After:
after

## License

### DIFF
--- a/packages/core/addon/initializers/inject-c3-enhancements.js
+++ b/packages/core/addon/initializers/inject-c3-enhancements.js
@@ -173,7 +173,7 @@ export function initialize() {
    * Overrides the defaults tooltipPosition to allow moving the tooltip to the left of the cursor
    * if it would go past the right side of the chart
    *
-   * https://github.com/c3js/c3/blob/f339e290a7ddd73221e9ffe01cba6356951e22d7/src/tooltip.js#L175-L215
+   * https://github.com/c3js/c3/blob/v0.7.9/src/tooltip.js#L175-L215
    * @method tooltipPosition
    * @override
    * @returns {Object} - the top and left offset positions for the tooltip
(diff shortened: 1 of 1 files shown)


jkusa:
Can we also make this change in the c3 repo as well?

kevinhinterlong:
> Can we also make this change in the c3 repo as well?

I made an issue for it https://github.com/c3js/c3/issues/2754 to see if they are interested

> Pull requests are welcome, though please post an issue first to see whether such a change is desirable.
> https://github.com/c3js/c3#pull-requests

packages/core/addon/initializers/inject-c3-enhancements.js
```diff
@@ -168,6 +168,58 @@ export function initialize() {
...
+      } else {
+        tooltipLeft = verticalBarX + 20; // defaults to 20px after
+        tooltipRight = tooltipLeft + tWidth;
+        chartRight = svgLeft + this.currentWidth - this.getCurrentPaddingRight();
+        tooltipTop = mouseY + 15;
+      }
+
+      // override here to move the tooltip an equal amount to the left of the vertical bar
+      if (tooltipRight > chartRight) {
+        tooltipLeft = verticalBarX - tWidth - 20;
```
jkusa: The original code had a comment for Firefox, did we check this on Safari, FF, Chrome?

kevinhinterlong: I haven't seen any issue with it on Safari / FF / Chrome

PR #673, opened 2020-02-17
Update: Dashboard widgets - fetch by layout order and limit concurrency

daniellevinson:
## Description

Use `ember-concurrency` tasks to enqueue widgets requests by order in layout and add an option to limit concurrency.
Cancel running or enqueued tasks when leaving the route.

## License


packages/dashboards/tests/unit/services/dashboard-data-test.js
```diff
@@ -154,6 +156,72 @@ module('Unit | Service | dashboard data', function(hooks) {
...
+      filters: []
+    };
+
+    const widgets = [
+        { id: 1, dashboard: cloneDeep(dashboard), requests: [makeRequest(1), makeRequest(2), makeRequest(3)] },
+        { id: 2, dashboard: cloneDeep(dashboard), requests: [makeRequest(4), makeRequest(5), makeRequest(6)] }
+      ],
+      layout = [{ widgetId: 1, row: 0, column: 0 }, { widgetId: 2, row: 4, column: 0 }];
+
+    const done = assert.async();
```
kevinhinterlong: Move to top

packages/dashboards/tests/unit/services/dashboard-data-test.js
```diff
@@ -154,6 +156,72 @@ module('Unit | Service | dashboard data', function(hooks) {
...
+
+    try {
+      const data = service.fetchDataForWidgets(1, widgets, layout);
+      assert.equal(service._fetchTask.concurrency, 2, 'fetch task concurrency is correctly set by config');
+      await get(data, '1');
+    } catch (e) {
+      assert.deepEqual(fetchCalls, [1, 2], '2 first concurrent fetch rasks were running, others were cancelled');
```
kevinhinterlong:
```suggestion
      assert.deepEqual(fetchCalls, [1, 2], '2 first concurrent fetch tasks were running, others were cancelled');
```

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -36,54 +39,89 @@ export default Service.extend({
...
+    // sort widgets by order in layout
+    const sortedWidgets = arr(layout)
+      .sortBy('row', 'column')
+      .map(layoutItem => widgets.find(widget => widget.id == layoutItem.widgetId))
+      .filter(widget => widget);
+
+    // For each widget, concurrently execute a task that will fetch all widget's requests
+    return sortedWidgets.reduce(
+      (dataByWidget, widget) => ({
+        [widget.id]: DS.PromiseArray.create({
```
kevinhinterlong:
```js
dataByWidget[widget.id] = ...
return dataByWidget;
```

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -36,54 +39,89 @@ export default Service.extend({
...
+   * @param {Number} dashboardId
+   * @param {Object} widget - dashboard widget model
+   * @param {Array} decorators - array of functions to modify each request
+   * @param {Object} options - options for web service fetch
+   * @param {String} uuid - v1 UUID
+   * @returns {TaskInstance}
+   */
+  _widgetTask: task(function*(dashboardId, widget, decorators, options, uuid) {
+    const widgetId = widget.id;
+    const { dashboard, requests } = widget;
```
kevinhinterlong: `, id: widgetId } = widget;`

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -36,54 +39,89 @@ export default Service.extend({
...
+    let fetchTasks = [];
```
kevinhinterlong:
```suggestion
    const fetchTasks = [];
```

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -179,6 +217,19 @@ export default Service.extend({
...
   },
 
+  /**
+   * @property {Task} _fetchTask
+   * @private
+   * @param {Object} request
+   * @param {Object} options - options for web service fetch
+   * @returns {TaskInstance}
+   */
+  _fetchTask: task(function*(request, options) {
```
jkusa: Since services are singletons, can we make this stateless and return the task?

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -1,16 +1,19 @@
...
-import { all } from 'rsvp';
+import { task, all } from 'ember-concurrency';
 import { computed } from '@ember/object';
 import { v1 } from 'ember-uuid';
 import DS from 'ember-data';
+import config from 'ember-get-config';
+
+const FETCH_MAX_CONCURRENCY = config.navi.widgetsRequestsMaxConcurrency || Infinity;
 
 export default Service.extend({
```
kevinhinterlong: Native class? Looks like a lot of this got refactored

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -36,54 +39,89 @@ export default Service.extend({
...
+
+  /**
+   * @method cancelFetchDataForDashboard - cancels all running or enqueued fetch Task Instances
    */
-  fetchDataForDashboard(dashboard) {
-    return dashboard
-      .get('widgets')
-      .then(widgets => this.fetchDataForWidgets(dashboard.id, widgets, [], get(this, 'widgetOptions')));
+  cancelFetchDataForDashboard() {
+    this._fetchTask.cancelAll();
```
jkusa: Is it possible to have `fetchDataForWidgets` return a parent task that can be canceled. That way the singleton service doesn't have any _state_.

daniellevinson:
> Do we have a test that shows it succeeds in fetching and advancing to the next request?

yeah line 119

packages/dashboards/addon/services/dashboard-data.js
daniellevinson: for readability?

kevinhinterlong:
> > Do we have a test that shows it succeeds in fetching and advancing to the next request?
> 
> yeah line 119

👍 I read it before `widgetsRequestsMaxConcurrency: 2` was in config and was still making that assumption, thanks for pointing it out

packages/dashboards/addon/services/dashboard-data.js
jkusa: It would be nice if we can get away from DS.PromiseArray. Looks like we have a mix of usage:

DS.PromiseArray:
https://github.com/yahoo/navi/blob/271bc51d48620c041367492919b823450b0716e1/packages/dashboards/addon/components/navi-widget.js#L62

Native Promise
https://github.com/yahoo/navi/blob/271bc51d48620c041367492919b823450b0716e1/packages/dashboards/addon/templates/components/navi-widget.hbs#L48

kevinhinterlong: > for readability?

Yea seems simpler to read than creating a new object

packages/dashboards/tests/unit/services/dashboard-data-test.js
```diff
@@ -137,14 +153,85 @@ module('Unit | Service | dashboard data', function(hooks) {
...
+      async _fetch(request) {
+        fetchCalls.push(request.data);
+        await timeout(100);
+
+        // cancel all widget tasks
+        run(() => {
+          service._fetchRequestsForWidget.cancelAll();
+        });
+
+        later(() =>
```
kevinhinterlong: Was this supposed to be `return resolve({...` ?

This test feels a little confusing with the `timeout`/`cancelAll` being run in the `_fetch`. It might be simpler to do something like this
```js
return new Promise((resolve) => {
   const resolveRequest = () => resolve({ request, response: { data: request.data } });
   resolveRequests[request.data] = resolveRequest;
});
```
So that we can manually resolve each query or not and manually call `service._fetchRequestsForWidget.cancelAll();`

packages/dashboards/addon/components/navi-widget.js
```diff
@@ -1,71 +1,96 @@
...
+  }
+
+  /**
+   * @property {EmberArray|Null} data - widget data
+   */
+  @computed('taskInstance.{isSuccessful,value}')
+  get data() {
+    const { isSuccessful, value } = this.taskInstance || {};
+
+    return isSuccessful ? arr(value) : null;
```
jkusa: What is the value of `value` when `isSuccessful` is falsey?

daniellevinson: `null`

packages/dashboards/tests/unit/services/dashboard-data-test.js
daniellevinson: since now we won't be returning the `Task` from the service anymore and not calling `.cancelAll()`, I'll test only concurrency config here.

daniellevinson: actually will check that in the previous test

packages/dashboards/addon/services/dashboard-data.js
```diff
@@ -1,89 +1,155 @@
...
+
+      taskByWidget[widget.id] = taskInstance;
+
+      /**
+       * bubbling 403 errors causes acceptance test to fail https://github.com/emberjs/ember-qunit/issues/592
+       * task still expectedly fails.
+       * TODO: no need for catch block when ^ resolves (needlessly catches TaskCancelation errors as well)
+       */
+      taskInstance.catch(error => {
+        if (isForbidden(error) || (error && error.name === 'TaskCancelation')) {
```
jkusa: Do we still get forbidden messages in the UI?

packages/dashboards/addon/routes/dashboards/dashboard/view.js
```diff
@@ -92,12 +92,36 @@ export default Route.extend({
...
-      return { dashboard, dataForWidget: cachedWidgetData };
+      return { dashboard, taskByWidget: cachedWidgetData };
     }
-    const dataForWidget = await this.get('dashboardData').fetchDataForDashboard(dashboard);
-    this.set('_widgetDataCache', dataForWidget);
 
-    return { dashboard, dataForWidget };
+    /**
+     * if we don't await here, when going back and forth between dashboards while data is still loading
+     * `deactivate` gets called after these lines, cancelling new tasks and setting the new cache to null
```
jkusa: Do we have testing around this issue?

packages/dashboards/tests/unit/services/dashboard-data-test.js
```diff
@@ -280,7 +296,9 @@ module('Unit | Service | dashboard data', function(hooks) {
       }
     ];
 
-    const data = service.fetchDataForWidgets(1, widgets);
+    const layout = [{ widgetId: 1 }, { widgetId: 2 }, { widgetId: 3 }];
+
+    const data = service.fetchDataForWidgets(1, widgets, layout);
```
jkusa: Do we have any dashboard acceptance tests that have more widgets than the max concurrency and validates all dashboard widgets load?

packages/dashboards/addon/services/dashboard-data.js
daniellevinson: Yes we still check `{{#if (is-forbidden @taskInstance.error)}}`

packages/dashboards/addon/routes/dashboards/dashboard/view.js
daniellevinson: Umm don't think so, will see if I can write one quickly, would be great to find out why it happens too. To be clear it's not related to this PR

packages/dashboards/addon/services/dashboard-data.js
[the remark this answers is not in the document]
daniellevinson: Yep, changed

packages/dashboards/tests/unit/services/dashboard-data-test.js
daniellevinson: Added to the existing "dashboard success" test (dashboard has 3 widgets, max concur in test app is 2)

packages/dashboards/addon/routes/dashboards/dashboard/view.js
```diff
@@ -92,12 +92,36 @@ export default Route.extend({
...
+   * @returns {Boolean} - true after cancelling
+   */
+  _cancelWidgetDataTasks() {
+    const { _widgetDataCache } = this;
+
+    if (!_widgetDataCache) {
+      return true;
+    }
+
+    Object.values(_widgetDataCache).forEach(widgetTask => widgetTask.cancel());
```
kevinhinterlong: Supply a message to each .cancel call saying `dashboard ${id} canceled widget task ${widgetId}`?

packages/dashboards/addon/routes/dashboards/dashboard/view.js
daniellevinson: ok it's a little tricky to navigate directly between 2 dashboards but I've added a test that does this via the index route and also verifies cancelation. Will look further to find out why that happens as route/model does not await the requests anyway

packages/dashboards/addon/routes/dashboards/dashboard/view.js
daniellevinson: done

packages/dashboards/addon/routes/dashboards/dashboard/view.js
```diff
@@ -180,7 +176,20 @@ export default Route.extend({
...
+  },
+
+  actions: {
+    /**
+     * @override
+     * @action willTransition
+     * @param {Transition} transition
+     */
+    willTransition(transition) {
+      if (transition.targetName !== this.routeName) {
```
mattgraham1995: Does this get fired when a filter updates the query params for the route? if we update a filter while loading, will this cancel loading widgets?

daniellevinson: Good point, no test picked that up. I was sure when we update the filters we stay in the same route and the cancelation would be in the `model` hook if filters actually changed, but we actually transition to `dashboards.dashboard.index`. Need to make a small change here and add a test to make sure widgets still load, because right now when adding/removing an empty filter, it cancels loading widgets which we render from cache so nothing loads.

cythrawll:
If we aren't able to add that test today. let's go ahead and merge this.

PR #677, opened 2020-02-24
Update: BardLite - Allow any key in dimension value

balajimanoharan:
Update: BardLite - Allow any key in dimension value beyond just id and description

## License

### DIFF
--- a/packages/data/addon/mirage/routes/bard-lite.js
+++ b/packages/data/addon/mirage/routes/bard-lite.js
@@ -187,14 +187,15 @@ export default function(
 
         return newRows.concat(
           dimensionValues.map(value => {
+            let newRow = Object.assign({}, currentRow);
             Object.keys(value).forEach(key => {
               if (key === 'description') {
-                currentRow[`${dimension}|desc`] = value[key];
+                newRow[`${dimension}|desc`] = value[key];
               } else {
-                currentRow[`${dimension}|${key}`] = value[key];
+                newRow[`${dimension}|${key}`] = value[key];
               }
             });
-            return currentRow;
+            return newRow;
           })
         );
       }, []);
--- a/packages/data/addon/mirage/routes/bard-lite.js
+++ b/packages/data/addon/mirage/routes/bard-lite.js
@@ -7,7 +7,6 @@
 
 import { faker, Response } from 'ember-cli-mirage';
 import moment from 'moment';
-import { assign } from '@ember/polyfills';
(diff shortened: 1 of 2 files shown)


packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -187,13 +187,18 @@ export default function(
...
+          dimensionValues.map(value => {
             // TODO figure out why Object.assign refuses to work in Phantom even with Babel polyfill
-            assign({}, currentRow, {
-              [`${dimension}|id`]: value.id,
+            let newValue = assign({}, currentRow, {
               [`${dimension}|desc`]: value.description
-            })
-          )
+            });
+            const keys = Object.keys(value).filter(key => key !== 'description');
```
kevinhinterlong: Maybe just `.map` with an `if(key === description)`

packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -188,10 +187,12 @@ export default function(
 
         return newRows.concat(
           dimensionValues.map(value =>
-            // TODO figure out why Object.assign refuses to work in Phantom even with Babel polyfill
-            assign({}, currentRow, {
-              [`${dimension}|id`]: value.id,
-              [`${dimension}|desc`]: value.description
+            Object.keys(value).map(key => {
```
mattgraham1995: We're using map but not returning anything in the map callback. Is that the functionality we want?

kevinhinterlong: In addition to this, it should map each  key to an object then reduce to one object and use assign. Or just replace the map with forEach

balajimanoharan: ummm.. ya.. my bad.. fixed now.

packages/data/addon/mirage/routes/bard-lite.js
```diff
@@ -187,13 +186,16 @@ export default function(
...
-          )
+          dimensionValues.map(value => {
+            Object.keys(value).forEach(key => {
+              if (key === 'description') {
+                currentRow[`${dimension}|desc`] = value[key];
+              } else {
+                currentRow[`${dimension}|${key}`] = value[key];
+              }
+            });
+            return currentRow;
```
mattgraham1995: Won't this result in multiple copies of `currentRow` being returned in the `dimensionValues.map`?

PR #687, opened 2020-02-28
Adding dataSource property to request

### LINKED ISSUE #690: Add dataSource to requests

cythrawll:
Resolves #690

## Description

Adding a new property to request fragment meant to keep track which datasource the request is meant for.

## Proposed Changes

- Adding property to request fragment
- Make sure that it works in cloning.
- Make sure that switching logical tables, switches datasource.

## License


packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/fragments/Request.kt
```diff
@@ -27,6 +27,7 @@ data class Request(
     var sort: Array<Sort> = arrayOf(),
     var having: Array<Having> = arrayOf(),
 
+    var dataSource: String = "",
```
kevinhinterlong: Should this have a value? Seems like as is it would make navi use `''` as `dataSource` instead of letting it fall back to default

packages/reports/tests/unit/routes/reports/new-test.js
```diff
@@ -15,6 +15,7 @@ const NEW_MODEL = {
   createdOn: null,
   request: {
     bardVersion: 'v1',
+    dataSource: 'dummy',
```
kevinhinterlong: Maybe add an explicit test somewhere that if `dataSource` wasn't defined it will fall back to default correctly

packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/fragments/Request.kt
cythrawll: You're right, will fix

packages/reports/tests/unit/routes/reports/new-test.js
cythrawll: That's actually what this test covers. that line is the expected model. the _newModel test creates a model without passing in a datasource.

packages/core/addon/models/bard-request/request.js
```diff
@@ -82,6 +83,7 @@ export default Fragment.extend(Validations, {
   intervals: fragmentArray('bard-request/fragments/interval', {
     defaultValue: []
   }),
+  dataSource: DS.attr('string', { defaultValue: getDefaultDataSourceName() }),
```
balajimanoharan: Would the saved request fragments without the `datasource` prop still work?

cythrawll: Yes.

PR #704, opened 2020-03-06
Fix: "Start date should be before end date" error for same date interval

daniellevinson:
## Description

The issue is not with the error being displayed but with the rendered end date which is inconsistent with the error.
When rendering the end date we subtract 1 from the interval's actual end date.
In the scenario where the user chooses a start date (e.g. May 16) which is only 1 time grain beyond the displayed end date (May 15):
- Interval has equal dates now (May 16 - May 16) which correctly displays an error (a valid range is start < end).
- Since `asMomentsForTimePeriod` adds 1 to the displayed end date if dates are equal the user sees May 16 - May 16 instead of May 16 - May 15.
- Choose a different start date and the end date updates back to May 15.

## Proposed Changes

Add an argument to `asMomentsForTimePeriod` which indicates whether end date should be inclusive if dates are equal, defaults to `true`. Pass `false` when rendering the range.

## Screenshots

Easy to see on demo app. Pretty positive it's the only scenario where this is happening.

## License


packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1653,6 +1653,116 @@ module('Acceptance | Navi Report', function(hooks) {
...
+
+    await visit('/reports/1');
+    await selectChoose('.filter-builder__select-trigger', 'Between');
+
+    assert.dom('.filter-values--date-range-input__low-value').hasText('Nov 09, 2015', 'The start date is Nov 09, 2015');
+    assert.dom('.filter-values--date-range-input__high-value').hasText('Nov 15, 2015', 'The end date is Nov 15, 2015');
+
+    //start date beyond end date
+    await clickTrigger('.filter-values--date-range-input__low-value');
+    await click($('.ember-power-calendar-day--current-month:contains(18)')[0]);
```
mattgraham1995: Can you check if there's an attribute property with the day value on this element? We should try to avoid adding more jquery where possible.

packages/reports/tests/acceptance/navi-report-test.js
```diff
@@ -1653,6 +1653,116 @@ module('Acceptance | Navi Report', function(hooks) {
...
+
+    assert
+      .dom('.navi-info-message__error-list')
+      .hasText(
+        'The start date should be before end date',
+        '"The start date should be before end date" error is rendered when start date is 1 day beyond end date'
+      );
+
+    //select month grain
+    await clickItem('timeGrain', 'Month');
```
kevinhinterlong: Maybe make a test for `week`, `year`, etc.

Also heads up depending on if my column config item PR gets merged in I moved the `navi-report-test` to it's own folder and made a another test file so we can start splitting it up by what's being tested

PR #706, opened 2020-03-06
Fix: Revert tracked query params

kevinhinterlong:
## Description
- Query params are technically owned (managed) by the ember framework so we should be using set to notify them to update
- https://discuss.emberjs.com/t/query-params-tracked/17467/21
- https://github.com/emberjs/ember.js/issues/18715

^^ we saw the same issue above where putting `@tracked` made the page unusable

## Proposed Changes
Revert them from `@tracked` to just plain properties

## License

### DIFF
--- a/packages/directory/addon/controllers/directory.js
+++ b/packages/directory/addon/controllers/directory.js
@@ -3,8 +3,7 @@
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 import Controller from '@ember/controller';
-import { tracked } from '@glimmer/tracking';
-import { getProperties, get, computed, action } from '@ember/object';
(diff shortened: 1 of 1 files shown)


kevinhinterlong:
@jkusa Any ideas on how to test this? It seems like it still _technically_ works, but is just really slow and I don't think we have any performance tests yet.

Something like this?

```js
assert.expect(1);

// generate 100 reports

fillIn('.searchBar', 'search')'
triggerEvent('.searchBar', 'focusout');

// wait 2s

assert.deepEqual(
  findAll('.results').map(el => el.textContent.trim()),
  ['results'],
  'The search finished'
)
```

jkusa:
Maybe this could be useful:
https://api.qunitjs.com/assert/timeout

PR #717, opened 2020-03-16
Update: Specify search field for table selector

ldkge:
Resolves #

## Description
Add the searchField property to the navi-table-select so that you can specify which field you want to use for search when search is enabled.

## Proposed Changes

- Add searchField property.

## License

### DIFF
--- a/packages/reports/addon/components/navi-table-select.js
+++ b/packages/reports/addon/components/navi-table-select.js
@@ -18,7 +18,7 @@ import layout from '../templates/components/navi-table-select';
 
 @templateLayout(layout)
 @tagName('')
-export default class NaviTableSelectComponent extends Component {
+class NaviTableSelectComponent extends Component {
   /**
    * @property {Boolean} searchEnabled
    */
@@ -29,3 +29,5 @@ export default class NaviTableSelectComponent extends Component {
    */
   searchField = 'longName';
 }
+
+export default NaviTableSelectComponent;
--- a/packages/reports/addon/components/navi-table-select.js
+++ b/packages/reports/addon/components/navi-table-select.js
@@ -7,6 +7,7 @@
  *          selected=selected
  *          onChange=onChange
  *          searchEnabled=searchEnabled
+ *          searchField=searchField
  *          tagName=tagName
  *        }}
  */
@@ -22,5 +23,10 @@ export default Component.extend({
   /**
    * @property {Boolean} searchEnabled
    */
-  searchEnabled: false
+  searchEnabled: false,
+
+  /**
+   * @property {String} searchField
+   */
+  searchField: 'longName'
 });
--- a/packages/reports/addon/components/navi-table-select.js
+++ b/packages/reports/addon/components/navi-table-select.js
@@ -1,15 +1,15 @@
 /**
- * Copyright 2017, Yahoo Holdings Inc.
+ * Copyright 2020, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
- * Usage: {{navi-table-select
- *          options=options
- *          selected=selected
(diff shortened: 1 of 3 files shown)


packages/reports/addon/components/navi-table-select.js
```diff
@@ -7,6 +7,7 @@
  *          selected=selected
  *          onChange=onChange
  *          searchEnabled=searchEnabled
+ *          searchField=searchField
```
kevinhinterlong: Can we make this a native class, update docs/copyright

packages/reports/addon/templates/components/navi-table-select.hbs
```diff
@@ -5,6 +5,7 @@
   @options={{@options}}
   @selected={{@selected}}
   @searchEnabled={{searchEnabled}}
+  @searchField={{searchField}}
```
kevinhinterlong:
```suggestion
  @searchEnabled={{this.searchEnabled}}
  @searchField={{this.searchField}}
```

kevinhinterlong:
Maybe add a test for searching by `longName`?

packages/reports/addon/components/navi-table-select.js
```diff
@@ -22,5 +23,10 @@ export default Component.extend({
   /**
    * @property {Boolean} searchEnabled
    */
-  searchEnabled: false
+  searchEnabled: false,
```
kevinhinterlong: Can we change this component to be a native class?

packages/reports/tests/integration/components/navi-table-select-test.js
```diff
@@ -58,15 +58,45 @@ module('Integration | Component | navi table select', function(hooks) {
...
+      'All options are shown'
+    );
+
+    await fillIn('.ember-power-select-search-input', '2');
+    assert.deepEqual(
+      findAll('.ember-power-select-option').map(el => el.textContent.trim()),
+      ['network2'],
+      'Filtered options are shown'
+    );
+  });
```
colemanProjects: can you add a test to clear, or fillIn empty. and all options should be shown again. 

and another test, to make sure if the input doesn't match, the appropriate thing is shown

packages/reports/addon/components/navi-table-select.js
ldkge: what do we mean by native class?

mattgraham1995: Line 18. `export default Class NaviTableSelect extends Component {`

kevinhinterlong:
```js
import { layout as templateLayout, tagName } from '@ember-decorators/component';

@templateLayout(layout)
@tagName('')
class NativeTableSelectComponent extends Component {
  /**
   * @property {Boolean} searchEnabled
   */
  searchEnabled = false;

  /**
   * @property {String} searchField
   */
  searchField = 'longName';
}

export default NativeTableSelectComponent;
```

Then in the template
```hbs
<div class="navi-table-select" ...attributes>
  {{!-- rest of template here --}}
</div>
```

ldkge: ohhh I got what you mean now

packages/reports/addon/components/navi-table-select.js
```diff
@@ -1,26 +1,33 @@
...
- *          searchEnabled=searchEnabled
- *          tagName=tagName
- *        }}
+ * Usage: <NaviTableSelect
+ *          @options={{options}}
+ *          @selected={{selected}}
+ *          @onChange={{onChange}}
+ *          @searchEnabled={{searchEnabled}}
+ *          @searchField={{searchField}}
+ *          @tagName={{tagName}}
```
jkusa:
```suggestion
```

PR #720, opened 2020-03-17
Update: Add column config to error route

jkusa:
## Description

Add column config to error route

## Proposed Changes

- Add column config to error route
- Replace `{{with}}` with `{{let}}`

## Screenshots

Screen Shot 2020-03-17 at 9 08 03 AM

## License


packages/reports/tests/acceptance/navi-report/navi-report-test.js
```diff
@@ -1187,32 +1187,30 @@ module('Acceptance | Navi Report', function(hooks) {
   });
 
   test('Error data request', async function(assert) {
-    assert.expect(1);
+    const originalFeatureFlag = config.navi.FEATURES.enableRequestPreview;
```
kevinhinterlong: Can we leave the original test as is and copy this over to `navi-report/navi-column-config-test`

jkusa: Hmm, seems the new asserts don't warrant a new test. `navi-report-test.js` is huge and should be broken into several use-case related files. I will start this with moving this test into an error route acceptance test file.

PR #722, opened 2020-03-19
Column config fixups

kevinhinterlong:
## Description
Resolves some small issues we've found in the column config
- Font family not being applied so bold doesn't work as expected
- Searching a metrics parameters is not working
- Column relabeling was visible in the UI, but not functional so we've hidden it for now

## License

### DIFF
--- a/packages/reports/addon/templates/components/navi-column-config/metric.hbs
+++ b/packages/reports/addon/templates/components/navi-column-config/metric.hbs
@@ -1,14 +1,14 @@
 {{!-- Copyright 2020, Yahoo Holdings Inc. Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms. --}}
 {{#if this.column.fragment.metric.hasParameters}}
   <div class="navi-column-config-base__option navi-column-config-metric__parameter">
-    <label for="{{this.classId}}-columnParameter" class="navi-column-config-metric__parameter-label navi-column-config-base__option-label">Metric Parameter</label>
+    <label for="{{this.classId}}-columnParameter" class="navi-column-config-metric__parameter-label navi-column-config-base__option-label">{{capitalize this.paramName}} Type</label>
     <PowerSelect 
       @triggerClass="navi-column-config-metric__parameter-trigger navi-column-config-base__option-input"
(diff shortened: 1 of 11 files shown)


packages/reports/addon/templates/components/navi-column-config/metric.hbs
```diff
@@ -7,7 +7,8 @@
       @triggerId="{{this.classId}}-columnParameter" 
       @dropdownClass="navi-column-config-metric__parameter-dropdown"
       @selected={{this.currentParameter}}
-      @searchPlaceholder="Type to search"
+      @searchField="id"
+      @searchPlaceholder="Type to search by id"
```
jkusa: Can this be `Search ${paramName}`?

packages/reports/addon/templates/components/navi-column-config/metric.hbs
```diff
@@ -7,7 +7,8 @@
       @triggerId="{{this.classId}}-columnParameter" 
       @dropdownClass="navi-column-config-metric__parameter-dropdown"
       @selected={{this.currentParameter}}
-      @searchPlaceholder="Type to search"
+      @searchField="id"
```
jkusa: Why `id over `description`?

kevinhinterlong: We can change it to be description in both places, although id's being unique makes them nice for searching
https://github.com/yahoo/navi/blob/950c05b03d3d05a8c64ca8eee56cde79148bc852/packages/reports/addon/templates/components/metric-config.hbs#L18

packages/reports/addon/templates/components/navi-column-config/metric.hbs
kevinhinterlong: What does paramName refer to, `paramName = Revenue` or `paramName = description`?

packages/reports/addon/templates/components/navi-column-config/metric.hbs
jkusa: I agree about unique ID values. Just like most dimensions, users like to search by nice names. I don't think users will not really care about the param ID values.

PR #724, opened 2020-03-20
Update: Ember-composable-helpers to 3.1.1

kevinhinterlong:
## Description
`ember-composable-helpers` to `3.1.1` 

## License

### DIFF
--- a/packages/core/app/helpers/ember-array.js
+++ b/packages/core/app/helpers/ember-array.js
@@ -1,1 +1,9 @@
-export { default, emberArray } from 'ember-composable-helpers/helpers/array';
+import { helper } from '@ember/component/helper';
+import { A as emberArray } from '@ember/array';
+
+export function array(params = []) {
+  // slice params to avoid mutating the provided params
+  return emberArray(params.slice());
+}
+
+export default helper(array);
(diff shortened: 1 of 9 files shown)


packages/core/app/helpers/ember-array.js
```diff
@@ -1 +1,9 @@
-export { default, emberArray } from 'ember-composable-helpers/helpers/array';
+import { helper } from '@ember/component/helper';
```
jkusa: Do we still need this? GH shows only one usage of `ember-array`. Can we just use: https://api.emberjs.com/ember/3.17/classes/Ember.Templates.helpers/methods/array?anchor=array

kevinhinterlong: Yes we still use it at least in the `report-visualization`
https://github.com/yahoo/navi/blob/cec92ce52d1937f10a63633c48cb096b703a37c4/packages/reports/addon/templates/components/report-visualization.hbs#L5
So this would require refactoring our `navi-visualizations/`

We can make an issue to refactor them

kevinhinterlong: https://github.com/yahoo/navi/issues/726

PR #730, opened 2020-03-24
Update: Distinguish between dimensions and metrics in column config

daniellevinson:
## Description

Distinguish between dimensions and metrics in column config

## Screenshots

Mar-24-2020 16-42-32

## License


daniellevinson:
Update: Using `<NaviIcon @icon="minus-circle" />` instead of ⊖ character, the latter renders larger and uglier in firefox with any of our fonts.

Mar-24-2020 20-54-11

PR #734, opened 2020-03-27
Update: Add multi-datasource dashboard support

### LINKED ISSUE #666: Add multi datasource dashboard support
Allow dashboards to run widgets from multiple data sources.

cythrawll:
Resolves #666 also known as the issue of the beast

## Description

Adds Multi datasource support to dashboards

## Proposed Changes

- Update fetch to pass dataSource of widget to adapter
- Update global filters to specify the dataSource and carry it around
- Make sure dashboard serializer and cloning includes dataSource in dimension name
- Global dashboard dashboard dimension selector specifies dataSource in tooltip and resolves conflicting categories.

## Screenshots
Screen Shot 2020-03-27 at 4 39 20 PM

## License

### DIFF
--- a/packages/core/addon/serializers/dashboard.js
+++ b/packages/core/addon/serializers/dashboard.js
@@ -22,11 +22,24 @@ export default AssetSerializer.extend({
     return this._super(...arguments);
   },
 
+  /**
+   * Overrides default serialize method to add datasources to filter object
+   *
+   * @param {Snapshot} snapshot
+   * @returns {Object} serialized dashboard
+   */
   serialize(snapshot) {
-    const filterSources = snapshot.attr('filters').map(filter => filter.attr('dimension').source);
+    const buildKey = (dimension, values) => `${dimension}[${values.join(',')}]`;
+    const filterSources = snapshot.attr('filters').reduce((dimensionSources, filter) => {
+      dimensionSources[buildKey(filter.attr('dimension').name, filter.attr('rawValues'))] = filter.attr(
+        'dimension'
+      ).source;
+      return dimensionSources;
+    }, {});
     const dashboard = this._super(...arguments);
     dashboard.data.attributes.filters = dashboard.data.attributes.filters.map((filter, i) => {
-      filter.dimension = filterSources[i] ? `${filterSources[i]}.${filter.dimension}` : filter.dimension;
+      const key = buildKey(filter.dimension, filter.values);
+      filter.dimension = filterSources[key] ? `${filterSources[key]}.${filter.dimension}` : filter.dimension;
       return filter;
     });
     return dashboard;
--- a/packages/core/addon/serializers/dashboard.js
+++ b/packages/core/addon/serializers/dashboard.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019, Yahoo Holdings Inc.
+ * Copyright 2020, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 
@@ -20,6 +20,16 @@ export default AssetSerializer.extend({
       return this._super(type, newPayload);
     }
     return this._super(...arguments);
+  },
+
+  serialize(snapshot) {
+    const filterSources = snapshot.attr('filters').map(filter => filter.attr('dimension').source);
+    const dashboard = this._super(...arguments);
+    dashboard.data.attributes.filters = dashboard.data.attributes.filters.map((filter, i) => {
+      filter.dimension = filterSources[i] ? `${filterSources[i]}.${filter.dimension}` : filter.dimension;
+      return filter;
+    });
+    return dashboard;
   },
 
   /**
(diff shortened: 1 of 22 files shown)


packages/core/addon/serializers/dashboard.js
```diff
@@ -22,6 +22,16 @@ export default AssetSerializer.extend({
     return this._super(...arguments);
   },
 
+  serialize(snapshot) {
```
mattgraham1995: Jsdoc

packages/core/addon/serializers/dashboard.js
```diff
@@ -22,6 +22,16 @@ export default AssetSerializer.extend({
...
+    const filterSources = snapshot.attr('filters').map(filter => filter.attr('dimension').source);
+    const dashboard = this._super(...arguments);
+    dashboard.data.attributes.filters = dashboard.data.attributes.filters.map((filter, i) => {
+      filter.dimension = filterSources[i] ? `${filterSources[i]}.${filter.dimension}` : filter.dimension;
+      return filter;
+    });
+    return dashboard;
```
mattgraham1995:
```suggestion
    const filterSources = snapshot.attr('filters').reduce((dimensionSources, filter) => { 
      const { name, source } = filter.attr('dimension');
      dimensionSources[name] = [...(dimensionSources[name] || []), source]; 
      return dimensionSources
    }, {});
    const dashboard = this._super(...arguments);
    dashboard.data.attributes.filters = dashboard.data.attributes.filters.map((filter, i) => {
      const { dimension } = filter;
      filter.dimension = filterSources[dimension] ? `${filterSources[dimension]}.${dimension}` : dimension;
      return filter;
    });
    return dashboard;
```
Could we do something like this so we're not relying on the order of each list before and after calling super? I think it might be more robust if we use the dimension id as the key for assigning the source in the filters.

packages/dashboards/addon/controllers/dashboards/dashboard/view.js
```diff
@@ -44,26 +44,36 @@ export default Controller.extend(ReportToWidget, {
      */
     async updateFilter(dashboard, originalFilter, changeSet) {
       const origFilter = originalFilter.serialize();
+      origFilter.dataSource = originalFilter.dimension.source;
       const newFilters = get(dashboard, 'filters')
         .toArray()
-        .map(fil => fil.serialize()); //Native array of serialized filters
+        .map(fil => {
+          const newFil = fil.serialize();
```
mattgraham1995: why doesn't the serialize function keep the data source in the dimension?

packages/dashboards/addon/templates/components/dashboard-dimension-selector.hbs
```diff
@@ -9,6 +9,6 @@
     beforeOptionsComponent="power-select-search"
     as |dimension|
   }}
-    {{dimension.longName}}
+    <div title="{{dimension.dataSource}}">{{dimension.longName}}</div>
```
mattgraham1995:
```suggestion
    <div title="Source: {{dimension.dataSource}}">{{dimension.longName}}</div>
```
Maybe add a little context to the tooltip?

packages/dashboards/tests/acceptance/mutli-datasource-dashboards-test.js
```diff
@@ -0,0 +1,110 @@
...
+      ['Age', 'Container'],
+      'Dimensions are properly labeled in filters'
+    );
+
+    assert.deepEqual(
+      findAll('.filter-builder-dimension__operator').map(el => el.textContent.trim()),
+      ['Equals', 'Not Equals'],
+      'Dimension filter operators are shown correctly'
+    );
+    //ember-power-select-multiple-remove-btn
```
mattgraham1995:
```suggestion
```

packages/dashboards/addon/controllers/dashboards/dashboard/view.js
cythrawll: Because it's not part of the serialization

packages/core/addon/serializers/dashboard.js
cythrawll: Problem is that we need to account for different datasources that have the same dimension name. Because we might have dimensions with the same base name but different valid values to filter by.

Best I could do is build a key with the dimension name AND values.

packages/dashboards/tests/acceptance/mutli-datasource-dashboards-test.js
```diff
@@ -0,0 +1,109 @@
...
+            '.ember-power-select-multiple-option span:not(.ember-power-select-multiple-remove-btn)'
+          )
+        ].map(el => el.textContent.trim())
+      ),
+      [['under 13 (1)', '13-17 (2)', '18-20 (3)'], ['Bag (1)']],
+      'Dimension value selector showing the right values'
+    );
+  });
+
+  test('Create new multisource daashboard', async function(assert) {
```
daniellevinson:
```suggestion
  test('Create new multisource dashboard', async function(assert) {
```

PR #736, opened 2020-04-01
Update: Lift column config to report route

kevinhinterlong:
## Description
While playing around with the column config, if you move things around/open/close/etc those changes are destroyed every time you transition to a new report route such as when running a report/etc

## Proposed Changes
Lift up the column config to the report route level so that it persists between transitions

## Screenshots
persisted column config

## License

### DIFF
--- a/packages/reports/tests/integration/components/report-view-test.js
+++ b/packages/reports/tests/integration/components/report-view-test.js
@@ -2,13 +2,11 @@ import { A } from '@ember/array';
 import { helper as buildHelper } from '@ember/component/helper';
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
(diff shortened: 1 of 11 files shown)


packages/reports/tests/integration/components/report-view-test.js
```diff
@@ -211,49 +209,4 @@ module('Integration | Component | report view', function(hooks) {
       .dom('.report-view__visualization-no-results')
       .hasText('No results available.', 'A message is displayed when the response has no data');
   });
-
-  test('toggle columns drawer', async function(assert) {
```
cythrawll, line 215: Is this test covered elsewhere now?

kevinhinterlong: Good call, I moved it up now

PR #742, opened 2020-04-06
Fix: Advanced date picker

kevinhinterlong:
## Description
The advanced date picker currently adds an extra period to the end of the interval, but this is confusing as it's not what the user typed in

## Proposed Changes
Update the advanced date picker so that it does not modify the date from user input

## License

### DIFF
--- a/packages/reports/tests/acceptance/report-test.js
+++ b/packages/reports/tests/acceptance/report-test.js
@@ -1985,6 +1985,38 @@ module('Acceptance | Navi Report', function(hooks) {
       .hasText('May 30, 2015', 'Calendar defaults "all" grain  to show the lowest grain which is day');
   });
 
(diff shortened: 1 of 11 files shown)


packages/reports/tests/acceptance/report-test.js
```diff
@@ -1985,6 +1985,38 @@ module('Acceptance | Navi Report', function(hooks) {
       .hasText('May 30, 2015', 'Calendar defaults "all" grain  to show the lowest grain which is day');
   });
 
+  test('Date picker advanced doesnt modify interval', async function(assert) {
```
mattgraham1995, line 1988:
```suggestion
  test("Date picker advanced doesn't modify interval", async function(assert) {
```
sorry

PR #744, opened 2020-04-07
New: Request v2 model fragments

daniellevinson:
⚠️ This is not a breaking change but will not be merged into master

## Description

Model request v2

Not included in this PR:
- Adding metadata that we currently do in transformations
- Multi datasource support (easy to add if we decide to leave it where it is now)
- Model methods used by consumers

## License

### DIFF
--- a/packages/core/addon/models/bard-request-v2/fragments/filter.js
+++ b/packages/core/addon/models/bard-request-v2/fragments/filter.js
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import attr from 'ember-data/attr';
+import { validator, buildValidations } from 'ember-cp-validations';
+import BaseFragment from './base';
+
+const Validations = buildValidations({
+  operator: validator('presence', {
+    presence: true,
+    message: 'The `operator` filter field cannot be empty'
+  }),
+  values: [
+    validator('length', {
+      min: 1,
+      allowNone: false,
+      message() {
+        const { field } = this.model;
+        return `${field} filter must have at least one value`;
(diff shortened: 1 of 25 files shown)


packages/core/addon/models/bard-request-v2/request.js
```diff
@@ -0,0 +1,96 @@
...
+
+export default Fragment.extend(Validations, {
+  filters: fragmentArray('bard-request-v2/fragments/filter', { defaultValue: [] }),
+  columns: fragmentArray('bard-request-v2/fragments/column', { defaultValue: [] }),
+  table: attr('string'),
+  sort: fragmentArray('bard-request-v2/fragments/sort', { defaultValue: [] }),
+  limit: attr('number'),
+  requestVersion: attr('string', { defaultValue: '2.0' }),
+  dataSource: attr('string', { defaultValue: getDefaultDataSourceName() }),
+  responseFormat: attr('string', { defaultValue: 'json' }),
```
cythrawll, line 55: Do we need this, is this part of the spec?

daniellevinson: I wasn't sure but looks like it can be removed

jkusa: We can remove

packages/core/addon/models/bard-request-v2/fragments/filter.js
```diff
@@ -0,0 +1,36 @@
...
+  values: [
+    validator('length', {
+      min: 1,
+      allowNone: false,
+      message() {
+        const { field } = this.model;
+        return `${field} filter must have at least one value`;
+      }
+    }),
+    validator('array-empty-value', {
```
jkusa, line 23: Going forward I think we can allow this. We went away from checking this in dashboards.

PR #753, opened 2020-04-10
New: Opensource GraphQL adapter

mattgraham1995:
## Description
Navi is quickly aligning its metadata to work with Elide. An adapter is necessary to actually query Elide for metadata through GraphQL.

## Proposed Changes

- Add the previously closed-sourced Elide metadata adapter to navi-data
- Add packages for using GraphQL in navi-data

## License


packages/data/addon/adapters/elide-metadata.js
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2020, Verizon Media
```
cythrawll, line 2: Wrong ©️

packages/data/addon/serializers/elide-metadata.js
```diff
@@ -0,0 +1,13 @@
+import EmberObject from '@ember/object';
+
+export default class GraphQLMetadataSerializer extends EmberObject {
+  /**
+   * @method normalize
+   * normalizes the graphql metadata response
+   * @param payload {Object} - graphql response payload
+   * @returns {Object} - normalized JSON object
+   */
+  normalize(/*{ data: tables = [] }*/) {
```
cythrawll, line 10: shouldn't this return the argument unchanged instead of `void`?

packages/data/addon/services/navi-metadata-apollo.js
```diff
@@ -0,0 +1,76 @@
...
+
+    return headersLink.concat(httpLink);
+  }
+
+  /**
+   * @method _buildURLPath
+   * @returns {String} complete url including host and namespace
+   */
+  _buildURLPath() {
+    const host = getHost();
```
cythrawll, line 71: use `configHost` rather than `getHost` and pass options through here.

packages/data/tests/dummy/mirage/config.js
```diff
@@ -1,11 +1,16 @@
...
 export default function() {
   this.passthrough('/write-coverage');
+
   // Mock bard facts + metadata
   for (let dataSource of config.navi.dataSources) {
     this.urlPrefix = `${dataSource.uri}/v1`;
     BardMeta.call(this);
   }
+
+  this.urlPrefix = `${config.navi.dataSources[0].uri}/v1`;
```
cythrawll, line 14: Can we loop through dataSources to create multiple GraphQL datasource mocks?

packages/data/addon/services/navi-metadata-apollo.js
mattgraham1995: I think we may have to come back to this due to the way that the apollo client is instantiated. It seems like it's not meant to change the api url after instantiating the service so it may be best to define a separate service for each data source. That could make it simpler if there are different configurations needed for each data source's requests as well.

packages/data/addon/adapters/elide-metadata.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  /**
+   * @method fetchMetadata
+   * @param {String} type
+   * @param {String} id
+   * @param {Object} options
+   * @returns {Object} JSON response for type
+   */
+  async fetchMetadata(type, id, options = {}) {
```
jkusa, line 34: Maybe we can clean up the general adapter interface long term. `fetchAll` & `fetchById` seem like better names.

mattgraham1995: I agree. I just wanted to get what we had close-sourced into the open-source repo first. It's a big PR so I didn't want to fill it with lots of changes that would make it harder to review.

jkusa: Yeah, just a note for future work around navi-data

PR #756, opened 2020-04-15
Update: Small UX cleanup for column config

kevinhinterlong:
## Description
Some small cleanup
- missing metric selector icon
- smaller metric/dimension selector
- less margin for visualization
- don't show parameters id

## Screenshots

Before | After
before after

## License


kevinhinterlong:
show id if needed

jkusa:
Are the metric/dimension indicators centered with the text?

PR #766, opened 2020-04-21
Update: Split loadMetadata into multiple functions

mattgraham1995:
## Description
`loadMetadata` in the `bard-metadata` service is difficult to override due to the length of the function.

## Proposed Changes

- Create more private functions and delegate the loadMetadata functionality to them so that overriding the bard-metadata service in an app is much easier and cleaner.

## License


packages/data/addon/services/bard-metadata.js
```diff
@@ -65,45 +65,58 @@ export default class BardMetadataService extends Service {
...
-      const payload = await this._adapter.fetchAll(
-        'table',
-        assign(
-          {
-            query: { format: 'fullview' }
-          },
-          options
-        )
-      );
+      const payload = await this._fetchMetadataForSource(options);
```
balajimanoharan, line 68: Can we call this `_fetchMetadata` since we are not just passing the source?

PR #773, opened 2020-04-23
GraphQL Serializer and Schema Updates

### LINKED ISSUE #648: Elide Aggregation Adapter - Metadata Serializer
Implement serializer that will normalize responses from Elide GraphQL metadata requests.

mattgraham1995:
Resolves #648 

## Description
We need to normalize responses from the graphql metadata endpoint so that we can construct navi-data models.
We also need to make sure the test schema we use in navi-data is close to the actual schema in Elide.

## Proposed Changes

- Create `elide-metadata` serializer that can normalize a `tables` query response from the `elide-metadata` adapter
- update graphql testing data schema to better match elide 
- `navi-data/models/metadata/table.cardinalitySize  ---> cardinality`
- Create typescript files with types defined for raw graphql responses and normalized forms

## License


packages/data/addon/serializers/elide-metadata.ts
```diff
@@ -0,0 +1,181 @@
...
+  timeZone: TODO;
+};
+
+export default class ElideMetadataSerializer extends EmberObject {
+  /**
+   * Transform the elide metadata into a shape that our internal data models can use
+   * @private
+   * @method _normalizeTableConnection - normalizes the table connection object
+   * @param {Object} tableConnection - table connection with array of edges to table nodes
+   * @param {String} source - datasource of the payload
```
kevinhinterlong, line 52: `{String}` -> `{string}`

packages/data/addon/serializers/elide-metadata.ts
```diff
@@ -0,0 +1,181 @@
...
+  category: string;
+  description: string;
+  cardinality: string;
+  metricIds: string[];
+  dimensionIds: string[];
+  timeDimensionIds: string[];
+  source: string;
+};
+
+type NormalizedColumn = {
```
kevinhinterlong, line 28: Are these the Payloads used to construct `Metric`, `Dimension`, etc or are they specific to this part? 

It would be nice to move them to their metadata files, so we can use them in the bard-metadata serializer as well

packages/data/addon/models/metadata/column.ts
```diff
@@ -8,6 +8,15 @@ import { inject as service } from '@ember/service';
 import KegService from '../../services/keg';
 import Table from './table';
 
+export type ColumnNode = {
```
jkusa, line 11: Does it make sense to put the raw graphql types in the model file? Seems like more of a serializer thing.

packages/data/addon/models/metadata/column.ts
```diff
@@ -9,6 +9,16 @@ import KegService from '../../services/keg';
 import Table from './table';
 
 export type ColumnType = 'ref' | 'formula' | 'field';
+export type NormalizedColumn = {
```
jkusa, line 12: Can we want to make this:
```ts
interface ColumnMetadata {
...
}
```
and make the model:
```ts
export default class ColumnMetadataModel extends EmberObject implements ColumnMetadata { 
```

packages/data/addon/models/metadata/column.ts
```diff
@@ -9,22 +9,22 @@ import KegService from '../../services/keg';
...
   tags: string[];
-};
+}
 
 export type BaseExtendedAttributes = {
   description?: string;
 };
 
-export default class Column extends EmberObject {
+export default class Column extends EmberObject implements ColumnMetadata {
```
jkusa, line 27: Just to be clear, can we name the models in this PR like this:
```suggestion
export default class ColumnMetadataModel extends EmberObject implements ColumnMetadata {
```

mattgraham1995: Oh sorry missed the model name change

packages/data/addon/models/metadata/column.ts
```diff
@@ -9,12 +9,22 @@ import KegService from '../../services/keg';
 import Table from './table';
 
 export type ColumnType = 'ref' | 'formula' | 'field';
+export interface ColumnMetadata {
```
kevinhinterlong, line 12: This is kind of unclear to me. The name makes it sound like this is what we expect a column metadata to act/look like, but in our usage it appears to be the shape of the payload needed to construct a column

jkusa: True, we probablly should use constructors/factories to build models and use interfaces to reflect public constracts.

PR #783, opened 2020-04-24
Update: Octane filter builders

### LINKED ISSUE #782: Reports components/filter-builders

kevinhinterlong:
Resolves #782 

## Description
Convert filter builder components to octane

## License

### DIFF
--- a/packages/reports/addon/components/filter-builders/base.js
+++ b/packages/reports/addon/components/filter-builders/base.js
@@ -1,23 +1,19 @@
 /**
- * Copyright 2019, Yahoo Holdings Inc.
+ * Copyright 2020, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Base class for filter builders.
  */
 
 import Component from '@ember/component';
-import { get, computed } from '@ember/object';
+import { action, computed } from '@ember/object';
 import layout from 'navi-reports/templates/components/filter-builders/base';
 import { readOnly } from '@ember/object/computed';
+import { layout as templateLayout, tagName } from '@ember-decorators/component';
 
-export default Component.extend({
-  layout,
-
-  /**
-   * @property {Array} classNames
-   */
-  classNames: ['filter-builder'],
-
+@templateLayout(layout)
+@tagName('')
+class BaseFilterBuilderComponent extends Component {
   /**
    * @property {Object} filter
    * @property {Object} filter.subject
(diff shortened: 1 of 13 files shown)


packages/reports/addon/components/filter-builders/base.js
```diff
@@ -27,42 +23,47 @@ export default Component.extend({
    * @property {String} filter.operator.valuesComponent - name of component to use for selecting filter values
    * @property {Array} filter.values
    */
-  filter: undefined,
+  get filter() {
```
jkusa, line 26: does this need to be a getter?

packages/reports/addon/components/filter-builders/base.js
```diff
@@ -27,42 +23,47 @@ export default Component.extend({
...
    * @property {String} displayName - display name for the filter
    */
-  displayName: readOnly('filter.subject.name'),
+  @readOnly('filter.subject.name') displayName;
 
   /**
    * @property {Array} supportedOperators - list of valid values for filter.operator
    */
-  supportedOperators: computed(function() {
+  @computed
```
jkusa, line 38: `supportedOperators = []`?

packages/reports/addon/components/filter-builders/date-dimension.js
```diff
@@ -1,28 +1,29 @@
...
-  requestFragment: undefined,
+  requestFragment = undefined;
 
   /**
    * @property {Array} supportedOperators
    * @override
    */
-  supportedOperators: computed(function() {
+  @computed
+  get supportedOperators() {
```
jkusa, line 26: literal?

packages/reports/addon/components/filter-builders/metric.js
```diff
@@ -1,26 +1,28 @@
...
-
+@templateLayout(layout)
+@tagName('')
+class MetricFilterBuilderComponent extends BaseFilterBuilderComponent {
   /**
    * @property {Object[]} supportedOperators
    */
-  supportedOperators: computed(function() {
+  @computed
+  get supportedOperators() {
```
jkusa, line 25: same

packages/reports/addon/components/filter-builders/metric.js
```diff
@@ -63,26 +65,28 @@ export default Base.extend({
...
+  requestFragment = undefined;
 
   /**
    * @property {String} displayName - display name for the filter with metric and parameters
    */
-  displayName: computed('filter.subject.{metric,parameters}', function() {
+  @computed('filter.subject.{metric,parameters}')
+  get displayName() {
     let metric = get(this, 'filter.subject');
     return metricFormat(metric, metric.metric?.name);
```
jkusa, line 81:
```suggestion
    const metric = this.filter.subject;
    return metricFormat(metric, metric.metric?.name);
```

packages/reports/addon/components/filter-builders/base.js
kevinhinterlong: I can try it with just `filter;`, but `filter = undefined` results in an error. I think it's trying to do `this.filter = undefined` during the constructor but in subclasses we have `get filter()` already defined, so it thinks we are trying to set a getter only property

kevinhinterlong: Ok coming back to this, if we do
```
filter = undefined; // breaks
filter; // breaks
// filter; // works, but confusing
get filter() { // works, slightly less confusing
  return undefined;
}
```

basically the same thing with `supportedOperators` so I'm opting to leave it as is

PR #784, opened 2020-04-27
Breaking: Move interval and related classes to navi-data

mattgraham1995:
## Description
With the Request V2.0, the request no longer directly contains an interval. This means that the navi data adapter will need to be able to derive an interval string from the dateTime filter. 

## Proposed Changes

- Move `classes/interval`, `classes/duration`, `date`, and `duration-utils` from navi-core to navi-data

## License


mattgraham1995:
May not be needed. Will reopen if necessary

PR #788, opened 2020-04-28
New: Bard Facts Adapter for Request V2

mattgraham1995:
## Description
The request V2 shape will not work with the existing bard-facts adapter.

## Proposed Changes

- Create bard-facts-v2 adapter to send requests to fili with the new v2 request

## License

### DIFF
--- a/packages/data/addon/adapters/bard-facts-v2.ts
+++ b/packages/data/addon/adapters/bard-facts-v2.ts
@@ -60,10 +60,10 @@ type AliasFn = (column: string) => string;
 
 /**
  * @function formatDimensionFieldName
- * @param {string} field - dimension id
- * @param {Parameters} parameters - parameters object
+ * @param field - dimension id
+ * @param parameters - parameters object
  * @param {boolean} includeFieldProj - whether to include the projected field in the formatted name e.g. age|id  OR  age
- * @returns {string} formatted dimension name
+ * @returns formatted dimension name
  */
 function formatDimensionFieldName(field: string, parameters: Parameters, includeFieldProj = true): string {
   const [fieldName, displayField = 'id'] = field.split('.'); // default dimension field projection to "id"
@@ -77,8 +77,8 @@ function formatDimensionFieldName(field: string, parameters: Parameters, include
 
 /**
  * Serializes a list of filters to a fili filter string
- * @param {Array<Filter>} filters - list of filters to be ANDed together for fili
- * @returns {string} serialized filter string
+ * @param filters - list of filters to be ANDed together for fili
+ * @returns serialized filter string
  */
 export function serializeFilters(filters: Filter[]): string {
   return filters
@@ -111,8 +111,8 @@ export default class BardFactsAdapter extends EmberObject {
    *
    * @method _buildDimensionsPath
    * @private
-   * @param {Object} request
-   * @return {String} dimensions path
+   * @param request
+   * @return dimensions path
    */
   _buildDimensionsPath(request: RequestV2 /*options*/): string {
     const dimensions = array(request.columns)
@@ -131,8 +131,8 @@ export default class BardFactsAdapter extends EmberObject {
    *
    * @method _buildDateTimeParam
    * @private
-   * @param {Object} request
-   * @return {String} dateTime param value
+   * @param request
+   * @return dateTime param value
    */
   _buildDateTimeParam(request: RequestV2): string {
     const dateTimeFilters = request.filters.filter(fil => fil.field === 'dateTime');
@@ -145,8 +145,8 @@ export default class BardFactsAdapter extends EmberObject {
    *
    * @method _buildMetricsParam
    * @private
-   * @param {Object} request
-   * @return {String} metrics param value
+   * @param request
+   * @return metrics param value
    */
   _buildMetricsParam(request: RequestV2): string {
     const metrics = request.columns.filter(col => col.type === 'metric');
@@ -164,10 +164,10 @@ export default class BardFactsAdapter extends EmberObject {
    *
    * @method _buildFiltersParam
    * @private
-   * @param {Object} request
-   * @return {String} filters param value
+   * @param request
+   * @return filters param value
    */
-  _buildFiltersParam(request: RequestV2) {
+  _buildFiltersParam(request: RequestV2): string | undefined {
     const filters = request.filters.filter((fil: Filter) => fil.type === 'dimension');
 
     if (filters?.length) {
@@ -181,11 +181,11 @@ export default class BardFactsAdapter extends EmberObject {
    *
    * @method _buildSortParam
    * @private
-   * @param {Object} request
-   * @param {function} aliasFunction function that returns metrics from aliases
-   * @return {String} sort param value
+   * @param request
+   * @param aliasFunction function that returns metrics from aliases
+   * @return sort param value
    */
-  _buildSortParam(request: RequestV2, aliasFunction: AliasFn = a => a) {
+  _buildSortParam(request: RequestV2, aliasFunction: AliasFn = a => a): string | undefined {
     const { sort } = request;
 
     if (sort.length) {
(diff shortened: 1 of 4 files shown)


packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+   * Builds a metrics param string for a request
+   *
+   * @method _buildMetricsParam
+   * @private
+   * @param {Object} request
+   * @return {String} metrics param value
+   */
+  _buildMetricsParam(request: Request): string {
+    const metrics = request.columns.filter(col => col.type === 'metric');
+    const metricIds = (metrics || []).map(metric =>
```
kevinhinterlong, line 151: Shouldn't metrics always be defined if columns is defined?
```suggestion
    const metricIds = metrics.map(metric =>
```

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+   * @private
+   * @param {Object} request
+   * @param {Object} options - optional host options
+   * @return {String} URL Path
+   */
+  _buildURLPath(request: Request, options: RequestOptions = {}) {
+    const host = configHost(options);
+    const { namespace } = this;
+    const table = request.table;
+    const timeGrain = request.columns.find(col => col.field === 'dateTime')?.parameters.grain || 'all';
```
kevinhinterlong, line 244:
```suggestion
    const timeGrain = request.columns.find(col => col.field === 'dateTime')?.parameters?.grain || 'all';
```

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+  operator: string;
+  values: string[];
+};
+type Sort = {
+  field: string;
+  parameters: Parameters;
+  direction: SortDirection;
+};
+type Request = {
+  filters: Filter[];
```
kevinhinterlong, line 51: Are we expecting that any of these might be null/undefined? I see some cases below where we add `?.` for runtime safety so we should probably clean that up either by

- at least have the functions take in `Partial<Request>` to force us to have build and run time safety for partial request objects which is effectively what we're doing now, but manually
- maybe assume that we strictly follow our types and possibly adding some asserts to be safer, but less flexible
- or by only passing `Column[]|undefined`/`Filter[]|undefined`/etc to clarify what the function actually needs

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+  /**
+   * Builds a having param string for a request
+   *
+   * @method _buildHavingParam
+   * @private
+   * @param {Object} request
+   * @param {function} aliasFunction function that returns metrics from aliases
+   * @return {String} having param value
+   */
+  _buildHavingParam(request: Request, aliasFunction = (a: string) => a) {
```
kevinhinterlong, line 216: type AliasFn = (column: string) => string

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+  }
+  return `${fieldName}${parameterString}${includeFieldProj ? '|' + displayField : ''}`;
+}
+
+/**
+ * Serializes a list of filters to a fili filter string
+ * @param {Array<Filter>} filters - list of filters to be ANDed together for fili
+ * @param {Filter} filter - filters object
+ */
+export function serializeFilters(filters: Filter[]) {
```
kevinhinterlong, line 81:
```suggestion
export function serializeFilters(filters: Filter[]): string {
```
also jsdoc

packages/data/tests/unit/adapters/bard-facts-v2-test.js
```diff
@@ -0,0 +1,965 @@
+import { module, test } from 'qunit';
```
kevinhinterlong, line 1: It would be nice if the test was also typescript, but might be a lot of work to go back and redo it

packages/data/addon/adapters/bard-facts-v2.ts
mattgraham1995: I think we shouldn't allow sort or filter to be undefined/null on the request. They should always at least have an empty array. The `?.` checks are probably just leftover from the v1 adapter, but aren't necessary anymore.

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+  page?: number;
+  perPage?: number;
+  format?: string;
+  cache?: boolean;
+  queryParams?: Dict<string>;
+};
+type Query = RequestOptions & Dict<string | number | boolean>;
+type SortDirection = typeof SORT_DIRECTIONS[number];
+type ColumnType = 'metric' | 'dimension' | 'timeDimension';
+type Parameters = Dict<string>;
```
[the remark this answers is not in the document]
mattgraham1995: Since parameters end up in the api request string, it makes since to me that their values should be strings. I'm not sure if there's a case where we specifically need a parameter value to be a number.

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,406 @@
...
+   * @param {Object} request
+   * @return {String} dimensions path
+   */
+  _buildDimensionsPath(request: Request /*options*/): string {
+    const dimensions = array(request.columns)
+      .filterBy('type', 'dimension')
+      .map(dim => formatDimensionFieldName(dim.field, dim.parameters, false));
+
+    return dimensions.length
+      ? `/${array(dimensions)
```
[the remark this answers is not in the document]
mattgraham1995: The map call at the end of the dimensions declaration returns a native array

packages/data/addon/adapters/bard-facts-v2.ts
kevinhinterlong: Ok that's what I was hoping 🙏

cythrawll:
Can you merge this into request-v2 branch?

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,408 @@
...
+   * @property {Service} ajax
+   */
+  @service ajax: TODO;
+
+  /**
+   * Builds the dimensions path for a request
+   *
+   * @method _buildDimensionsPath
+   * @private
+   * @param {Object} request
```
cythrawll, line 114: Don't include type annotation in jsdoc if it's already annotated in typescript.

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,408 @@
...
+   * Builds a having param string for a request
+   *
+   * @method _buildHavingParam
+   * @private
+   * @param {Object} request
+   * @param {function} aliasFunction function that returns metrics from aliases
+   * @return {String} having param value
+   */
+  _buildHavingParam(request: RequestV2, aliasFunction: AliasFn = a => a) {
+    const having = request.filters.filter(fil => fil.type === 'metric');
```
cythrawll, line 219: Filters don't have type field. have to use a different method for distinction. Probably can use columnMetaType

packages/data/addon/adapters/bard-facts-v2.ts
```diff
@@ -0,0 +1,408 @@
...
+type Column = {
+  field: string;
+  parameters: Parameters;
+  type: ColumnType;
+  alias: string;
+};
+type Filter = {
+  field: string;
+  parameters: Parameters;
+  type: ColumnType;
```
cythrawll, line 41: no `type` in this model

packages/data/addon/adapters/bard-facts-v2.ts
mattgraham1995: Filters are going to have a type. I discussed it with Jon and opened a PR against the request v2 RFC. We should probably add type to Sort as well.

mattgraham1995:
@cythrawll I figured it was fine for this to go into master since it's all new files

packages/data/addon/adapters/bard-facts-v2.ts
cythrawll: Yeah we should add it to sorts, (we should add it to the base fragment )

PR #792, opened 2020-05-01
Fix: Typescript some data utils to fix npm publish

mattgraham1995:
Resolves #

## Description

Predeploy step broke on `ember ts:precompile` because of imports from modules with no types defined. Could've fixed it by using `// @ts-ignore` but these will need to be typescript at some point anyway.

## Proposed Changes

- Typescript metric and adapter utils in navi-data

## License

### DIFF
--- a/packages/data/addon/utils/metric.ts
+++ b/packages/data/addon/utils/metric.ts
@@ -1,19 +1,27 @@
 /**
- * Copyright 2018, Yahoo Holdings Inc.
+ * Copyright 2020, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 import { isEmpty } from '@ember/utils';
 import { get } from '@ember/object';
 
+type Parameters = Dict<string>;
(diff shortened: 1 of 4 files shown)


packages/data/addon/utils/metric.ts
```diff
@@ -1,19 +1,27 @@
...
  */
 import { isEmpty } from '@ember/utils';
 import { get } from '@ember/object';
 
+type Parameters = Dict<string>;
+interface ColumnAttributes {
+  name: string;
+  parameters: Parameters;
+}
+interface MetricObject {
```
kevinhinterlong, line 13: We'll have to come back later to extract this out to something like this commit from Jon's earlier work

https://github.com/yahoo/navi/blob/a7b5bf862cd79dfd71c1f7cf7bea9564c7a9b26f/packages/data/types/global.d.ts#L17-L65

mattgraham1995: Yeah we definitely need some global types for the most common shapes

PR #797, opened 2020-05-05
Update: Column config - scroll to and highlight most recently added column

daniellevinson:
## Description

When adding or cloning a column, open and highlight the item and scroll to it.

## Proposed Changes

- Add a controller property `lastAddedColumn` that will save last added column type and id
- Reset it when removing a column, reverting unsaved changes or navigating away
- the `navi-column-config` component will pass a `isLastAdded` argument to each item
- if `true`, the item will be scrolled to, have a highlighted background and config will be open
- had to switch back to a regular `each` loop since `animated-each` causes scrolling to flicker. Instead, added `animated-if` for fade and collapse effect when removing a column
- disable removing a date time column when no `all` time grain in metadata

## Screenshots

May-05-2020 16-32-33

## License

### DIFF
--- a/packages/reports/addon/components/navi-column-config.js
+++ b/packages/reports/addon/components/navi-column-config.js
@@ -63,6 +63,7 @@ class NaviColumnConfig extends Component {
         name,
         displayName: this.getDisplayName(dimension, 'dimension', visualization),
         isFiltered: filteredDimensions.includes(name),
+        isRemovable: true,
         fragment: dimension
       };
     });
@@ -74,6 +75,7 @@ class NaviColumnConfig extends Component {
         name,
         displayName: this.getDisplayName(metric, 'metric', visualization),
         isFiltered: filteredMetrics.includes(name),
+        isRemovable: true,
         fragment: metric
       };
     });
@@ -86,6 +88,7 @@ class NaviColumnConfig extends Component {
         name: 'dateTime',
         displayName: this.getDisplayName(timeGrainObject, 'timeDimension', visualization),
         isFiltered: true,
+        isRemovable: timeGrains.find(grain => grain.id === 'all') ? true : false,
         fragment: 'dateTime',
         timeGrain,
         timeGrains: timeGrains.filter(grain => grain.id !== 'all')
--- a/packages/reports/addon/components/navi-column-config.js
+++ b/packages/reports/addon/components/navi-column-config.js
@@ -21,7 +21,7 @@
  */
 import Component from '@ember/component';
 import layout from '../templates/components/navi-column-config';
-import { action, computed } from '@ember/object';
+import { action, computed, set } from '@ember/object';
 import { capitalize } from '@ember/string';
 import move from 'ember-animated/motions/move';
 import { easeOut, easeIn } from 'ember-animated/easings/cosine';
@@ -210,11 +210,11 @@ class NaviColumnConfig extends Component {
    * @param {Object} column - contains type and name of column to remove from request
    */
   @action
-  removeColumn(column, { componentElement }) {
+  removeColumn(column) {
     const { type, fragment } = column;
     const removalHandler = this[`onRemove${capitalize(type)}`];
     if (removalHandler) {
-      componentElement.classList.add('navi-column-config-item--removing');
+      set(column, 'isRemoving', true);
       later(() => removalHandler(fragment), 200);
     }
   }
--- a/packages/reports/addon/components/navi-column-config.js
+++ b/packages/reports/addon/components/navi-column-config.js
@@ -25,31 +25,13 @@ import { action, computed } from '@ember/object';
 import { capitalize } from '@ember/string';
 import move from 'ember-animated/motions/move';
 import { easeOut, easeIn } from 'ember-animated/easings/cosine';
-import { fadeIn, fadeOut } from 'ember-animated/motions/opacity';
 import { layout as templateLayout, tagName } from '@ember-decorators/component';
 import { inject as service } from '@ember/service';
-import { assert } from '@ember/debug';
+import { later } from '@ember/runloop';
 
 @tagName('')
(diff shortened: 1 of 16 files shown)


packages/reports/addon/components/navi-column-config.js
```diff
@@ -113,6 +98,26 @@ class NaviColumnConfig extends Component {
...
 
+  /**
+   * @property {Object} lastAddedItem - the column that has been added last
+   */
+  @computed('columns.[]', 'lastAddedColumn')
+  get lastAddedItem() {
+    const { columns, lastAddedColumn } = this;
+
+    if (lastAddedColumn) {
+      for (let i = columns.length - 1; i >= 0; i--) {
```
cythrawll, line 109: Can we change this to a return columns.find(...)

packages/reports/addon/components/navi-column-config.js
```diff
@@ -214,7 +219,10 @@ class NaviColumnConfig extends Component {
   removeColumn(column) {
     const { type, fragment } = column;
     const removalHandler = this[`onRemove${capitalize(type)}`];
-    removalHandler?.(fragment);
+    if (removalHandler) {
+      set(column, 'isRemoving', true);
+      later(() => removalHandler(fragment), 200);
```
kevinhinterlong, line 224: I think we might want to just directly call this. Just playing around with it makes removing columns feel slow

packages/reports/addon/components/navi-column-config/item.js
```diff
@@ -69,12 +69,20 @@ class NaviColumnConfigItemComponent extends Component {
...
    * @param element - element inserted
    */
   @action
   setupElement(element) {
     this.componentElement = element;
+
+    if (this.isLastAdded) {
+      set(this, 'isColumnConfigOpen', true);
+      next(() => {
+        element.parentElement.scrollTop = element.offsetTop - element.parentElement.offsetTop;
```
kevinhinterlong, line 83: Found [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) which seems a little cleaner, but either way is fine

```suggestion
        this.componentElement.scrollIntoView(true);
```

packages/reports/addon/controllers/reports/report.js
```diff
@@ -110,6 +119,29 @@ export default class ReportsReportController extends Controller {
...
+  /**
+   * Opens the column config drawer and updates the last added column
+   * @param {ReportBuilderComponent} reportBuilder - The report builder component
+   * @param {String} columnType - the added column type
+   * @param {Object} fragment - the added request fragment
+   */
+  @action
+  onBeforeAddItem(reportBuilder, columnType, fragment) {
+    this.send('updateColumnDrawerOpen', true, reportBuilder);
+    this.send('updateLastAddedColumn', columnType, fragment);
```
kevinhinterlong, line 142: Maybe call these directly?

packages/reports/addon/controllers/reports/report.js
```diff
@@ -110,6 +119,29 @@ export default class ReportsReportController extends Controller {
...
 
+  /**
+   * Updates the last added column
+   * @param {String} type - the added column type
+   * @param {Object} fragment - the added request fragment
+   * @param {String} fragment.id - fragment id
+   */
+  @action
+  updateLastAddedColumn(type, { id }) {
+    set(this, 'lastAddedColumn', { type, name: type === 'timeDimension' ? 'dateTime' : id });
```
kevinhinterlong, line 130: Any reason we don't just use the fragment here? Seems like an easier way to search because now we rely on searching backwards through the columns and having the newest one appended there. So if we wanted clone to add right after in the list it might not work

kevinhinterlong:
Another related thing for both this and #799 is that we might want to add some extra padding to the bottom of the list. There's already a little bit for metric/dimension selector, but I think adding a lot (maybe `200px`) would be really useful here.

Also we might want to consider checking first that we only open columns that can actually be parameterized (`column.type !== 'dimension'`)

Doesn't have to be in this PR, but they should be relatively small changes

packages/reports/addon/components/navi-column-config/item.js
daniellevinson: yeah much better, I didn't like the component scrolling its parent

packages/reports/addon/components/navi-column-config.js
daniellevinson: if we do `.slice().reverse()` first or something

packages/reports/addon/components/navi-column-config.js
```diff
@@ -104,6 +88,7 @@ class NaviColumnConfig extends Component {
         name: 'dateTime',
         displayName: this.getDisplayName(timeGrainObject, 'timeDimension', visualization),
         isFiltered: true,
+        isRemovable: timeGrains.find(grain => grain.id === 'all') ? true : false,
```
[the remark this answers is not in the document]
daniellevinson: we have a couple of tables without the `all` timegrain in our apps

packages/reports/addon/controllers/reports/report.js
daniellevinson: The fragment is not yet created when we call `updateLastAddedColumn`.
It's a little on the awkward side because there's no guarantee it will be added successfully.. but if I queue `updateLastAddedColumn` _after_ adding, the `did-render` modifier (`setupElement`) is already called.

daniellevinson: If we wanted to clone right after we would need to change it to be index based or something similar

packages/reports/addon/components/navi-column-config/item.js
daniellevinson: so I just noticed the scroll doesn't work if the drawer is closed before adding :/

PR #803, opened 2020-05-12
Update: Run ts:precompile during test build

mattgraham1995:
## Description
Problems with a previous PR that were only caught in the npm publish step.

## Proposed Changes

- Fix typescript problems
- Add ts:precompile check to build so we can catch those problems before merge

## License


packages/data/tests/dummy/mirage/scenarios/graphql.ts
```diff
@@ -1,4 +1,6 @@
-export default function(server) {
+// TODO: Replace any with type supplied by new version of mirage
```
kevinhinterlong, line 1: This will be really nice

PR #805, opened 2020-05-13
Update: Use autocomplete="new-password"

kevinhinterlong:
## Description

Having some issues with Chrome trying to be helpful and trying to autocomplete on our multiselect

## Proposed Changes

- [autocomplete="new-password"](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#Preventing_autofilling_with_autocompletenew-password) to really suggest that the browser does not try to autocomplete

## License

### DIFF
--- a/packages/core/addon/templates/components/power-select-search.hbs
+++ b/packages/core/addon/templates/components/power-select-search.hbs
@@ -6,7 +6,7 @@
     <NaviIcon @icon="search" />
     <input 
       type="search" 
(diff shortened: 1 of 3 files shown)


kevinhinterlong:
> This won't cause chrome to suggest a password for our input fields will it?

I'm not seeing any weird behavior locally with new-password on chrome/firefox/safari

PR #806, opened 2020-05-13
Update: Open Date Time on new report

kevinhinterlong:
## Description
With the new experience, we should show users how they can configure metrics/time grain by automatically opening the date time

## Proposed Changes
- Show the date time as being added when creating a new report

## Screenshots
new-report

## License

### DIFF
--- a/packages/reports/addon/components/navi-column-config.js
+++ b/packages/reports/addon/components/navi-column-config.js
@@ -114,7 +114,7 @@ class NaviColumnConfig extends Component {
         });
     }
 
-    return columns.length === 1 ? columns[0] : null;
+    return columns.length === 1 && column.type === 'timeDimension' ? columns[0] : null;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/reports/addon/components/navi-column-config.js
```diff
@@ -115,7 +114,7 @@ class NaviColumnConfig extends Component {
         });
     }
 
-    return null;
+    return columns.length === 1 ? columns[0] : null;
```
daniellevinson, line 117:
```suggestion
    return columns.length === 1 && column.type === 'timeDimension' ? columns[0] : null;
```
just to cover the unlikely case where you enter a report with no time grain and one column

packages/reports/addon/components/navi-column-config.js
```diff
@@ -115,7 +114,7 @@ class NaviColumnConfig extends Component {
         });
     }
 
-    return null;
+    return columns.length === 1 && column.type === 'timeDimension' ? columns[0] : null;
```
daniellevinson, line 117: Oops that should've been 
```suggestion
    return columns.length === 1 && columns[0].type === 'timeDimension' ? columns[0] : null;
```

PR #812, opened 2020-05-15
Update: Throttle and highlight clicks on metric/dimension selections

mattgraham1995:
## Description
Users find it hard to tell when they've added a metric or dimension in the new column config, so they end up adding the same metric/dimension multiple times.

## Proposed Changes

- Highlight a clicked metric/dimension so that users have some feedback that something has happened
- Throttle clicks on metrics/dimensions to prevent unintentional additions/deletions

## Screenshots

It doesn't show in the gif, but selected metrics/dimensions still have a light `@denali-brand-100` background color.
May-15-2020 02-26-49

## License

### DIFF
--- a/packages/reports/addon/components/dimension-selector.js
+++ b/packages/reports/addon/components/dimension-selector.js
@@ -14,6 +14,7 @@
  *   {{/dimension-selector}}
  */
 
+import { throttle } from '@ember/runloop';
 import { readOnly, mapBy } from '@ember/object/computed';
 import Component from '@ember/component';
 import { get, computed, action } from '@ember/object';
@@ -21,6 +22,8 @@ import { A as arr } from '@ember/array';
 import layout from '../templates/components/dimension-selector';
 import { featureFlag } from 'navi-core/helpers/feature-flag';
 import { getDefaultTimeGrain } from 'navi-reports/utils/request-table';
+
+export const THROTTLE_TIME = 750; // milliseconds
 
 export default class DimensionSelector extends Component {
   layout = layout;
@@ -168,12 +171,13 @@ export default class DimensionSelector extends Component {
       }, {});
   }
 
-  /*
-   * @action itemClicked
-   * @param {Object} item
+  /**
+   * Pass clicked dimenion to action handler
+   * @param {Object} item - grouped list item for clicked dimension
+   * @param {Node} target - DOM Node for clicked dimension
    */
-  @action
-  itemClicked(item) {
+  doItemClicked(item, target) {
+    target.focus(); // firefox does not focus a button on click in MacOS specifically
     const type = item.dateTimeDimension ? 'TimeGrain' : 'Dimension';
     const enableRequestPreview = featureFlag('enableRequestPreview');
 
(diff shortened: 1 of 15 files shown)


packages/reports/addon/templates/components/dimension-selector.hbs
```diff
@@ -18,7 +18,7 @@
     {{#if item.dateTimeDimension}}
       {{#if (feature-flag "enableRequestPreview")}}
         <div class="grouped-list__item-container {{if this.selectedTimeGrain "grouped-list__item-container--selected grouped-list__item-container--disabled"}}">
-          <span class="grouped-list__item-label"
+          <button class="grouped-list__item-label"
```
kevinhinterlong, line 21: type="button"

Maybe double check the focus in firefox issue with this?

mattgraham1995: Doesn't fix the firefox issue it seems, but I've added it.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Clicking_and_focus

The animation stopped working because I decided, being the genius that I am, to shorten the `blur` callback I passed to the event listener. That broke it, but it should be working now if you want to try again.

kevinhinterlong: Missed type=button for this instance

packages/reports/app/styles/navi-reports/components/grouped-list.less
```diff
@@ -39,6 +39,10 @@
     font-size: @font-size-mid;
 
     &-label {
+      font: inherit;
```
daniellevinson, line 42: 🔤

packages/reports/addon/templates/components/dimension-selector.hbs
```diff
@@ -18,16 +18,17 @@
...
             role="button"
```
daniellevinson, line 22: do we need role button if it's a button?

daniellevinson:
probably should change this in metric config as well.. lost the `cursor: pointer` there and no animation

packages/reports/addon/components/dimension-selector.js
```diff
@@ -22,6 +23,37 @@ import layout from '../templates/components/dimension-selector';
...
+export function BlurOnAnimationEnd(target, button) {
+  const listItemContainer = target.closest('.grouped-list__item-container--selected');
+  if (listItemContainer) {
+    // Detect the end of the css animation depending on browser and blur the button
+    let animationEndEvent;
+    for (let animation in ANIMATIONS) {
+      if (listItemContainer.style[animation] != undefined) {
+        animationEndEvent = ANIMATIONS[animation];
+      }
+    }
```
daniellevinson, line 49: will something like `listItemContainer.classList.add('some-animation-class-like-navi-column-config-item--last-added')` work?
then we might be able to get rid of the prefixes and the focus that creates this dotted border on firefox
Screen Shot 2020-05-18 at 11 50 54 AM
if we need to restart the animation for the same item we can maybe do https://css-tricks.com/restart-css-animation/#article-header-id-0

packages/reports/addon/components/metric-selector.js
```diff
@@ -13,6 +13,7 @@
  *   {{/metric-selector}}
  */
 
+import { throttle } from '@ember/runloop';
```
daniellevinson, line 16: 2020

packages/reports/addon/components/dimension-selector.js
mattgraham1995: This was my first approach, but because these elements exist inside vertical collection, relying on the addition and removal of these elements to play the animations was causing animation replays on scroll. I can try to rework it by attaching a separate animation class, but I feel like it might just be easier to do this to fix the outline:
https://stackoverflow.com/questions/71074/how-to-remove-firefoxs-dotted-outline-on-buttons-as-well-as-links

packages/reports/addon/templates/components/dimension-selector.hbs
mattgraham1995: From what I can tell, it doesn't hurt anything and there may be instances where accessibility software benefits from an explicit role rather than relying on the default.

mattgraham1995:
@daniellevinson Metric config updated and the datetime issue have been fixed. We don't need to throttle without request preview enabled.

packages/reports/addon/components/metric-selector.js
```diff
@@ -114,6 +117,23 @@ class MetricSelectorComponent extends Component {
...
   }
+
+  /**
+   * @action
+   * @param {Object} metric - grouped list item for clicked metric
+   * @param {Event.target} target - clicked metric element
+   */
+  @action
+  metricClicked(metric, { target }) {
+    const button = target.closest('button.grouped-list__item-label');
```
daniellevinson, line 128: move it inside the if?

PR #813, opened 2020-05-18
Update: Column config accordion

daniellevinson:
## Description

Column config accordion

## Proposed Changes

- Only one open column at a time
- Enable clicking Date Time in the dimension selector even if it's already added. The action will open and highlight the column, if it's not already open.

## Screenshots

May-17-2020 20-42-26

## License

### DIFF
--- a/packages/reports/addon/components/navi-column-config.js
+++ b/packages/reports/addon/components/navi-column-config.js
@@ -123,9 +123,9 @@ class NaviColumnConfig extends Component {
   @service metricName;
 
   /**
-   * @property {Object} currentlyOpenFragment - fragment of the column that is currently open
+   * @property {Object} currentlyOpenColumn - the column that is currently open
    */
-  currentlyOpenFragment = null;
+  currentlyOpenColumn = null;
 
   /**
    * @method getDisplayName
@@ -232,7 +232,7 @@ class NaviColumnConfig extends Component {
    */
   @action
   openColumn(column) {
-    this.set('currentlyOpenFragment', column?.fragment);
+    this.set('currentlyOpenColumn', column);
   }
 
   /**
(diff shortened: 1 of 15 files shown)


packages/reports/addon/templates/components/navi-column-config.hbs
```diff
@@ -8,6 +10,8 @@
           <NaviColumnConfig::Item
             @column={{column}}
             @isLastAdded={{eq column this.lastAddedItem}}
+            @isOpen={{eq column.fragment this.currentlyOpenFragment}}
```
jkusa, line 13: What is the thinking behind checking/storing `column.fragment` vs just the `column` obj?

daniellevinson: The fragments determine if we re-render the columns, for example changing the time grain would create a new column.
Since we're comparing only the fragment, no real reason to store the entire column.

jkusa: In general, I think it is nicer to use higher levels of abstraction if there is no benefit to using lower level concepts (like performance)? If we can avoid leaking the`fragment` concept, I think we should.

packages/reports/addon/components/navi-column-config.js
```diff
@@ -221,12 +226,35 @@ class NaviColumnConfig extends Component {
...
+  }
+
+  /**
+   * Opens the date time column when it's the only column
+   * @action
+   */
+  @action
+  openDefaultColumn() {
+    const { columns, openColumn } = this;
+    if (columns.length === 1 && columns[0].type === 'timeDimension') {
```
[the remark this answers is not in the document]
daniellevinson: I think we have this condition in other places as well... can we save this refactor for when we have this new column? It will probably introduce other changes in the code

PR #818, opened 2020-05-22
Add Role model and relationship to user to navi-webservice

ldkge:
Resolves #

## Description
Add Role model and relationship to user to navi-webservice. 

Permission settings are coming in a different PR.

## Proposed Changes

- Add Role model and tests
- Add roles relationship to User model and update tests

## License


packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/Role.kt
```diff
@@ -0,0 +1,47 @@
...
+    @get:NotBlank
+    var id: String? = null
+
+    @get:CreationTimestamp
+    @get:Column(columnDefinition = "timestamp default current_timestamp")
+    @get:Temporal(TemporalType.TIMESTAMP)
+    @get:UpdatePermission(expression = "nobody")
+    var createdOn: Date? = null
+
+    @get:CreationTimestamp
```
mattgraham1995, line 42: Should this be `UpdateTimestamp`?
https://docs.jboss.org/hibernate/orm/4.3/javadocs/org/hibernate/annotations/UpdateTimestamp.html

packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/User.kt
```diff
@@ -44,6 +44,12 @@ class User {
     @get:UpdatePermission(expression = "nobody")
     var createdOn: Date? = null
 
+    @get:CreationTimestamp
```
mattgraham1995, line 47: `UpdateTimestamp`?

packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/User.kt
```diff
@@ -75,4 +81,7 @@ class User {
             inverseJoinColumns = arrayOf(JoinColumn(name = "dashboard_id", referencedColumnName = "id"))
     )
     var favoriteDashboards: Collection<Dashboard> = arrayListOf()
-}
\ No newline at end of file
+
+    @get:OneToMany(targetEntity = Role::class)
```
mattgraham1995, line 85: Is there an annotation for just a `toMany` relationship since the role doesn't actually have a relationship to the users?

mattgraham1995: Looking into it, this is functionally a `toMany` annotation since `roles` doesn't have a `ManyToOne` annotation on user so you can ignore this comment.

ldkge: There's one but I haven't found a code example and I get errors like `Could not determine type for: java.util.Collection, at table: navi_users, for columns: [org.hibernate.mapping.Column(roles)]`

PR #819, opened 2020-05-22
Update: Add tootlip support for metric parameters.

cythrawll:
- also allow param arguments with name over description.

## Description

Allows description to be a property of metric function parameters that can be used as tooltip helpers.

Allows the shape of metric function parameter arguments to be `{id, name}` as well as support `{id, description}`

## Proposed Changes

- Add description property to metric function parameters.
- If said description exists show tooltip help next to parameter in metric config
- Add support to use `name` rather than `description` if the values have a `name` param. 

FUTURE TODO: Normalize all function arguments to `{name, id, description}` shape. even if arguments come from external sources like a dimension.

## Screenshots

Screen Shot 2020-05-22 at 4 15 25 PM

## License

### DIFF
--- a/packages/reports/addon/components/navi-column-config/metric.js
+++ b/packages/reports/addon/components/navi-column-config/metric.js
@@ -44,7 +44,7 @@ class NaviColumnConfigMetricComponent extends Component {
     if (!metric.hasParameters) {
       return [];
     }
-    return metric.arguments.map(arg => arg.id);
+    return metric.arguments.map(arg => arg);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/reports/addon/components/navi-column-config/metric.js
```diff
@@ -44,7 +44,7 @@ class NaviColumnConfigMetricComponent extends Component {
     if (!metric.hasParameters) {
       return [];
     }
-    return metric.arguments.map(arg => arg.id);
+    return metric.arguments.map(arg => arg);
```
mattgraham1995, line 47:
```suggestion
    return metric.arguments;
```

PR #822, opened 2020-05-26
Update: SupportedGrains serialize as objects

mattgraham1995:
## Description
Supported Grains need to be serialized as objects into the time dimension model

## Proposed Changes

- Normalize support grains connection into objects

## License


packages/data/addon/serializers/elide-metadata.ts
```diff
@@ -191,7 +196,9 @@ export default class ElideMetadataSerializer extends EmberObject {
         tableId,
         source,
         tags: node.columnTags,
-        supportedGrains: node.supportedGrains,
+        supportedGrains: node.supportedGrains.edges.map((edge: Edge<TimeDimensionGrainNode>) => {
```
kevinhinterlong, line 199: Does `Edge<TimeDimensionGrainNode>` need to be manually typed?

PR #823, opened 2020-05-27
Update: Use get-shallow when using column id as key

mattgraham1995:
## Description
Elide is going to use a convention of `` `${table.id}.${columnId}` `` for its column ids. This causes problems when we try to use the basic `get` helper that is built-in to Ember since it sees that `.` in the middle of the id as a nested access indicator. 

## Proposed Changes

- Add a `get-shallow` helper that will allow us to use column ids as keys

## License

### DIFF
--- a/packages/core/addon/helpers/get-shallow.ts
+++ b/packages/core/addon/helpers/get-shallow.ts
@@ -13,9 +13,7 @@
 import { helper } from '@ember/component/helper';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type Params = [Dict<any>, string];
-export function getShallow(params: Params) {
-  const [obj, key] = params;
+export function getShallow([obj, key]: [Dict<any>, string]) {
   return obj[key];
 }
 
--- a/packages/core/addon/helpers/get-shallow.ts
+++ b/packages/core/addon/helpers/get-shallow.ts
@@ -1,0 +1,22 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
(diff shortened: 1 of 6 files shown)


packages/core/addon/helpers/get-shallow.ts
```diff
@@ -0,0 +1,22 @@
...
+ * obj = { 'foo.bar': 'baz' };
+ * {{get-shallow obj "foo.bar"}}
+ * -> 'baz'
+ * {{get obj "foo.bar"}}
+ * -> undefined
+ */
+import { helper } from '@ember/component/helper';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type params = [Dict<any>, string];
```
kevinhinterlong, line 16:
```suggestion
type Params = [Dict<any>, string];
```

daniellevinson:
do we need to update metric-config.hbs too?

packages/core/addon/helpers/get-shallow.ts
```diff
@@ -13,8 +13,8 @@
 import { helper } from '@ember/component/helper';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type params = [Dict<any>, string];
-export function getShallow(params: params) {
+type Params = [Dict<any>, string];
+export function getShallow(params: Params) {
```
kevinhinterlong, line 17: Actually, maybe just do
```suggestion
export function getShallow([obj, key]: [object, string]) {
```

mattgraham1995:
@daniellevinson Updated

packages/core/addon/helpers/get-shallow.ts
mattgraham1995: Going to keep `Dict<any>` since typescript seems to have problems with using `object` there.
Screen Shot 2020-05-27 at 1 21 41 PM

PR #824, opened 2020-05-27
Fix: Enable Request Preview disables metric-config

cythrawll:
## Description

Disable metric config gear menu when request preview is on: It obstructs column preview and doesn't give enough functionality

## Proposed Changes

- Removes gear icon when the enableRequestPreview flag is on

## License

### DIFF
--- a/packages/reports/addon/templates/components/metric-selector.hbs
+++ b/packages/reports/addon/templates/components/metric-selector.hbs
@@ -38,7 +38,7 @@
         </NaviIcon>
 
         <div class="grouped-list__icon-set {{if (not (can-having metric)) "grouped-list__icon-set--no-filter"}}">
-          {{#if (and (get metric "hasParameters") (not (feature-flag "enableRequestPreview")))}}
+          {{#if (and metric.hasParameters (not (feature-flag "enableRequestPreview")))}}
             <MetricConfig
               @metric={{metric}}
               @request={{@request}}
--- a/packages/reports/addon/templates/components/metric-selector.hbs
+++ b/packages/reports/addon/templates/components/metric-selector.hbs
@@ -38,7 +38,7 @@
         </NaviIcon>
 
         <div class="grouped-list__icon-set {{if (not (can-having metric)) "grouped-list__icon-set--no-filter"}}">
-          {{#if (get metric "hasParameters")}}
+          {{#if (and (get metric "hasParameters") (not (feature-flag "enableRequestPreview")))}}
             <MetricConfig
               @metric={{metric}}
               @request={{@request}}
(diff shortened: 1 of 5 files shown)


packages/reports/tests/acceptance/column-config-test.js
```diff
@@ -1364,6 +1348,9 @@ module('Acceptance | Navi Report | Column Config', function(hooks) {
       ['Date Time (Day)', 'Platform Revenue (CAD)'],
       'Clicking the filtered option changes the metrics parameter'
     );
+
+    await click(findAll('.grouped-list__group-header').filter(el => el.textContent.includes('Revenue'))[0]);
+    assert.dom('.metric-config').doesNotExist();
```
mattgraham1995, line 1353: Can we add an assert message?
```suggestion
    assert.dom('.metric-config').doesNotExist('Metric config is not visible when request preview is enabled');
```

packages/reports/addon/templates/components/metric-selector.hbs
```diff
@@ -38,7 +38,7 @@
         </NaviIcon>
 
         <div class="grouped-list__icon-set {{if (not (can-having metric)) "grouped-list__icon-set--no-filter"}}">
-          {{#if (get metric "hasParameters")}}
+          {{#if (and (get metric "hasParameters") (not (feature-flag "enableRequestPreview")))}}
```
kevinhinterlong, line 41: Can this be

```suggestion
          {{#if (and metric.hasParameters (not (feature-flag "enableRequestPreview")))}}
```

packages/reports/tests/acceptance/metric-parameters-test.js
```diff
@@ -90,65 +90,6 @@ module('Acceptance | navi-report - metric parameters', function(hooks) {
...
-    assert.expect(7);
-
-    const originalFeatureFlag = config.navi.FEATURES.enableRequestPreview;
-
-    config.navi.FEATURES.enableRequestPreview = true;
-
-    await visit('/reports/1/view');
-
-    //adding a metric with default params
-    await clickItem('metric', 'Platform Revenue');
```
daniellevinson, line 103: do we have another test for default params?

cythrawll: Yup it's in this test file.

daniellevinson: I mean, should we still test the default param case when the flag is turned on, even if there's no config that opens up?

daniellevinson:
There's probably some feature flag dependent logic we can remove from the metric-config component

cythrawll:
@daniellevinson okay removed metric-config logic.

PR #828, opened 2020-05-29
Update: Don't rerun request after reordering table columns

daniellevinson:
## Description

There's no need to rerun the request after reordering table columns.

## Proposed Changes

Reordering metric columns changes the order of the metrics fragment array.
Change `hasRequestRun` method to return true if the only diff is the order of metrics, otherwise false.

## Screenshots

May-29-2020 15-23-17

## License

### DIFF
--- a/packages/reports/addon/controllers/reports/report/view.js
+++ b/packages/reports/addon/controllers/reports/report/view.js
@@ -7,6 +7,12 @@ import { isEqual, omit } from 'lodash-es';
 import Controller, { inject as controller } from '@ember/controller';
 import { canonicalizeMetric } from 'navi-data/utils/metric';
 
+/**
+ * @param {Object} request
+ * @returns {Array} canonicalized metrics sorted alphabetically
+ */
+const sortedMetrics = request => (request.metrics || []).map(metric => canonicalizeMetric(metric)).sort();
+
 export default class ReportViewController extends Controller {
   /*
    * @property {Controller} reportController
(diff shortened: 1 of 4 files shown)


packages/reports/addon/controllers/reports/report/view.js
```diff
@@ -1,28 +1,48 @@
...
 
-    return isEqual(modifiedRequest, this.model.get('request'));
-  })
-});
+    if (!isEqual(sortedMetrics(request), sortedMetrics(modifiedRequest))) {
+      //changes in metrics outside of order
+      return false;
+    }
+
+    if (!isEqual(omit(modifiedRequest, 'metrics'), omit(request, 'metrics'))) {
```
kevinhinterlong, line 40: Just for consistency
```suggestion
    if (!isEqual(omit(request 'metrics'), omit(modifiedRequest,, 'metrics'))) {
```

packages/reports/tests/acceptance/report-test.js
```diff
@@ -2141,6 +2141,31 @@ module('Acceptance | Navi Report', function(hooks) {
...
 
+  test('reordering metrics does not rerun the request', async function(assert) {
+    assert.expect(1);
+    await visit('/reports/2');
+
+    server.urlPrefix = `${config.navi.dataSources[0].uri}/v1`;
+    server.get('/data/*path', () => {
+      assert.ok(false, 'Request was rerun');
+    });
+
```
kevinhinterlong, line 2152: It would be nice to assert the order before too just to verify what changed

PR #830, opened 2020-06-02
Remove trend formatting

kevinhinterlong:
## Description
"Threshold" metrics relied on metadata stating they were in a "Trend" category. Instead we'll rely on apps providing their own way of rendering metrics and indicating whether or not a metric can be aggregated.

## Proposed Changes
- Replace this behavior with a "canAggregateSubtotal" flag to show a column can be aggregated
- Remove threshold metric type

## License


packages/core/tests/unit/models/table-test.js
```diff
@@ -585,9 +592,9 @@ module('Unit | Model | Table Visualization Fragment', function(hooks) {
    * @param {Array} thresholds - array of thresholds
    * @returns {Object} request object
    */
-  function buildTestRequest(dimensions = [], metrics = [], thresholds = [], timeGrain = 'day') {
+  function buildTestRequest(dimensions = [], metrics = [], timeGrain = 'day') {
     return {
-      metrics: [...metrics, ...thresholds].map(m => {
+      metrics: [...metrics].map(m => {
```
mattgraham1995, line 597:
```suggestion
      metrics: metrics.map(m => {
```

PR #831, opened 2020-06-03
Update: Support deserializing metadata objects from ids that contain "."

mattgraham1995:
Resolves #

## Description
Reports saved with a metric or dimension with an id like `tableA.metric1` would not be successfully deserialized in the metadata transform due to the `.` character in the id.

## Proposed Changes

- Extract the namespace from before only the first `.` in the id.

## License

### DIFF
--- a/packages/core/addon/transforms/dimension.js
+++ b/packages/core/addon/transforms/dimension.js
@@ -4,6 +4,8 @@
  */
 
 import BaseMetadataTransform from './base-metadata-transform';
+import { isPresent } from '@ember/utils';
+import config from 'ember-get-config';
 
 export default BaseMetadataTransform.extend({
   /**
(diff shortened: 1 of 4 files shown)


packages/core/addon/transforms/dimension.js
```diff
@@ -24,8 +26,10 @@ export default BaseMetadataTransform.extend({
   deserialize(serialized) {
     let namespace = null;
 
-    if (serialized.includes('.')) {
-      [namespace, serialized] = serialized.split('.');
+    if (isPresent(config.navi.dataSources) && serialized.includes('.')) {
+      const splitName = serialized.split('.');
+      namespace = splitName.shift();
+      serialized = splitName.join('.');
```
jkusa, line 32: Can we add tests for these cases

mattgraham1995: Tests added

PR #851, opened 2020-06-18
Basic Elide fact mock

mattgraham1995:
## Description
Need to mock responses for new async api requests

## Proposed Changes

- Add (currently unused) mocks for submitting async request and result generation

## License


packages/data/package.json
```diff
@@ -35,8 +35,10 @@
     "ember-cli-htmlbars": "^4.2.2",
     "ember-cli-mirage": "~1.1.6",
     "ember-cli-mirage-graphql": "^0.3.1",
+    "ember-cli-moment-shim": "^3.8.0",
```
cythrawll, line 38: Do we need this now that we can just import ember?

mattgraham1995: When I didn't include the shim, it wasn't able to find the moment module. Is there some other configuration for ember-moment to include the ES6 module? I wasn't able to find anything about it

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -4,15 +4,182 @@
...
       ids(records, _, ids) {
         return Array.isArray(ids) ? records.filter(record => ids.includes(record.id)) : records;
       }
     }
+  },
+  mutations: {
+    asyncQuery(asyncQueries, { op, data }) {
+      const queryIds = data.id ? [data.id] : [];
+      const existingQueries = asyncQueries.find(queryIds) || [];
+      if (op === 'UPSERT' && existingQueries.length < 1) {
```
kevinhinterlong, line 165:
```suggestion
      if (op === 'UPSERT' && existingQueries.length === 0) {
```

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -4,15 +4,182 @@
...
+
+            // Add each metric
+            rows = rows.map(row => {
+              const metrics = columns.metric.reduce((metricsObj, metric) => {
+                metricsObj[metric] = faker.finance.amount();
+                return metricsObj;
+              }, {});
+
+              Object.keys(metrics).forEach(metric => (row[metric] = metrics[metric]));
+              return row;
```
kevinhinterlong, line 128: Could make this more direct

packages/data/package.json
cythrawll: ember-auto-import should be able to do it.

mattgraham1995: Screen Shot 2020-06-22 at 3 44 41 PM
Removing the shim results in this even with ember-auto-import

cythrawll: Did you install moment by itself?

mattgraham1995: For some reason I thought we needed ember-moment here, but we don't use the template helpers so we can just use `moment`.

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -4,15 +4,180 @@
...
+            if (columns.timeDimension.length > 0) {
+              dates = _getDates(args.filter);
+            }
+
+            // Convert each date into a row of data
+            let rows = dates.map(date => {
+              return {
+                dateTime: date
+              };
+            });
```
jkusa, line 103:
```suggestion
            let rows = dates.map(dateTime => ({ dateTime }));
```

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -4,15 +4,180 @@
...
+
+            return JSON.stringify({
+              data: {
+                [table]: {
+                  edges: rows.map(row => {
+                    return {
+                      node: row
+                    };
+                  })
```
jkusa, line 136:
```suggestion
                  edges: rows.map(node => ({ node }))
```

PR #861, opened 2020-06-22
Update: Add async fetch adapter interface

### LINKED ISSUE #855: Add async adapter interfaces to navi-data/adapters/fact-interface.ts

jkusa:
Resolves #855

## Description

Add async fetch adapter interface

## Proposed Changes

- Add async fetch adapter interface based on async RFC

## License

### DIFF
--- a/packages/data/addon/adapters/fact-interface.ts
+++ b/packages/data/addon/adapters/fact-interface.ts
@@ -3,6 +3,8 @@
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 
+import NaviFactsModel from 'navi-data/models/navi-facts';
+
 export type RequestV1 = TODO;
 
 export type RequestOptions = {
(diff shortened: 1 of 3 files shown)


packages/data/addon/adapters/fact-interface.ts
```diff
@@ -3,6 +3,8 @@
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 
+import NaviFactsModel from 'navi-data/models/navi-facts';
```
jkusa, line 6: Ran into prettier issues when trying to import just the type. We might need to upgrade to prettier 2.x

packages/data/addon/adapters/fact-interface.ts
```diff
@@ -58,7 +60,29 @@ export type RequestV2 = {
...
+  TIMEDOUT = 'timedout',
+  FAILURE = 'failure'
+}
+
+export interface AsyncQuery {
+  requestId: string;
+  principalName: string;
+  request: RequestV1 | RequestV2;
+  status: QueryStatus;
+  result: NaviFactsModel;
```
mattgraham1995, line 77: This should be 
```suggestion
  result: AsyncQueryResult;
```
based on the RFC

jkusa: Right, this is for the short term. Need to come back.

PR #892, opened 2020-07-20
GraphQL Mocks for basic Elide Fact Queries

mattgraham1995:
## Description
We need a sort of Bard-lite solution for fact queries sent through GraphQL.

## Proposed Changes

- Add handler for mocking fact responses to GraphQL queries. 

## License


packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -62,81 +62,94 @@ function _parseGQLQuery(queryStr) {
...
+      });
+
+      // Add each metric
+      rows = rows.map(row => {
+        columns.metrics.forEach(metric => {
+          row[metric] = faker.finance.amount();
+        });
+
+        return row;
+      });
```
jkusa, line 125:
```suggestion
      rows = rows.map(row => ({ ...row, [metric]: faker.finance.amount() });
```

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -62,81 +62,94 @@ function _parseGQLQuery(queryStr) {
...
+  return null;
+}
+
+const OPTIONS = {
+  fieldsMap: {
+    AsyncQuery: {
+      result(_, db, parent) {
+        return {
+          httpStatus: 200,
+          contentLength: 5,
```
jkusa, line 150: is this required here?

packages/data/addon/services/navi-facts.ts
```diff
@@ -21,7 +21,7 @@ export default class NaviFactsService extends Service {
    * @param {String} type
    * @returns {Adapter} adapter instance for type
    */
-  _adapterFor(type = 'bard-facts'): NaviFactAdapter {
+  _adapterFor(type = 'elide-facts'): NaviFactAdapter {
```
jkusa, line 24: Can we revert the changes in this file?

packages/data/addon/services/navi-facts.ts
```diff
@@ -68,9 +68,8 @@ export default class NaviFactsService extends Service {
    * @returns {Promise} - Promise with the bard response model object
    */
   fetch(this: NaviFactsService, request: RequestV1, options: RequestOptions): Promise<NaviFactsModel> {
-    const type = config.navi.dataSources[0].type;
-    const adapter = this._adapterFor(type);
-    const serializer = this._serializerFor(type);
+    const adapter = this._adapterFor();
+    const serializer = this._serializerFor();
```
jkusa, line 72: Can we update this so fili instances still work?

packages/data/tests/unit/services/navi-facts-elide-test.js
```diff
@@ -0,0 +1,68 @@
...
+      dimension: 'dimension4',
+      operator: 'in',
+      values: ['v3', 'v4']
+    }
+  ],
+  having: [
+    {
+      metric: 'metric1',
+      operator: 'gt',
+      value: 0 // TODO switch to `values: [0]` after `value` backwards compatibility has been removed
```
jkusa, line 25: Can this be `values`?

packages/data/tests/unit/services/navi-facts-elide-test.js
```diff
@@ -0,0 +1,68 @@
...
+
+  hooks.beforeEach(function() {
+    this.service = this.owner.lookup('service:navi-facts');
+    GraphQLScenario(this.server);
+  });
+
+  test('fetch', async function(assert) {
+    assert.expect(2);
+
+    const response = await this.service.fetch(TestRequest);
```
jkusa, line 48: Can we do a deep equals assert on the response directly?

packages/data/addon/mirage/handlers/graphql.js
mattgraham1995: `metric` is undefined without the `columns.metrics.forEach(metric ...`

packages/data/tests/unit/services/navi-facts-elide-test.js
mattgraham1995: No because the values in the response are generated by faker

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -62,81 +63,86 @@ function _parseGQLQuery(queryStr) {
...
+      });
+
+      // Add each metric
+      rows = rows.map(row => {
+        columns.metrics.forEach(metric => {
+          row[metric] = faker.finance.amount();
+        });
+
+        return row;
+      });
```
jkusa, line 119: ah, I meant something like this:
```suggestion
      rows = rows.map((currRow) =>
        columns.metrics.reduce((row, metric) => ({ ...row, [metric]: faker.finance.amount() }), currRow));
```

packages/data/tests/unit/services/navi-facts-elide-test.js
jkusa: Can we set a faker seed with a base64 string of the query https://faker.readthedocs.io/en/master/#seeding-the-generator

packages/data/addon/services/navi-facts.ts
```diff
@@ -68,7 +69,9 @@ export default class NaviFactsService extends Service {
    * @returns {Promise} - Promise with the bard response model object
    */
   fetch(this: NaviFactsService, request: RequestV1, options: RequestOptions): Promise<NaviFactsModel> {
-    const type = config.navi.dataSources[0].type;
+    const dataSource = options?.dataSourceName || getDefaultDataSourceName();
+    const type = config.navi.dataSources?.find((source: { name: string; type: string }) => source.name === dataSource)
```
kevinhinterlong, line 73: I feel like this could be extracted into the adapter utils

packages/data/tests/unit/services/navi-facts-elide-test.js
```diff
@@ -0,0 +1,68 @@
...
+      dimension: 'dimension4',
+      operator: 'in',
+      values: ['v3', 'v4']
+    }
+  ],
+  having: [
+    {
+      metric: 'metric1',
+      operator: 'gt',
+      values: [0] // TODO switch to `values: [0]` after `value` backwards compatibility has been removed
```
kevinhinterlong, line 25:
```suggestion
      values: [0]
```

packages/data/tests/unit/services/navi-facts-elide-test.js
mattgraham1995: Faker requires the seed to be a base 10 int, so I just seeded it off the total length of the table name, args, and fields in the request

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -11,6 +11,13 @@ import moment from 'moment';
 const API_DATE_FORMAT = 'YYYY-MM-DD';
 const ASYNC_RESPONSE_DELAY = 5000; // ms before async api response result is populated
 
+function _getSeedForRequest(table, args, fields) {
```
kevinhinterlong, line 14: Could move this function up to data/mirage, but this should be fine too

https://github.com/yahoo/navi/blob/dd761f5cd895efb5aa63bd8efe1fd868428477e4/packages/app/mirage/config.js#L17-L29

mattgraham1995: I originally looked into doing something like this, but it's a bit more than what's needed for this instance. We just need a deterministic way of seeding faker based on the request. If it needed to be a unique seed for each request, then I would've used a proper hashing method.

packages/data/addon/services/navi-facts.ts
```diff
@@ -70,8 +70,7 @@ export default class NaviFactsService extends Service {
    */
   fetch(this: NaviFactsService, request: RequestV1, options: RequestOptions): Promise<NaviFactsModel> {
     const dataSource = options?.dataSourceName || getDefaultDataSourceName();
-    const type = config.navi.dataSources?.find((source: { name: string; type: string }) => source.name === dataSource)
-      ?.type;
+    const type = getDataSourceField('type', dataSource);
```
jkusa, line 73: Do we need a special function for this? Can, we just get the default data source and then pull out the field?

packages/data/addon/utils/adapter.ts
```diff
@@ -16,43 +15,36 @@ export type SourceAdapterOptions = {
...
+
   if (name) {
-    const host = config.navi.dataSources.find((dataSource: DataSource) => dataSource.name === name);
-    if (host && host[field]) {
-      return host[field];
+    const host = dataSources.find((dataSource: DataSource) => dataSource.name === name);
+    if (host) {
+      return host;
     }
     warn(
```
jkusa, line 30: Can we assert instead?

packages/data/addon/utils/adapter.ts
```diff
@@ -16,43 +15,36 @@ export type SourceAdapterOptions = {
...
-/**
- * Gets host by name in [{name: String, uri: String}]
- * config datastructure in config.navi.dataSources.
- *
- * @param name - name of host to get
- * @returns requested field of fact datasource to
- */
-export function getHost(name: string | undefined): string {
-  return getDataSourceField('uri', name);
+  return defaultDataSource
```
jkusa, line 37: Can this be its own function ?

packages/data/addon/utils/adapter.ts
```diff
@@ -14,29 +14,35 @@ export type SourceAdapterOptions = {
...
+ * @returns default data source if one is configured otherwise the first data source
+ */
+export function getDefaultDataSource(): DataSource {
+  const {
+    navi: { defaultDataSource, dataSources }
+  } = config;
+
+  return defaultDataSource
+    ? dataSources.find((source: DataSource) => source.name === defaultDataSource)
+    : dataSources[0];
```
jkusa, line 27:
```suggestion
  return defaultDataSource ? getDataSource(defaultDataSource): dataSources[0];
```

packages/data/addon/utils/adapter.ts
```diff
@@ -14,29 +14,35 @@ export type SourceAdapterOptions = {
...
-      {
-        id: 'navi-fact-host-not-configured'
-      }
-    );
+    assert(`Datasource ${name} should be configured in the navi environment`);
   }
-  return defaultDataSource
-    ? dataSources.find((source: DataSource) => source.name === defaultDataSource)
-    : dataSources[0];
+  return getDefaultDataSource();
```
jkusa, line 45: We have the function `getDefaultDataSource` do we need this functionality here?

PR #895, opened 2020-07-22
Visual regression test pr 

holderb5:
Resolves #
https://github.com/yahoo/navi/issues/847
Resolves the above for local testing, not for travis-ci

## Description
Integrates 5 basic ember backstop tests into navi, as well as docker connection and capability to run tests reliably in dev mode.
## Proposed Changes

- Adds in capability for ember backstop testing
- Upon cloning of repository running "BACKSTOP=true ember server" will allow developers to navigate to testing and see the backstop tests run, for backstop tests development is required
- docker command: docker run -p 3000:3000 --rm -it --mount type=bind,source="`pwd`",target=/src backstopjs/backstopjs:4.5.1 remote "--moby=true"
- results of tests can be seen at /backstop/backstop_data/html_report/

## License


packages/app/backstop.json
```diff
@@ -0,0 +1,51 @@
+{
```
jkusa, line 1: Is this file used?

jkusa, line 1: Is this file used?

packages/app/ember-backstop/backstop.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-env browser, node */
```
jkusa, line 1: Is this file used?

packages/app/package.json
```diff
@@ -18,42 +18,44 @@
     "lint:js": "eslint .",
     "lint": "npm run lint:hbs && npm run lint:js",
     "start": "ember serve",
+    "pretest": " ember backstop-remote& ",
```
jkusa, line 21: Can we check if the backstop environment var? Also, can we have a script for the docker command?

packages/app/package.json
```diff
@@ -18,42 +18,44 @@
...
     "ember-cli-app-version": "^3.2.0",
-    "ember-cli-babel": "^7.17.2",
-    "ember-cli-code-coverage": "^1.0.0-beta.8",
+    "ember-cli-babel": "^7.21.0",
+    "ember-cli-code-coverage": "^1.0.0-beta.9",
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-dependency-lint": "^1.1.3",
     "ember-cli-eslint": "^5.1.0",
-    "ember-cli-flash": "^1.8.1",
+    "ember-cli-flash": "^1.9.0",
```
jkusa, line 47: Are these version changes intentional?

packages/app/backstop.json
holderb5: Double checked and its not. Removed it now. Same with the backstop js in ember-backstop.

packages/app/package.json
holderb5: reverted the ones that were not intentional.

packages/app/config/dependency-lint.js
```diff
@@ -2,6 +2,7 @@
 
 module.exports = {
   allowedVersions: {
-    'ember-composable-helpers': '3.1.1 || 2.4.0'
+    'ember-composable-helpers': '3.2.0 || 2.4.0',
+    'ember-maybe-in-element': '0.4.0 || 0.2.0'
```
jkusa, line 6: Do we need both?

holderb5: Looking at these, ember dropdown element automatically uses ember-maybe-in-element 0.2.0 as a dependency so it causes an error without the conditional to have either 0.4.0 or 0.2.0

packages/app/ember-backstop/backstop.js
holderb5: As discussed offline, we don't reference it but builds fail without it

jkusa:
Build is not actually passing: https://travis-ci.com/github/yahoo/navi/jobs/415184500

jkusa:
Needs rebase and new screenshots after Denali. Closing for now

PR #907, opened 2020-07-24
Update: Use `Uncategorized` for grouped-list falsy group values

mattgraham1995:
## Description
We shouldn't use `undefined` or `null` as grouping values in the grouped-list

## Proposed Changes

- Use `Uncategorized` as the grouping value for falsy values

## License


packages/reports/addon/components/grouped-list.js
```diff
@@ -19,6 +19,7 @@ import { isBlank } from '@ember/utils';
 import layout from '../templates/components/grouped-list';
 import { groupBy, sortBy } from 'lodash-es';
 import { layout as templateLayout, tagName } from '@ember-decorators/component';
+import { capitalize } from '@ember/string';
```
mattgraham1995, line 22:
```suggestion
```

PR #911, opened 2020-07-28
Carbon Query Logger

dvarshney1:
## Description
Added QueryLogging Implementation for Carbon in Navi. 

## Proposed Changes
In package/webservice, the following files were added to reflect this change:
1. CarbonQueryLogger.kt: Has the implementation for Query Logging and overrides the QueryLogger interface from Elide.
2. QueryStatus.kt: Enum class with all the possible query statuses that a query could have when an response is received from Elide
3. QueryStats.kt: Our querylogging entity class.

## License

### DIFF
--- a/packages/webservice/app/build.gradle.kts
+++ b/packages/webservice/app/build.gradle.kts
@@ -11,15 +11,17 @@ plugins {
 
 repositories {
     mavenCentral()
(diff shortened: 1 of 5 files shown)


dvarshney1:
Closing this. Updated PR has been opened: #935

PR #930, opened 2020-07-31
Update: Add field parameter to dimensions

kevinhinterlong:
## Description
Navi makes a lot of assumptions regarding `id` and `desc` fields for dimensions

## Proposed Changes
- Every dimension now uses a `field` parameter defaulted to `'id'` we'll allow users to configure them later

## License

### DIFF
--- a/packages/data/addon/adapters/dimensions/bard.ts
+++ b/packages/data/addon/adapters/dimensions/bard.ts
@@ -8,23 +8,38 @@
 import { assert, warn } from '@ember/debug';
 import { inject as service } from '@ember/service';
 import { getOwner } from '@ember/application';
-import { assign } from '@ember/polyfills';
 import EmberObject from '@ember/object';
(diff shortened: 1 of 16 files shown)


packages/data/addon/adapters/dimensions/bard.ts
```diff
@@ -8,24 +8,39 @@
...
 
 const SEARCH_TIMEOUT = 30000;
 
 const CLIENT_ID = 'UI';
 
+type DimensionFilter = {
+  field: string;
+  operator: string;
+  values: string[];
+};
```
mattgraham1995, line 34: Should we enforce `type: 'dimension'` here?

kevinhinterlong: There’s no type for this object and the field is actually the dimension field. I didn’t think it was worth changing the interface since it is bard dimension specific

PR #931, opened 2020-07-31
Update: Refactor Metadata Serializers

jkusa:
## Description

Refactor Metadata Serializers

## Proposed Changes

- Create NaviMetadataSerializer interface
- Move Fili & Elide Serializers to `adapter:metedata` namespace
- Convert serializers to Typescript
- Move Bard metricFunctions fetch from service to adapter
- Refactor bard legacy `parameters` to `metricFunction` logic
- Update types

## License


packages/data/addon/serializers/metadata/metric-function.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param parameters - raw metric function parameters
+ * @param source - data source name
+ */
+export function constructFunctionArguments(
+  parameters: RawMetricFunctionArguments,
+  source: string
+): FunctionArgumentMetadataPayload[] {
+  return Object.keys(parameters).map(param => {
+    const paramObj: RawMetricFunctionArgument = parameters[param];
+    const { type, defaultValue, values, dimensionName, description } = paramObj;
```
mattgraham1995, line 32:
```suggestion
    const { type, defaultValue, values, dimensionName, description }: RawMetricFunctionArgument = parameters[param];
```
Can we do this with typescript?

packages/data/tests/unit/serializers/bard-test.js
```diff
@@ -0,0 +1,638 @@
+import { module, test } from 'qunit';
```
mattgraham1995, line 1: should this file be under `packages/data/tests/unit/serializers/metadata/bard-test.js` instead?

packages/data/tests/unit/serializers/bard-test.js
```diff
@@ -0,0 +1,638 @@
...
+        dimensions: Dimensions,
+        timeDimensions: TimeDimensions,
+        tables: Tables,
+        metricFunctions: ParameterConvertToMetricFunction
+      },
+      'One metric function is created for all metrics with only the currency parameter'
+    );
+  });
+
+  test('normalize `everything` with metric functions', function(assert) {
```
mattgraham1995, line 462: Do we have a test for when both legacy parameters and metric functions are provided? Since we're merging the constructed and provided metric functions now, it would be good to test that behavior

packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -0,0 +1,237 @@
...
+            const convertedToMetricFunction = this._getMetricFunctionFromParameters(metric, source);
+            if (convertedToMetricFunction) {
+              metric.metricFunctionId = convertedToMetricFunction.id;
+            }
+
+            const newMetric = this._constructMetric(metric, source);
+            currentMetrics[newMetric.id] = newMetric; // Add metric to all metrics list
+            tableMetricIds.add(newMetric.id); // Add metric id to table's metricIds list
+
+            if (convertedToMetricFunction) {
```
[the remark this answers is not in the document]
jkusa: good call

packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -0,0 +1,237 @@
...
+    });
+
+    const metricFunctions = rawMetricFunctions ? normalizeMetricFunctions(rawMetricFunctions, source) : [];
+
+    return {
+      tables,
+      dimensions: Object.values(dimensions),
+      metrics: Object.values(metrics),
+      timeDimensions: Object.values(timeDimensions),
+      metricFunctions: [...metricFunctions, ...Object.values(convertedToMetricFunctions)]
```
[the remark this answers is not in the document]
jkusa: Yeah, this will only happen if there are metrics with metric IDs and metrics with **only** `parameters` in a single data source. 
https://github.com/yahoo/navi/pull/931/files#diff-60c8edbf89613347ba5e51e792ff0808R180

packages/data/addon/serializers/metadata/metric-function.ts
jkusa: TS knows the shape from line 31.

mattgraham1995: Yeah but we're not using `paramObj` anywhere else so it seems like an unnecessary assignment.

packages/data/tests/unit/serializers/bard-test.js
jkusa: Yes, I moved this in my next PR: https://github.com/yahoo/navi/pull/933/files#diff-7dbc9ec7e182639bf180f0b38ddae3ea :)

packages/data/tests/unit/serializers/bard-test.js
jkusa: https://github.com/yahoo/navi/pull/933/files#diff-7dbc9ec7e182639bf180f0b38ddae3ea

jkusa: https://github.com/yahoo/navi/blob/c468cdac82f17003357ed717d967407c4fe8af78/packages/data/tests/unit/serializers/bard-test.js#L153-L183

PR #942, opened 2020-08-06
GraphQL Mock Query Argument Handling

mattgraham1995:
## Description
The elide fact mock is pretty bare bones and can't alter the results based on filters, sorts, or limits in the request.

## Proposed Changes

- Parse the GraphQL query for filters, sorts, and limits
- Generate values of the response based on all parts of the request

## License

### DIFF
--- a/packages/data/addon/mirage/handlers/graphql.js
+++ b/packages/data/addon/mirage/handlers/graphql.js
@@ -224,7 +224,6 @@ function _getDates(filters, requestedColumns) {
 }
 
 /**
- * @param {String} name
  * @param {Object} filter
  * @returns n random dimension values
  */
--- a/packages/data/addon/mirage/handlers/graphql.js
+++ b/packages/data/addon/mirage/handlers/graphql.js
@@ -7,9 +7,65 @@ import schema from 'navi-data/gql/schema';
 import gql from 'graphql-tag';
 import faker from 'faker';
 import moment from 'moment';
+import { capitalize } from '@ember/string';
+import { orderBy } from 'lodash-es';
 
-const API_DATE_FORMAT = 'YYYY-MM-DD';
 const ASYNC_RESPONSE_DELAY = 5000; // ms before async api response result is populated
+const DATE_FORMATS = {
+  hour: 'YYYY-MM-DD HH:MM:SS',
+  day: 'YYYY-MM-DD',
+  week: 'YYYY-MM-DD',
+  month: 'YYYY MMM',
+  quarter: 'YYYY [Q]Q',
+  year: 'YYYY'
+};
+const GRAINS = Object.keys(DATE_FORMATS);
+const TIME_DIMENSION_REGEX = new RegExp(
+  `(.+)(${Object.keys(DATE_FORMATS)
+    .map(grain => capitalize(grain))
+    .join('|')})`
+);
+const OPERATORS = {
+  eq: '==',
+  neq: '!=',
+  isIn: '=in=',
+  notIn: '=out=',
+  isNull: '=isnull=true',
+  notNull: '=isnull=false',
+  isEmpty: '=isempty=true',
+  lt: '=lt=',
+  gt: '=gt=',
+  le: '=le=',
+  ge: '=ge='
+};
+const FILTER_OPS = {
+  [OPERATORS.lt]: ([filterVal], vals) => vals.filter(val => val < filterVal),
+  [OPERATORS.gt]: ([filterVal], vals) => vals.filter(val => val > filterVal),
+  [OPERATORS.le]: ([filterVal], vals) => vals.filter(val => val <= filterVal),
+  [OPERATORS.ge]: ([filterVal], vals) => vals.filter(val => val >= filterVal),
+  [OPERATORS.isIn]: (filterVals, vals) => vals.filter(val => filterVals.includes(val)),
+  [OPERATORS.notIn]: (filterVals, vals) => vals.filter(val => !filterVals.includes(val)),
+  [OPERATORS.isNull]: (_, vals) => vals.filter(val => !val),
+  [OPERATORS.notNull]: (_, vals) => vals.filter(Boolean),
+  [OPERATORS.isEmpty]: () => [],
+  [OPERATORS.eq]: ([filterVal], vals) => vals.filter(val => val === filterVal),
+  [OPERATORS.neq]: ([filterVal], vals) => vals.filter(val => val !== filterVal)
+};
+const FILTER_REGEX = new RegExp(`(.*?)(?:\\((.+?)\\))?(${Object.values(OPERATORS).join('|')})\\((.+?)\\)`);
+
+//create an inclusive interval of dates based on filter values and operator
+const DATE_FILTER_OPS = {
+  [OPERATORS.lt]: ([val], grain) => [null, moment(val).subtract(1, grain)],
+  [OPERATORS.gt]: ([val], grain) => [moment(val).add(1, grain), null],
+  [OPERATORS.le]: ([val]) => [null, moment(val)],
+  [OPERATORS.ge]: ([val]) => [moment(val), null],
+  [OPERATORS.isIn]: () => [null, null], //TODO: Not sure if the following operators are really supported for dates
+  [OPERATORS.isNull]: () => [null, null],
+  [OPERATORS.notNull]: () => [null, null],
+  [OPERATORS.isEmpty]: () => [null, null],
+  [OPERATORS.eq]: () => [null, null],
+  [OPERATORS.neq]: () => [null, null]
+};
 
 function _getSeedForRequest(table, args, fields) {
   const tableLength = table.length;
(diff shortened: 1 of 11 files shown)


packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -7,9 +7,65 @@ import schema from 'navi-data/gql/schema';
...
+const FILTER_OPS = {
+  [OPERATORS.lt]: ([filterVal], vals) => vals.filter(val => val < filterVal),
+  [OPERATORS.gt]: ([filterVal], vals) => vals.filter(val => val > filterVal),
+  [OPERATORS.le]: ([filterVal], vals) => vals.filter(val => val <= filterVal),
+  [OPERATORS.ge]: ([filterVal], vals) => vals.filter(val => val >= filterVal),
+  [OPERATORS.isIn]: (filterVals, vals) => vals.filter(val => filterVals.includes(val)),
+  [OPERATORS.notIn]: (filterVals, vals) => vals.filter(val => !filterVals.includes(val)),
+  [OPERATORS.isNull]: (_, vals) => vals.filter(val => !val),
+  [OPERATORS.notNull]: (_, vals) => vals.filter(Boolean),
+  [OPERATORS.isEmpty]: () => [],
```
jkusa, line 50: Does this work?

packages/data/addon/serializers/elide-facts.ts
```diff
@@ -7,11 +7,11 @@
 
 import EmberObject from '@ember/object';
 import NaviFactSerializer, { ResponseV1 } from './fact-interface';
-import { AsyncQueryResponse, RequestV1 } from 'navi-data/adapters/fact-interface';
+import { AsyncQueryResponse, RequestV1, RequestV2 } from 'navi-data/adapters/fact-interface';
 
 export default class ElideFactsSerializer extends EmberObject implements NaviFactSerializer {
-  normalize(payload: AsyncQueryResponse, request: RequestV1): ResponseV1 | undefined {
-    const requestTable: string = request.logicalTable.table;
+  normalize(payload: AsyncQueryResponse, request: RequestV1 | RequestV2): ResponseV1 | undefined {
```
jkusa, line 13: Can we just assume v2 from now on?

jkusa, line 13: Can we just assume v2 from now on?

packages/data/tests/dummy/mirage/scenarios/graphql-blockhead.ts
```diff
@@ -1,11 +1,25 @@
 // TODO: Replace any with type supplied by new version of mirage
+import { grains } from './graphql';
```
jkusa, line 2: Can we rebase on this with the new file name?

packages/data/addon/mirage/handlers/graphql.js
jkusa: Not sure this makes sense in an analytics context. We don't support collection.

packages/data/addon/serializers/elide-facts.ts
kevinhinterlong: This is merging onto master not `request-v2`

packages/data/tests/dummy/mirage/scenarios/graphql-blockhead.ts
mattgraham1995: The new file names are in the request v2 branch and this is being merged into master

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -18,42 +74,136 @@ function _getSeedForRequest(table, args, fields) {
   return tableLength + argsLength + fieldsLength;
 }
 
+function _createInterval(filters) {
+  const intervals = filters.map(f => DATE_FILTER_OPS[f.operator](f.values, f.grain));
+  const lowerBound = moment.max(intervals.map(interval => interval[0]).filter(Boolean));
+  const upperBound = moment.min(intervals.map(interval => interval[1]).filter(Boolean));
```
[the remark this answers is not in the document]
mattgraham1995: Is it not? I believe it works for finding the intersection of multiple intervals. Do you have a counter example that it wouldn't work for?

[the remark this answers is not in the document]
kevinhinterlong:
```js
const intervals = [
  [‘2020-01-01’, ‘2020-01-04’],
  [‘2020-01-01’, ‘2020-01-02’],
  [‘2020-01-02’, ‘2020-01-03’],
  [‘2020-01-03’, ‘2020-01-04’],
]

const lowerBound = ‘2020-01-03’
const upperBound = ‘2020-01-02’
```

So it should technically be `2020-01-01/2020-01-04`, but this gives `2020-01-03/2020-01-02`. Although we could just accept this as a limitation. Didn’t look too closely, but does it handle invalid intervals?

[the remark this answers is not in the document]
mattgraham1995: Why should it be `2020-01-01/2020-01-04`? This is condensing multiple intervals on a single time dimension down to one interval so it should create an interval that's an intersection of those. The returned values should be valid for all the filtered intervals right? 

I do need to handle the case where the lowerInterval is greater than the upperInterval though. Invalid intervals should return an empty array for dates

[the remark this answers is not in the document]
mattgraham1995: I added logic to handle the case where lowerBound is later than upperBound and for intervals missing either an upper or lower bound. I also added some tests to make sure the mock's handling of those cases is correct.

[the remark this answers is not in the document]
kevinhinterlong: You're right, I was thinking of fili intervals rather than time dimension filters (since we `&&` filters together) so that is the expected behavior. Although I wonder how we would request data over multiple distinct intervals given this format.

Maybe we could have 
```js
{
  type: 'timeDimension',
  field: 'dateTime',
  parameters: { grain: 'day' },
  operator: 'bet',
  values: ['2020-01-01', '2020-01-02', '2020-01-04', '2020-01-05']
}
```
which would turn into `dateTime=2020-01-01/2020-01-02,2020-01-04/2020-01-05` for fili or just opt for multiple requests

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -19,41 +80,173 @@ function _getSeedForRequest(table, args, fields) {
...
+  } else if (!interval[1]) {
+    start = interval[0].startOf(grain);
+    let monthAhead = moment(start).add(1, 'month');
+    let current = moment();
+    end = monthAhead.isSameOrBefore(current) ? monthAhead.startOf(grain) : current.startOf(grain); //Default end to current or 1 month past start, whichever is closer to start
+  } else {
+    start = interval[0].startOf(grain);
+    end = interval[1].startOf(grain);
+  }
+  for (let i = start; i.isSameOrBefore(end); i.add(1, grain)) {
```
kevinhinterlong, line 151: I think we can remove this and the `startOf(grain)` logic and rely on [getDatesForInterval](https://github.com/yahoo/navi/blob/893642dac365059d1d5cf374b8784431cc687173/packages/core/addon/utils/date.js#L27) One other thing that takes into account is the difference between `week`/`isoweek`. I'm not sure which elide is reporting on but with fili it's the isoweek

packages/data/addon/mirage/handlers/graphql.js
```diff
@@ -129,6 +377,20 @@ function _getResponseBody(db, parent) {
         )
       );
 
+      // handle limit in request
+      if (first < rows.length) {
```
kevinhinterlong, line 381: Can first be null? I think in request v2 we decided that limit is `number|null`

packages/data/tests/unit/services/navi-facts-elide-test.js
```diff
@@ -33,6 +34,67 @@ const TestRequest = {
...
+      values: ['2014-01-07'],
+      parameters: {},
+      type: 'timeDimension'
+    }
+  ],
+  sorts: [
+    { field: 'eventTimeDay', parameters: {}, type: 'timeDimension', direction: 'asc' },
+    { field: 'dimension2', parameters: {}, type: 'dimension', direction: 'desc' }
+  ],
+  limit: 15,
```
kevinhinterlong, line 93: Can we add a test for sort and for limit? One with limit being null too

packages/data/addon/mirage/handlers/graphql.js
mattgraham1995: yeah when it's null this evaluates to false and no limit is applied

mattgraham1995: Oh actually `first` will be undefined instead of null because when the request has a null limit, it just doesn't include it in the query. `undefined < ${number}` will always be false

kevinhinterlong: 👍 gotcha, worth having a test though

packages/data/addon/mirage/handlers/graphql.js
mattgraham1995: Yeah unfortunately navi-data doesn't have access to the interval util class since navi-core depends on navi-data. Good call on isoWeek though. I'm sure elide uses that as well.

kevinhinterlong: Oops, yea I thought we used that for the bard mock

mattgraham1995:
Moving to a PR to merge into `request-v2` instead

PR #944, opened 2020-08-06
Upgrading to Elide 5 PR 15

aklish:
Upgrades to Elide 5 PR 15 build.

There is a security change in PR15 that breaks the User bean create permission.  This is a workaround pending a more expensive modeling change in Navi.

I also added the Hibernate validator (which was missing) which turns on JSR303 validations.

## License

### DIFF
--- a/packages/webservice/app/build.gradle.kts
+++ b/packages/webservice/app/build.gradle.kts
@@ -16,9 +16,10 @@ repositories {
 dependencies {
     implementation(project(":models"))
     implementation("org.springframework.boot:spring-boot-starter-security")
-    implementation("com.yahoo.elide", "elide-spring-boot-starter", "5.0.0-pr12")
+    implementation("com.yahoo.elide", "elide-spring-boot-starter", "5.0.0-pr15")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("com.h2database", "h2", "1.3.176")
+    implementation( "org.hibernate", "hibernate-validator", "6.1.5.Final")
     implementation("io.micrometer","micrometer-core", "1.5.1")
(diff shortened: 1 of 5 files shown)


packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/User.kt
```diff
@@ -4,13 +4,11 @@
  */
 package com.yahoo.navi.ws.models.beans
 
-import com.yahoo.elide.annotation.CreatePermission
-import com.yahoo.elide.annotation.DeletePermission
-import com.yahoo.elide.annotation.Include
-import com.yahoo.elide.annotation.UpdatePermission
+import com.yahoo.elide.annotation.*
```
moizarafat, line 7: We can use individual imports instead of `* `.

packages/webservice/app/build.gradle.kts
```diff
@@ -16,9 +16,10 @@ repositories {
 dependencies {
     implementation(project(":models"))
     implementation("org.springframework.boot:spring-boot-starter-security")
-    implementation("com.yahoo.elide", "elide-spring-boot-starter", "5.0.0-pr12")
+    implementation("com.yahoo.elide", "elide-spring-boot-starter", "5.0.0-pr15")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("com.h2database", "h2", "1.3.176")
+    implementation( "org.hibernate", "hibernate-validator", "6.1.5.Final")
```
moizarafat, line 22: is this dependency needed? The changes in the PR don't seem to be making use of validators.

aklish: Without this, none of the existing validators actually work.

PR #955, opened 2020-08-12
Update: Create columnFunction for dimension fields

### LINKED ISSUE #921: Normalize fili fields as values to  a `field` parameter

kevinhinterlong:
Resolves #921 

## Description
Currently in request v2, we have no way of selecting the `field` parameter on dimension columns. Since we've made columnFunction generic, we can now link to it from a dimension and create a function for picking from those dimension fields

## Proposed Changes
Create a columnFunction for every unique set of dimension fields and link to them as a `field` parameter on dimensions

## License

### DIFF
--- a/packages/data/addon/models/metadata/dimension.ts
+++ b/packages/data/addon/models/metadata/dimension.ts
@@ -23,7 +23,7 @@ export interface DimensionMetadata extends ColumnMetadata {
 export interface DimensionMetadataPayload extends ColumnMetadataPayload {
   fields?: Field[];
   cardinality?: Cardinality;
-  storageStrategy?: 'loaded' | 'none';
+  storageStrategy?: TODO<'loaded' | 'none' | null>;
 }
 
 export default class DimensionMetadataModel extends ColumnMetadataModel
@@ -133,7 +133,7 @@ export default class DimensionMetadataModel extends ColumnMetadataModel
     return field?.name || this.primaryKeyFieldName;
   }
 
-  storageStrategy?: 'loaded' | 'none';
+  storageStrategy?: TODO<'loaded' | 'none' | null>;
 
   /**
    * @property {Promise} extended - extended metadata for the dimension that isn't provided in initial table fullView metadata load
--- a/packages/data/addon/serializers/metadata/bard.ts
+++ b/packages/data/addon/serializers/metadata/bard.ts
@@ -17,7 +17,6 @@ import {
   INTRINSIC_VALUE_EXPRESSION,
   ColumnFunctionParametersValues
 } from 'navi-data/models/metadata/function-parameter';
-import { bind } from 'lodash-es';
 
 const LOAD_CARDINALITY = config.navi.searchThresholds.contains;
 const MAX_LOAD_CARDINALITY = config.navi.searchThresholds.in;
@@ -38,12 +37,11 @@ type RawColumnPayload = {
   description?: string;
   longName: string;
   category?: string;
-  uri?: string;
 };
 
 export type RawDimensionPayload = RawColumnPayload & {
   datatype: TODO<'text' | 'date'>;
-  storageStrategy?: TODO<'loaded' | 'none' | null>;
+  storageStrategy?: TODO<'loaded' | 'none'>;
   cardinality: number;
   fields: RawDimensionField[];
 };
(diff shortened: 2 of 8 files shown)


packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -283,23 +335,44 @@ export default class BardMetadataSerializer extends EmberObject implements NaviM
...
+   * @param dimensions - raw dimensions
+   * @param dataSourceName - data source name
+   */
+  private normalizeTimeDimensions(
+    table: RawTablePayload,
+    dimensions: RawDimensionPayload[],
+    dataSourceName: string
+  ): TimeDimensionMetadataPayload[] {
+    return this.normalizeDimensions(dimensions, dataSourceName)
+      .filter(d => d.valueType === 'date')
```
jkusa, line 368: Do we have a test that excutes this?

packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -59,31 +65,30 @@ export type RawColumnFunctionArguments = {
...
   name: string;
   longName: string;
-  description: string;
-  category: string;
+  description?: string;
+  category?: string;
 };
 
 export default class BardMetadataSerializer extends EmberObject implements NaviMetadataSerializer {
+  private namespace = '_fili_generated_';
```
jkusa, line 90: This name might give a little more context:
```suggestion
  private namespace = 'normalizer-generated_';
```

packages/data/addon/models/metadata/dimension.ts
```diff
@@ -23,7 +23,7 @@ export interface DimensionMetadata extends ColumnMetadata {
 export interface DimensionMetadataPayload extends ColumnMetadataPayload {
   fields?: Field[];
   cardinality?: Cardinality;
-  storageStrategy?: 'loaded' | 'none';
+  storageStrategy?: TODO<'loaded' | 'none' | null>;
```
jkusa, line 26: Are we planning to come back to this?

packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -13,39 +13,45 @@ import NaviMetadataSerializer, { MetadataPayloadMap, EverythingMetadataPayload }
...
+};
+
+type RawColumnPayload = {
   name: string;
   description?: string;
   longName: string;
-  category: string;
-  storageStrategy: TODO;
+  category?: string;
+  uri?: string;
```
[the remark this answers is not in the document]
kevinhinterlong: It was being used in tests, I'll remove it

packages/data/addon/serializers/metadata/bard.ts
kevinhinterlong: Yep, we see it from this timeDimension
https://github.com/yahoo/navi/blob/c96e5c5404e4fd9d1cc097d4220d15b454809161/packages/data/tests/unit/serializers/metadata/bard-test.ts#L407-L414

packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -113,15 +118,26 @@ export default class BardMetadataSerializer extends EmberObject implements NaviM
 
           // Construct each dimension / time-dimension
           timegrain.dimensions.forEach(dimension => {
-            const { datatype: valueType } = dimension;
-            const accDimensionList = valueType === 'date' ? currentTimeDimensions : currentDimensions;
-            const accTableDimensionList = valueType === 'date' ? tableTimeDimensionIds : tableDimensionIds;
+            const isTimeDimension = dimension.datatype === 'date';
+
+            const normalize = isTimeDimension
+              ? bind(this.normalizeTimeDimensions, this, table) // call function with table partially applied
```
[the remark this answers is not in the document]
kevinhinterlong: Looks like I have to manually specify the types, both this and lodash were using `any`

packages/data/addon/models/metadata/dimension.ts
kevinhinterlong: Maybe. It should at least be better defined because previously we had this as what was returned from fili
https://github.com/yahoo/navi/blob/71325fcfe781aac9bdc20f874437ff9d9bc72d0f/packages/data/addon/serializers/metadata/bard.ts#L29-L39
so `storageStrategy` was always defined, but we also had this
https://github.com/yahoo/navi/blob/c96e5c5404e4fd9d1cc097d4220d15b454809161/packages/data/addon/serializers/metadata/bard.ts#L408

so that's why I'm updating it

PR #956, opened 2020-08-12
Update: Add Navi dimension interfaces

jkusa:
## Description

Add Navi dimension interfaces

## Proposed Changes

- Dimension adapter infterface
- Dimension serializer interface
- Dimension base model

## License

### DIFF
--- a/packages/data/addon/adapters/dimensions/interface.ts
+++ b/packages/data/addon/adapters/dimensions/interface.ts
@@ -41,7 +41,7 @@ export default interface NaviDimensionAdapter {
   /**
    * Get dimension values for a search string
    * @param dimension - requested dimension
-   * @param string - query string
+   * @param query - query string
    * @param options - adapter options
    */
   search(dimension: DimensionColumn, query: string, options?: AdapterOptions): Promise<unknown>;
--- a/packages/data/addon/adapters/dimensions/interface.ts
+++ b/packages/data/addon/adapters/dimensions/interface.ts
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import { Parameters } from '../facts/interface';
+import DimensionMetadataModel from 'navi-data/models/metadata/dimension';
+
+export type AdapterOptions = {
+  clientId?: string;
+  timeout?: number;
+  page?: number;
+  perPage?: number;
+};
+
+export type DimensionFilter = {
+  operator: string;
+  values: string[];
+};
+
+export interface DimensionColumn {
+  columnMetadata: DimensionMetadataModel;
+  parameters?: Parameters;
+}
+
+export default interface NaviDimensionAdapter {
+  /**
+   * Get all values for a dimension column
+   * @param dimension - requested dimension
+   * @param options - adapter options
+   */
+  all(dimension: DimensionColumn, options?: AdapterOptions): Promise<unknown>;
+
+  /**
+   * Get dimension value for a filter predicate
+   * @param dimension - requested dimension
+   * @param predicate - filter criteria
+   * @param options - adapter options
+   */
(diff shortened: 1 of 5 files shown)


packages/data/tests/unit/models/navi-dimension-test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface TestContext extends Context {
+  metadataService: NaviMetadataService;
+}
+
+module('Unit | Model | navi dimension', function(hooks) {
+  setupTest(hooks);
+  setupMirage(hooks);
+  hooks.beforeEach(async function(this: TestContext) {
+    this.metadataService = this.owner.lookup('service:navi-metadata') as NaviMetadataService;
```
kevinhinterlong, line 19: Can we just assign this as a top level object instead of having to set `this: TestContext` in the tests

packages/data/addon/models/navi-dimension.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import EmberObject from '@ember/object';
+import { DimensionColumn } from 'navi-data/adapters/dimensions/interface';
+
+export default class NaviDimensionModel extends EmberObject {
+  /**
+   * Dimension column definition
```
kevinhinterlong, line 10: `@property` jsdoc

packages/data/addon/models/navi-dimension.ts
```diff
@@ -0,0 +1,25 @@
...
+   */
+  dimensionColumn!: DimensionColumn;
+
+  /**
+   * Dimension value
+   */
+  value!: unknown;
```
kevinhinterlong, line 17: For fili are you thinking something like `{ id: 1, description: 'dim description' }`?

packages/data/addon/adapters/dimensions/interface.ts
```diff
@@ -0,0 +1,48 @@
...
+   * @param dimension - requested dimension
+   * @param predicate - filter criteria
+   * @param options - adapter options
+   */
+  find(dimension: DimensionColumn, predicate: DimensionFilter[], options?: AdapterOptions): Promise<unknown>;
+
+  /**
+   * Get dimension values for a search string
+   * @param dimension - requested dimension
+   * @param string - query string
```
kevinhinterlong, line 44:
```suggestion
   * @param query - query string
```

packages/data/addon/models/navi-dimension.ts
jkusa: Do we need this? VSCode seems to understand this is a property:

Screen Shot 2020-08-11 at 10 58 40 PM

packages/data/addon/models/navi-dimension.ts
jkusa: For Fili it might look like this: 

```ts
export default class FiliDimensionModel extends NaviDimensionModel {
  private rawValue = { id: 1, description: 'dim description' };

  get value() {
    return this.rawValue[this.dimensionColumn.parameters.field];
  }

  get displayValue() {
   return `${this.value} (${this.dimensionContext})`
  }
}
```

packages/data/tests/unit/models/navi-dimension-test.ts
jkusa: I know we have done that in the past. Instead of relying on global like variables between tests, I feel this better documents what is generally available in the test environment.

packages/data/addon/models/navi-dimension.ts
kevinhinterlong: I think you're right

packages/data/tests/unit/models/navi-dimension-test.ts
kevinhinterlong: If we didn't have to specify `this` for every test I wouldn't have any objections. I don't feel too strongly either way, but I'm more in favor of how we did it before

PR #962, opened 2020-08-14
Update: Fix elide request and response names

### LINKED ISSUE #953: Elide Fact Processing - Remove table id from Elide column ids in adapter request
**Is your feature request related to a problem? Please describe.**
We expect elide column ids to be prefixed with their table ids in the metadata. This won't work when we actually send those ids in a request to elide. When we query elide, we have to query against a table so the table id at the front of the column id is not necessary.

Request should have the column id without the table id attached to it
`tableA.column1` => `column1`

So the adapter should format the column ID like so
```graphql
query {
  tableA {
    edges {
      node {
        column1 # tableA.column1 => column1
      }
    }
  }
}
```

The table id is the name of the top level query

### LINKED ISSUE #961: Normalize Elide Fact Response Fields.

jkusa:
Resolves #953 
Resolves #961

## Description

Update: Fix elide request and response names

## Proposed Changes

- Update elide adapter and serializer
- Move `queryStrForField` private method to export function `getElideField`

## License

### DIFF
--- a/packages/data/addon/serializers/facts/elide.ts
+++ b/packages/data/addon/serializers/facts/elide.ts
@@ -8,19 +8,43 @@
 import EmberObject from '@ember/object';
 import NaviFactSerializer, { ResponseV1 } from './interface';
 import { AsyncQueryResponse, RequestV2 } from 'navi-data/adapters/facts/interface';
+import { getElideField } from 'navi-data/adapters/facts/elide';
+import { canonicalizeMetric } from 'navi-data/utils/metric';
 
 export default class ElideFactsSerializer extends EmberObject implements NaviFactSerializer {
-  normalize(payload: AsyncQueryResponse, request: RequestV2): ResponseV1 | undefined {
-    const requestTable = request.table;
-    const responseStr = payload?.asyncQuery.edges[0].node.result?.responseBody;
-    if (responseStr) {
-      const response = JSON.parse(responseStr);
-      const rawRows = response.data[requestTable].edges;
-      return {
-        rows: rawRows.map((row: { node: TODO }) => row.node),
-        meta: {}
-      };
+  /**
+   * @param payload - raw payload string
+   * @param request - request object
+   */
+  private processResponse(payload: string, request: RequestV2): ResponseV1 {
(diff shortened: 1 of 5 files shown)


packages/data/addon/serializers/facts/elide.ts
```diff
@@ -8,19 +8,43 @@
...
+    const totalFields = normalizedFields.length;
+    const rows = new Array(totalRows);
+
+    let r = totalRows;
+    while (r--) {
+      const rawRow = rawRows[r].node;
+      const newRow: Record<string, unknown> = (rows[r] = {});
+      let f = totalFields;
+      while (f--) {
+        newRow[normalizedFields[f]] = rawRow[elideFields[f]];
```
mattgraham1995, line 40: We'll need to confirm that the fields in the response will have the parameters added to them in the same format. Can we add a TODO to revisit this once Elide has parameterized column support?

mattgraham1995: Also, do we have a test with a parameterized column?

jkusa: We need to revisit elide parameters. We may need to implement aliases. Currently, the mock does not response with the params and requesting the same field with two different param values returns nothing. Since elide does not support params at the moment, we can skip the test.

packages/data/addon/serializers/facts/elide.ts
```diff
@@ -8,19 +8,43 @@
...
 export default class ElideFactsSerializer extends EmberObject implements NaviFactSerializer {
+  /**
+   * @param payload - raw payload string
+   * @param request - request object
+   */
+  private processResponse(payload: string, request: RequestV2): ResponseV1 {
+    const response = JSON.parse(payload);
+
+    const { table } = request;
+    const elideFields = request.columns.map(({ field, parameters }) => getElideField(field, parameters));
```
jkusa, line 23:
```suggestion
    // TODO revisit when Elide adds parameter support
    const elideFields = request.columns.map(({ field, parameters }) => getElideField(field, parameters));
```

PR #965, opened 2020-08-17
Fix: Component to determine if show ids on params

cythrawll:
- supports power-select grouped content

## Description

Need to support power select options for display. Was going to make the (name (id)) display logic complex so extracted logic to component.

## Proposed Changes

- Added component to replace the loooooooong if logic and normalized power select options.
- Added flat/flatMap array support for typescript

## License


packages/reports/addon/templates/components/navi-column-config/metric.hbs
```diff
@@ -22,11 +22,10 @@
...
         as | arg |
       >
-        {{! TODO: Simplify this when we normalize function arguments }}
-        {{#if (eq (get (filter-by (if arg.name "name" "description") (if arg.name arg.name arg.description) allParams) "length") 1)}}
-          {{if arg.name arg.name arg.description}}
-        {{else}}
+        {{#if (should-show-id argument=arg params=allParams)}}
           {{if arg.name arg.name arg.description}} ({{arg.id}})
+        {{else}}
+          {{if arg.name arg.name arg.description}}
```
mattgraham1995, line 28: Maybe instead of a helper, it could be a component called `parameter-list-item` or something like that?

packages/reports/addon/helpers/should-show-id.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type Arg = {
+  name?: string;
+  description?: string;
+  groupName?: string;
+  options?: Array<Arg>;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function shouldShowId(_unused: any[], { argument, params }: { argument: Arg; params?: Arg[] }): any {
```
kevinhinterlong, line 18:
```suggestion
export function shouldShowId(_unused: never, { argument, params }: { argument: Arg; params?: Arg[] }): boolean {
```

packages/reports/types/global.d.ts
```diff
@@ -7,3 +7,8 @@ declare module 'navi-reports/templates/*' {
 
 type Dict<T = string> = { [key: string]: T };
 type TODO<T = any> = T;
+
+interface Array<T> {
+  flat(): Array<T>;
+  flatMap(func: (x: T) => T): Array<T>;
+}
```
kevinhinterlong, line 14: Revert this and update every [tsconfig](https://stackoverflow.com/a/53559473)
```diff

    "target": "es2017",
+    "lib": [
+      "es2019"
+    ],
```

packages/reports/addon/helpers/should-show-id.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!Array.isArray(params)) {
+    return true;
+  }
+  const paramArray = params as Array<Arg>;
```
kevinhinterlong, line 22: After tsconfig update, we can remove the `as __` casts and this intermediate variable

packages/reports/addon/helpers/should-show-id.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * Helper to determine if we sould see the id field, next to name/description in a list.
+ * Now supports arguments list that are shaped for powerSelect grouped content.
+ */
+import { helper } from '@ember/component/helper';
+
+type Arg = {
```
kevinhinterlong, line 10: Could maybe make this a union type?
```ts
type Arg =
  | { name: string; description?: string }
  | { name: string; description?: string; groupName: string; options: Array<Arg> };
```

packages/reports/tests/integration/components/parameter-list-item-test.ts
```diff
@@ -0,0 +1,49 @@
...
+module('Integration | Component | parameter-list-item', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it decides whether to show id or not', async function(assert) {
+    this.set('arg', { name: 'foo', id: 'sid' });
+    this.set('allParams', [{ name: 'foo' }, { name: 'foo' }, { name: 'bar' }]);
+
+    await render(hbs`<ParameterListItem @argument={{this.arg}} @parameters={{this.allParams}} />`);
+
+    assert.equal(this.element?.textContent?.trim(), 'foo (sid)', 'should show id when name is duplicated');
```
mattgraham1995, line 15: Can we use `assert.dom`?

cythrawll: good call

PR #979, opened 2020-08-26
Fixes for filter builder tests

holderb5:
# Description
Resolve test failures for filter builders by changing to request-v2 model structures 
-

## License

### DIFF
--- a/packages/reports/addon/components/filter-builders/base.js
+++ b/packages/reports/addon/components/filter-builders/base.js
@@ -18,7 +18,6 @@ class BaseFilterBuilderComponent extends Component {
    * @property {Object} requestFragment;
    * @property {Object} filter
    * @property {Object} filter.subject
-   * @property {String} requestFragmentname - display name for the filter subject
    * @property {Object} filter.operator - object with the same shape as the objects in `supportedOperators` property
    * @property {String} filter.operator.name - display name for operator
    * @property {String} filter.operator.valuesComponent - name of component to use for selecting filter values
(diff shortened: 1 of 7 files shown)


packages/reports/addon/components/filter-builders/base.js
```diff
@@ -27,10 +27,13 @@ class BaseFilterBuilderComponent extends Component {
...
   }
 
+  get requestFragment() {
+    return undefined;
+  }
   /**
    * @property {String} displayName - display name for the filter
    */
-  @readOnly('filter.subject.columnMetadata.name') displayName;
+  @readOnly('requestFragment.columnMetadata.name') displayName;
```
jkusa, line 36: should this be?

```suggestion
  @readOnly('filter.columnMetadata.name') displayName;
```

packages/reports/tests/integration/components/filter-builders/base-test.js
```diff
@@ -8,18 +8,22 @@ import Component from '@ember/component';
...
-  subject: {
+const requestFragment = {
+  type: 'dimension',
+  columnMetadata: {
     name: 'Device Type'
-  },
+  }
+};
+
+const filter = {
```
jkusa, line 18: Can we simply use this object/shape:

https://github.com/yahoo/navi/blob/request-v2/packages/core/addon/models/bard-request-v2/fragments/filter.ts

jkusa:
As discussed offline, need to rework the shape of the filter args. Closing for now.

PR #991, opened 2020-09-08
Update: Add displayName getter in base.ts, and tests for it in column…

holderb5:
Resolves #991 
Resolves the above for local testing, not for travis-ci

## Description
Add in a displayName getter into base.ts fragment, as well as add tests for it and the column display name getter

## License


packages/core/addon/models/bard-request-v2/fragments/base.ts
```diff
@@ -13,6 +13,8 @@ import { canonicalizeMetric } from 'navi-data/utils/metric';
 import NaviMetadataService from 'navi-data/services/navi-metadata';
 import { Parameters, ColumnType } from 'navi-data/adapters/facts/interface';
 import MetadataModelRegistry from 'navi-data/models/metadata/registry';
+// @ts-ignore
```
kevinhinterlong, line 16: I think this should be typed
```suggestion
```

packages/core/tests/unit/models/bard-request-v2/fragments/column-test.js
```diff
@@ -120,4 +120,18 @@ module('Unit | Model | Fragment | BardRequest - Column', function(hooks) {
...
     );
   });
+
+  test('Display Name', async function(assert) {
+    const column = mockModel.columns.objectAt(0);
+
+    column.set('field', 'revenue');
+    column.set('parameters', { currency: 'USD' });
+    column.set('type', 'metric');
+    column.set('alias', 'revenueUSD');
```
jkusa, line 130: Can we add a non parameter test?

packages/core/tests/unit/models/bard-request-v2/fragments/column-test.js
```diff
@@ -120,4 +120,18 @@ module('Unit | Model | Fragment | BardRequest - Column', function(hooks) {
...
+
+    assert.equal(column.displayName, 'revenueUSD (USD)');
```
jkusa, line 132: Can we add assert messages?

packages/core/addon/models/bard-request-v2/fragments/column.ts
```diff
@@ -18,7 +18,6 @@ export default class ColumnFragment extends BaseFragment implements Column {
 
   @service naviFormatter!: NaviFormatterService;
 
-  @computed('columnMetadata', 'alias')
```
kevinhinterlong, line 21: The import wasn't removed

packages/data/addon/services/navi-formatter.ts
```diff
@@ -13,7 +13,7 @@ export default class NaviFormatterService extends Service {
     const paramValues = Object.values(allParams);
 
     const name = alias || columnMetadata?.name || '--';
```
kevinhinterlong, line 15:
```suggestion
    const name = columnMetadata?.name || '--';
```

packages/data/addon/services/navi-formatter.ts
```diff
@@ -13,7 +13,7 @@ export default class NaviFormatterService extends Service {
...
-    if (paramValues.length) {
+    if (paramValues.length && alias === null) {
       return `${name} (${paramValues.join(',')})`;
     } else {
       return name;
```
kevinhinterlong, line 19:
```suggestion
    if (paramValues.length && !alias) {
      return `${name} (${paramValues.join(',')})`;
    } else {
      return alias;
```

holderb5: So I tried this specific format and it caused a few errors with the tests, so I stuck with the original but changes alias === null to !alias

jkusa (approved):

packages/data/addon/services/navi-formatter.ts
```diff
@@ -13,7 +13,7 @@ export default class NaviFormatterService extends Service {
     const paramValues = Object.values(allParams);
 
     const name = alias || columnMetadata?.name || '--';
-    if (paramValues.length) {
+    if (paramValues.length && !alias) {
       return `${name} (${paramValues.join(',')})`;
     } else {
       return name;
```
kevinhinterlong, line 19: Can we just add something like this at the top of the function. I feel like that would be more explicit than this check
```js
if (alias) {
  return alias;
}
```

holderb5: Sure I'll try that!

kevinhinterlong (approved):

jkusa (approved):

PR #1024, opened 2020-09-15
Update: Use Column cid for table config/viz

### LINKED ISSUE #1001: RequestV2: Update Table Visualization
- [x] Use request v2 fragment
- [x] make use of `cid` if applicable
- [x] Don't expect `dateTime` field
- [x] Remove `id` & `desc` field references
- [x] Use column displayName

kevinhinterlong:
Resolves #1001 

## Description
Uses cid to store attributes per column

## License


packages/core/addon/utils/request.ts
```diff
@@ -166,14 +167,16 @@ export function normalizeV1toV2(request: RequestV1<string>, dataSource: string):
 
   //normalize dateTime column
   requestV2.columns.push({
+    cid: nanoid(10),
```
jkusa, line 170: Does the `defaultValue` not kick in here?

packages/data/addon/adapters/facts/interface.ts
```diff
@@ -43,6 +43,7 @@ export type SortDirection = typeof SORT_DIRECTIONS[number];
 export type ColumnType = 'metric' | 'dimension' | 'timeDimension';
 
 export type Column = {
+  cid: string;
```
jkusa, line 46: Is this required? I was thinking this would only be a PersistedColumn property as the adapter doesn't make use of it.

packages/core/addon/utils/request.ts
kevinhinterlong: I was just copying the implementation from the default value
https://github.com/yahoo/navi/blob/f804f05c6960fe7be5cca376ab90806624b91949/packages/core/addon/models/bard-request-v2/fragments/column.ts#L16-L17

packages/data/addon/adapters/facts/interface.ts
kevinhinterlong: For the most part no, but I do use it in the table serializer which uses the serialized version of request-v2. So I was expecting that it would be easier to keep `RequestV2` as the type also representing a serialized request fragment

jkusa:
```suggestion
  cid?: string;
```

jkusa (approved):

packages/core/addon/components/navi-cell-renderers/base.ts
```diff
@@ -18,7 +18,7 @@ export default class BaseCellRenderer extends Component<CellRendererArgs> {
   /**
    * value of the column
    */
-  @computed('args.column.fragment.canonicalName')
+  @computed('args.{column.fragment.canonicalName,data.[]}')
```
mattgraham1995, line 21: This is it Chief

mattgraham1995 (approved):

PR #1040, opened 2020-09-18
Fix: Missing dependent key in base cell renderer

### LINKED ISSUE #1037: Fix: Table Rendering Issue

mattgraham1995:
Resolves #1037 

## Description
Table occlusion was causing cells to reorder on scroll because cells weren't updating when their data changed, only when their canonical name changed

## Proposed Changes

- Update cell renderer value to update when data argument changes

## License


cythrawll (approved):

mattgraham1995:
Looks like #1024 already has this change. Closing in favor of that.

PR #1042, opened 2020-09-18
Update: Fixup for optional cid in table serializer

kevinhinterlong:
## Description
Optional cid on serialized RequestV2 makes it possibly undefined so add an assert requiring it to be defined

## License

### DIFF
--- a/packages/core/addon/serializers/table.ts
+++ b/packages/core/addon/serializers/table.ts
@@ -6,6 +6,7 @@
 import VisualizationSerializer from './visualization';
 import { parseMetricName, canonicalizeMetric } from 'navi-data/utils/metric';
 import { RequestV2, Column } from 'navi-data/adapters/facts/interface';
+import { assert } from '@ember/debug';
 
 interface FieldTypes {
   metric: 'metric';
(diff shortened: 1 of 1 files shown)


mattgraham1995 (approved):

packages/core/addon/serializers/table.ts
```diff
@@ -164,6 +165,7 @@ export function normalizeTableV2(
   const columnAttributes = Object.values(columnData).reduce((columns, columnInfo) => {
     const { tableColumn, requestColumn } = columnInfo;
     const { attributes } = tableColumn;
+    assert(`The request column ${requestColumn.field} has a cid`, requestColumn.cid);
```
jkusa, line 168:
```suggestion
    assert(`The request column ${requestColumn.field} does not have a present 'cid' field`, isPresent(questColumn.cid));
```

??

packages/core/addon/serializers/table.ts
```diff
@@ -6,6 +6,7 @@
 import VisualizationSerializer from './visualization';
 import { parseMetricName, canonicalizeMetric } from 'navi-data/utils/metric';
 import { RequestV2, Column } from 'navi-data/adapters/facts/interface';
+import { assert } from '@ember/debug';
```
jkusa, line 9:
```suggestion
import { assert } from '@ember/debug';
import { isPresent } from '@ember/utils';```

File: packages/core/addon/serializers/table.ts (hunk 1 above)
Comment (mattgraham1995) [reply packages/core/addon/serializers/table.ts:168]: I think we've used `should` pretty frequently for assert messages in the past. 
```suggestion
    assert(`The request column ${requestColumn.field} should have a present 'cid' field`, isPresent(questColumn.cid));
```

jkusa (approved):

PR #1050, opened 2020-09-23
Elide time macros

### LINKED ISSUE #1011: Support Time Macros in Elide Adapter

mattgraham1995:
Resolves #1011 

## Description
Elide doesn't have support for time macros like Fili does so we need to handle it in the adapter.

## Proposed Changes

- Move duration, interval, duration-utils, and utils/date to navi-data
- Move getDatesForInterval from utils/date into interval so as to break a circular dependency
- Convert timeDimension filters with 2 arguments in the elide adapter into date strings by creating an interval

## License


kevinhinterlong (approved):

packages/data/addon/adapters/facts/elide.ts
```diff
@@ -60,16 +63,24 @@ export default class ElideFactsAdapter extends EmberObject implements NaviFactAd
...
-      const { field, parameters, operator, values } = filter;
+      const { field, parameters, operator, values, type } = filter;
       const fieldStr = getElideField(field, parameters);
+      const operatorStr = OPERATOR_MAP[operator as FilterOperator] || `=${operator}=`;
+      let filterVals = values;
+      if (type === 'timeDimension' && filterVals.length === 2) {
+        const { start, end } = Interval.parseFromStrings(String(filterVals[0]), String(filterVals[1])).asMoments();
+        filterVals = [
+          start.format(ELIDE_API_DATE_FORMAT),
+          end?.format(ELIDE_API_DATE_FORMAT) || moment().format(ELIDE_API_DATE_FORMAT)
```
jkusa, line 74: Can end be undefined?

mattgraham1995: `Interval.parseFromStrings` potentially returns an `undefined` end, but we don't expect it to be undefined unless the end macro is `next` I believe. Since we're using `[inclusive,inclusive]` intervals with Elide, that shouldn't be an issue. I've just asserted that end is defined to get around the type check.

jkusa (approved):

PR #1051, opened 2020-09-25
Data export -  add resultType

ayeswarya:
Resolves #

## Description

Add resultType to AsyncQuery and AsyncQueryResult

## Proposed Changes

-

## License


mattgraham1995 (commented):
Is `resultType` already present on the API schema? If not, this change will break our current queries.

ayeswarya:
> Is `resultType` already present on the API schema? If not, this change will break our current queries.

Im not able to reply in comments, resultType is in the async API schema.

PR #1055, opened 2020-09-28
Update: Build using latest elide version and update Include rootLevel property

rishi-aga:
## Description

Build using latest elide version
Update @include annotation for Navi Beans (https://github.com/yahoo/elide/pull/1549/files)

## License

### DIFF
--- a/packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/hooks/UserValidationHook.kt
+++ b/packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/hooks/UserValidationHook.kt
@@ -28,7 +28,7 @@ class UserValidationHook : LifeCycleHook<User> {
         val principalName = requestScope?.user?.name
         val userName = user?.id
 
-        if (userName != principalName) {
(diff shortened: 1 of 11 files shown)


packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/hooks/UserValidationHook.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun execute(
+        operation: LifeCycleHookBinding.Operation?,
+        user: User?,
+        requestScope: RequestScope?,
+        changes: Optional<ChangeSpec>?
+    ) {
+        val principalName = requestScope?.user?.name
+        val userName = user?.id
+
+        if (userName != principalName) {
```
jkusa, line 31: Given the optionals could both provide `null`, I wonder if it is worth checking that `principalName` is not null and not empty

rishi-aga: added the check. I took this file from master branch so changes required there also.

jkusa (approved):

PR #1059, opened 2020-09-30
Update: Add not between operator to elide

### LINKED ISSUE #1044: Support `nbet` operator

holderb5:
Resolves #1044 

## License

### DIFF
--- a/packages/data/addon/adapters/facts/elide.ts
+++ b/packages/data/addon/adapters/facts/elide.ts
@@ -79,7 +79,7 @@ export default class ElideFactsAdapter extends EmberObject implements NaviFactAd
       }
 
       if (operator === 'nbet') {
-        return `${fieldStr}=le=(${filterVals[0]});${fieldStr}=ge=(${filterVals[1]})`;
+        return `${fieldStr}=lt=(${filterVals[0]});${fieldStr}=gt=(${filterVals[1]})`;
       }
 
       const valuesStr = filterVals.length ? `(${filterVals.map(v => `'${v}'`).join(',')})` : '';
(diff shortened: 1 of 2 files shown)


packages/data/addon/adapters/facts/elide.ts
```diff
@@ -78,6 +78,10 @@ export default class ElideFactsAdapter extends EmberObject implements NaviFactAd
         return `${fieldStr}=ge=(${filterVals[0]});${fieldStr}=le=(${filterVals[1]})`;
       }
 
+      if (operator === 'nbet') {
+        return `${fieldStr}=le=(${filterVals[0]});${fieldStr}=ge=(${filterVals[1]})`;
```
jkusa, line 82: Should this be:
```suggestion
        return `${fieldStr}=lt=(${filterVals[0]}),${fieldStr}=gt=(${filterVals[1]})`;
```

mattgraham1995, line 82: I think this should be `lt` and `gt`. I believe no rows that match values within the supplied interval should be included

mattgraham1995: Oops, yeah what Jon said

jkusa: Don't we want an `OR`?

holderb5: Yup you're correct, would that just be || in this case?

jkusa: The suggestion above is the correct syntax ^^

jkusa (approved):

mattgraham1995 (approved):

PR #1067, opened 2020-10-13
Response methods

jkusa:
## Description

Add `navi-fact-response` model with min/max time dimension methods for handling non-static date filters

## Proposed Changes

- Refactor column instance types
- Add `navi-fact-response` model 
- Adds min/max time dimension methods
- Adds `getCanonicalName` to base column metadata model
- Renames navi fact model test file name
- Use NaviFactResponse in NaviFacts model

## License

### DIFF
--- a/packages/data/addon/models/navi-fact-response.ts
+++ b/packages/data/addon/models/navi-fact-response.ts
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+
+import EmberObject from '@ember/object';
+import { ResponseV1 } from 'navi-data/serializers/facts/interface';
+import moment, { Moment, MomentInput } from 'moment';
+import { TimeDimensionColumn } from './metadata/time-dimension';
+
+function notNull<T>(t: T | null): t is T {
+  return t !== null;
+}
+
+export default class NaviFactResponse extends EmberObject implements ResponseV1 {
+  rows!: ResponseV1['rows'];
+  meta!: ResponseV1['meta'];
+
+  private getTimeDimensionAsMoments(column: TimeDimensionColumn): Moment[] {
+    const { columnMetadata, parameters } = column;
+    const field = columnMetadata.getCanonicalName(parameters);
+    const { rows = [] } = this;
+    return rows
+      .map(row => {
+        const value = row[field];
+        return value ? moment(value as MomentInput) : null;
+      })
+      .filter(notNull);
+  }
+
+  /**
+   * Get the max dateTime value for a column in moment form
+   */
+  getMaxTimeDimension(column: TimeDimensionColumn): Moment | null {
+    const moments = this.getTimeDimensionAsMoments(column);
+    return moments.length ? moment.max(moments) : null;
+  }
+
+  /**
+   * Get the min dateTime value for a column in moment form
+   */
+  getMinTimeDimension(column: TimeDimensionColumn): Moment | null {
+    const moments = this.getTimeDimensionAsMoments(column);
+    return moments.length ? moment.min(moments) : null;
+  }
+}
--- a/packages/data/addon/models/navi-fact-response.ts
+++ b/packages/data/addon/models/navi-fact-response.ts
@@ -13,8 +13,8 @@ function notNull<T>(t: T | null): t is T {
 }
 
 export default class NaviFactResponse extends EmberObject implements ResponseV1 {
-  rows!: ResponseV1['rows'];
-  meta!: ResponseV1['meta'];
+  readonly rows: ResponseV1['rows'] = [];
+  readonly meta: ResponseV1['meta'] = {};
 
   private getTimeDimensionAsMoments(column: TimeDimensionColumn): Moment[] {
     const { columnMetadata, parameters } = column;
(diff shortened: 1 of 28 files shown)


packages/data/tests/unit/models/navi-fact-response-test.ts
```diff
@@ -0,0 +1,134 @@
...
+    assert.equal(null, max, '`getMaxTimeDimension` returns null for missing values');
+
+    const min = response.getMinTimeDimension({ columnMetadata });
+    assert.equal(null, min, '`getMinTimeDimension` returns null for missing values');
+  });
+
+  test('getMaxTimeDimension/getMaxTimeDimension - value gaps', function(this: TestContext, assert) {
+    const rows = [
+      { 'table1.eventTimeDay': null },
+      { 'table1.eventTimeDay': '2014-04-03 00:00:00.000' },
```
kevinhinterlong, line 112: Add a test for gaps in the middle

packages/data/addon/models/navi-fact-response.ts
```diff
@@ -0,0 +1,46 @@
...
+import moment, { Moment, MomentInput } from 'moment';
+import { TimeDimensionColumn } from './metadata/time-dimension';
+
+function notNull<T>(t: T | null): t is T {
+  return t !== null;
+}
+
+export default class NaviFactResponse extends EmberObject implements ResponseV1 {
+  rows!: ResponseV1['rows'];
+  meta!: ResponseV1['meta'];
```
kevinhinterlong, line 17: Maybe we can do something like this
```suggestion
  rows: ResponseV1['rows'] = [];
  meta: ResponseV1['meta'] = {};
```

packages/data/addon/models/navi-fact-response.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  private getTimeDimensionAsMoments(column: TimeDimensionColumn): Moment[] {
+    const { columnMetadata, parameters } = column;
+    const field = columnMetadata.getCanonicalName(parameters);
+    const { rows = [] } = this;
+    return rows
+      .map(row => {
+        const value = row[field];
+        return value ? moment(value as MomentInput) : null;
```
kevinhinterlong, line 26:
```suggestion
        if (value) {
          const dateTime = moment(value as MomentInput);
          return dateTime.isValid() ? dateTime : null;
        }
        return null;
```

Playing around with `moment` I saw that

> ```
> moment.max(moment('abc'), moment()).format() === "Invalid Date"
> ```
 
so we should probably guard against that and add a test for it

packages/data/addon/models/navi-fact-response.ts
```diff
@@ -0,0 +1,46 @@
...
+      })
+      .filter(notNull);
+  }
+
+  /**
+   * Get the max dateTime value for a column in moment form
+   */
+  getMaxTimeDimension(column: TimeDimensionColumn): Moment | null {
```
kevinhinterlong, line 34: I think it would be useful to have a `getIntervalForTimeDimension(column: TimeDimensionColumn): Interval | null` since interval is now available in data

jkusa: Agreed, I wanted to get these primitives in first.

packages/data/addon/models/navi-fact-response.ts
jkusa: yeah, I guess we are doing this anyways in the normalizers.

packages/data/addon/models/navi-fact-response.ts
jkusa: Yeah, I tested that too. Wasn't sure how likely that was to happen and wanted to keep the checks to a minimum. `null`, `undefined`, `''` are definitely possible.

jkusa: > From version 2.10.5, if an invalid moment is one of the arguments, the result is an invalid moment.

packages/data/addon/models/navi-fact-response.ts
kevinhinterlong: 👍 can you make an issue to track it

jkusa: https://github.com/yahoo/navi/issues/1069

packages/data/addon/serializers/facts/interface.ts
```diff
@@ -10,8 +10,8 @@ import NaviFactResponse from 'navi-data/models/navi-fact-response';
 type ResponsePayload = any;
 
 export interface ResponseV1 {
```
kevinhinterlong, line 12: All readonly?
```suggestion
export interface ResponseV1 ReadOnly<{
```

mattgraham1995 (approved):

kevinhinterlong (approved):
I wasn't able to suggest the ending `>` for readonly and it's actually just `Readonly<T>`, but other than that looks good

packages/data/tests/unit/models/navi-fact-response-test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const minResponse = NaviFactResponse.create({ rows: minRows });
+
+    const min = minResponse.getMinTimeDimension({ columnMetadata });
+    assert.ok(
+      moment(minRows[2]['table1.eventTimeDay']).isSame(min),
+      '`getMinTimeDimension` returns min dateTime when date values have gaps'
+    );
+  });
+
+  test('getMaxTimeDimension/getMaxTimeDimension - invalid values', function(this: TestContext, assert) {
```
kevinhinterlong, line 142:
```suggestion
  test('getMaxTimeDimension/getMinTimeDimension - invalid values', function(this: TestContext, assert) {
```

jkusa:
`Readonly<T>` is not supported on `interface`s, you can do it with `type`s`. I reverted to the field level for now.

PR #1076, opened 2020-10-16
Table export

ayeswarya:
Resolves #

## Description
 UrlfordownloadQuery implementation with TableExport for Async API
## Proposed Changes

-

## License


packages/data/addon/adapters/facts/elide.ts
```diff
@@ -156,9 +193,31 @@ export default class ElideFactsAdapter extends EmberObject implements NaviFactAd
    * @param _request
    * @param _options
    */
-  async urlForDownloadQuery(_request: RequestV1, _options: RequestOptions): Promise<string> {
-    return 'TODO';
+  async urlForDownloadQuery(request: RequestV1, options: RequestOptions = {}): Promise<string> {
+    let response = await this.fetchDataForExportTask.perform(request, options);
+    let url = response.tableExport.edges[0].node.result?.url || '';
```
mattgraham1995, line 198: If there is an error in the request, I believe `edges` could be an empty array so we should account for that here.

mattgraham1995 (commented):
Do we have any tests for a case where we get an error in response from the API?

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -426,6 +427,59 @@ module('Unit | Adapter | facts/elide', function(hooks) {
...
+            {
+              node: {
+                id: requestObj.variables.ids[0],
+                query: 'foo',
+                queryType: 'GRAPHQL_V1_0',
+                status: 'COMPLETE',
+                result: {
+                  httpStatus: '200',
+                  contentLength: 2,
+                  responseBody: JSON.stringify({
```
mattgraham1995, line 457: Is this the right shape for a TableExportResponse?

Looks like it should be this shape:
https://github.com/yahoo/navi/pull/1076/files#diff-4c79f1d6c08140eccb2c41d95c498d9dca6ce3383b90127425ba179c7c0b66c9R141

mattgraham1995 (commented):
When is `urlForDownloadQuery` going to be used versus `fetchDataForExportTask`?

ayeswarya:
> When is `urlForDownloadQuery` going to be used versus `fetchDataForExportTask`?

According to my understanding urlForDownloadQuery is the external facing function which is also defined in the interface. fetchDataForExportTask will only be called from within urlForDownloadQuery.

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -556,14 +556,17 @@ module('Unit | Adapter | facts/elide', function(hooks) {
...
+    Server.post(`${HOST}/graphql`, () => [400, { 'Content-Type': 'application/json' }, JSON.stringify({ response })]);
 
     try {
       await adapter.fetchDataForRequest(TestRequest);
     } catch ({ errors }) {
-      const responseText = await errors[0].statusText;
-      assert.deepEqual(responseText, errors[0].messages, 'fetchDataForRequest an array of response objects on error');
+      assert.deepEqual(
+        response,
+        errors[0].result.response,
```
mattgraham1995, line 567: These two. lines should be swapped as the first argument is the actual value and the second argument is the expected value.

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -654,16 +657,15 @@ module('Unit | Adapter | facts/elide', function(hooks) {
...
 
     try {
       await adapter.urlForDownloadQuery(TestRequest, {});
     } catch ({ errors }) {
-      const responseText = await errors[0].statusText;
       assert.deepEqual(
-        responseText,
-        errors[0].messages,
+        response,
+        errors[0].result.response,
```
mattgraham1995, line 668: Same here

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -672,15 +674,17 @@ module('Unit | Adapter | facts/elide', function(hooks) {
...
+    Server.post(`${HOST}/graphql`, () => [200, { 'Content-Type': 'application/json' }, JSON.stringify({ response })]);
 
     try {
       await adapter.urlForDownloadQuery(TestRequest, {});
     } catch ({ errors }) {
-      const responseText = await errors[0].statusText;
-      assert.deepEqual(responseText, errors[0].messages, 'fetchDataForRequest an array of response objects on error');
+      assert.deepEqual(
+        response,
+        errors[0].result.response,
```
mattgraham1995, line 685: +1

mattgraham1995 (approved):

packages/data/addon/adapters/facts/elide.ts
```diff
@@ -156,9 +193,31 @@ export default class ElideFactsAdapter extends EmberObject implements NaviFactAd
    * @param _request
    * @param _options
    */
-  async urlForDownloadQuery(_request: RequestV1, _options: RequestOptions): Promise<string> {
-    return 'TODO';
+  async urlForDownloadQuery(request: RequestV1, options: RequestOptions = {}): Promise<string> {
+    let response = await this.fetchDataForExportTask.perform(request, options);
+    let url = response.tableExport.edges[0]?.node.result?.url || '';
```
jkusa, line 198:
```suggestion
    const response = await this.fetchDataForExportTask.perform(request, options);
    const url = response.tableExport.edges[0]?.node.result?.url;
```

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -568,4 +613,78 @@ module('Unit | Adapter | facts/elide', function(hooks) {
...
+        'urlForDownloadQuery returns an array of response objects on server error'
+      );
+    }
+  });
+
+  test('urlForDownloadQuery bad request - error', async function(assert) {
+    assert.expect(1);
+    const adapter: ElideFactsAdapter = this.owner.lookup('adapter:facts/elide');
+    const response = { message: 'Bad request' };
+    Server.post(`${HOST}/graphql`, () => [200, { 'Content-Type': 'application/json' }, JSON.stringify({ response })]);
```
jkusa, line 678: How are we determining that this is an error?

PR #1078, opened 2020-10-21
Fix: Use appropriate filter operator for time dimensions

mattgraham1995:
## Description
We can't currently filter on certain time dimensions on request-v2 because their metadataType is set to `TIME` rather than the expected `date`.

## Proposed Changes

- support a `time` metadataType as well for filter adds

## License

### DIFF
--- a/packages/reports/addon/consumers/request/filter.ts
+++ b/packages/reports/addon/consumers/request/filter.ts
@@ -117,7 +117,9 @@ export default class FilterConsumer extends ActionConsumer {
       const { request } = route.modelFor(routeName) as ReportModel;
 
       const findDefaultOperator = (type: string) => {
+        type = type.toLowerCase();
         const opDictionary: Record<string, string> = {
+          time: 'gte',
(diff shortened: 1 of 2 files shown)


kevinhinterlong (approved):
Can we also fix the `TOGGLE_FILTER` action? I think that should fix the column config filter button (can be done separately)

packages/reports/addon/consumers/request/filter.ts
```diff
@@ -117,7 +117,9 @@ export default class FilterConsumer extends ActionConsumer {
       const { request } = route.modelFor(routeName) as ReportModel;
 
       const findDefaultOperator = (type: string) => {
+        type = type.toLowerCase();
         const opDictionary: Record<string, string> = {
```
jkusa, line 121: It would be nice to type this record with `valueType`s and filter operator types

jkusa (approved):
Just the one comment, but not a blocker

PR #1091, opened 2020-10-28
Elide dimensions

jkusa:
#625  Description

Plumbing work to allow for elide dimension type ahead features

## Proposed Changes

- Add insert method to keg
- Add metadata fields to elide dimensions
- Metadata serializers to provide models

## License


packages/data/addon/services/keg.ts
```diff
@@ -85,6 +91,51 @@ export default class KegService extends Service {
     return this.pushMany(type, [rawRecord], options).firstObject;
   }
 
+  /**
+   * Inserts a single record into the store
```
kevinhinterlong, line 95: Maybe we can update the `pushMany` method to `create` and then call `insertMany` to make the difference more clear

packages/data/addon/services/keg.ts
```diff
@@ -85,6 +91,51 @@ export default class KegService extends Service {
...
+   *
+   * @param type - type name of the model type
+   * @param record - record to be inserted into the keg
+   * @param options - config object
+   */
+  insertMany(type: string, records: Array<KegRecord>, options: Partial<InsertOptions> = {}): EmberArray<KegRecord> {
+    const recordKeg = this._getRecordKegForType(type);
+    const idIndex = this._getIdIndexForType(type);
+    const namespace = options.namespace || this.defaultNamespace;
+    const identifierField = options?.identifierField || KegService.identifierField;
```
mattgraham1995, line 116:
```suggestion
    const identifierField = options.identifierField || KegService.identifierField;
```

packages/data/addon/services/keg.ts
```diff
@@ -85,6 +91,51 @@ export default class KegService extends Service {
...
+  }
+
+  /**
+   * Inserts an array of records into the store
+   *
+   * @param type - type name of the model type
+   * @param record - record to be inserted into the keg
+   * @param options - config object
+   */
+  insertMany(type: string, records: Array<KegRecord>, options: Partial<InsertOptions> = {}): EmberArray<KegRecord> {
```
mattgraham1995, line 112: Is pushMany being deprecated or is there a significant difference between this and that?

kevinhinterlong (approved):

packages/data/tests/unit/services/keg-test.js
```diff
@@ -113,13 +105,19 @@ module('Unit | Service | keg', function(hooks) {
...
+    assert.equal(Record1, foundRecord, 'after inserting a record it can be found');
+  });
+
+  test('pushing a record with an existing id update the record in the keg', function(assert) {
+    //Insert initial record
+    const pushedRecord = Keg.push('record', RawRecord1);
 
-    //Push new record with same id
+    //Insert new record with same id
     Keg.push('record', { id: 1, description: 'updated' });
```
mattgraham1995, line 121: Should this be using `insert` or `push`?

packages/data/tests/unit/services/keg-test.js
```diff
@@ -139,26 +137,39 @@ module('Unit | Service | keg', function(hooks) {
...
+    const newRecord = RecordFactory.create({ ...RawRecord1, description: 'updated' });
+    Keg.insert('record', newRecord);
+
+    let fetchedRecord = Keg.getById('record', 1);
 
-    let pushedRecord = Keg.push('notRecord', Record1, {
+    assert.deepEqual(
+      Keg.all('record'),
+      [newRecord],
+      'Pushing a record into the keg with an existing id does not add a new record'
```
mattgraham1995, line 153:
```suggestion
      'Inserting a record into the keg with an existing id does not add a new record'
```

mattgraham1995 (commented):
`insert` vs `push` seems a little confusing

packages/data/addon/services/keg.ts
jkusa: `options` is defaulted to `{}`

packages/data/addon/services/keg.ts
jkusa: `insert` does not populate a model (`model.create(payload)`, that is the main diff. But we will deprecated push methods at some point. Maybe when we remove keg as a service?

packages/data/addon/services/keg.ts
mattgraham1995: right that's why the optional chaining isn't needed, right? Or if it is needed, it should be added to the line above where we access `options.namespace`

jkusa: ah, read your comment the other way.

packages/data/tests/unit/services/keg-test.js
```diff
@@ -113,13 +105,19 @@ module('Unit | Service | keg', function(hooks) {
...
+    const foundRecord = Keg.getById('record', 1, 'navi');
+    assert.equal(Record1, foundRecord, 'after inserting a record it can be found');
+  });
+
+  test('pushing a record with an existing id update the record in the keg', function(assert) {
+    //Insert initial record
+    const pushedRecord = Keg.push('record', RawRecord1);
 
-    //Push new record with same id
+    //Insert new record with same id
```
jkusa, line 120:
```suggestion
    //Push new record with same id
```

mattgraham1995 (approved):

PR #1109, opened 2020-11-06
Update: Add fili consumer to keep filter in sync with dateTime column

### LINKED ISSUE #1000: Add Fili Consumer
- Keep filter grain and column sync

kevinhinterlong:
Resolves #1000 

## Description
With the new report view it's challenging to change the parameters of a filter and would be extra work to change a parameter in both the column and filter

## Proposed Changes
- Adds a fili specific consumer which will forward updates to the dateTime column to the dateTime filter so the grains are always in sync

## License

### DIFF
--- a/packages/reports/addon/consumers/request/fili.ts
+++ b/packages/reports/addon/consumers/request/fili.ts
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import { inject as service } from '@ember/service';
+import ActionConsumer from 'navi-core/consumers/action-consumer';
+import Route from '@ember/routing/route';
+import RequestActionDispatcher, { RequestActions } from 'navi-reports/services/request-action-dispatcher';
+import ColumnFragment from 'navi-core/models/bard-request-v2/fragments/column';
+import ReportModel from 'navi-core/models/report';
+import { getDataSource } from 'navi-data/utils/adapter';
+
+export default class FiliConsumer extends ActionConsumer {
+  @service requestActionDispatcher!: RequestActionDispatcher;
+
+  /**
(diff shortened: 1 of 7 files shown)


packages/reports/addon/consumers/request/fili.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Filters actions to only those updating fili requests
+   */
+  shouldConsumeAction(_actionName: string, route: Route) {
+    const { routeName } = route;
+    const { request } = route.modelFor(routeName) as ReportModel;
+    const { type } = getDataSource(request.dataSource);
+
+    return type === 'bard';
```
jkusa, line 24: 👍

packages/reports/tests/unit/consumers/request/fili-test.ts
```diff
@@ -0,0 +1,159 @@
...
+    consumer.send(
+      RequestActions.UPDATE_COLUMN_FRAGMENT_WITH_PARAMS,
+      { modelFor },
+      request.timeGrainColumn,
+      'grain',
+      'week'
+    );
+    assert.deepEqual(dispatchedActions, [], 'no actions are called when the request is for an elide dataSource');
+  });
+});
```
jkusa, line 159: Do we already have acceptance tests that verify this behavior?

kevinhinterlong: I don't think we load any elide data sources into the reports package

jkusa: Should we update this test name to include `bard/fili`

https://github.com/yahoo/navi/blob/59189827f8c06a72166df1c1b38671b03d7fbc09/packages/reports/tests/acceptance/date-filter-test.js#L70

jkusa (approved):
Just the one comment, othewise 👍

jkusa (approved):

PR #1112, opened 2020-11-09
Update: Add demo data source

jkusa:
## Description

Add demo data source

## Proposed Changes

- Add demo data source
- Updated to elide 5.0.0-pr24

## License


aklish (approved):

packages/webservice/app/src/main/resources/create-demo-data.sql
```diff
@@ -0,0 +1,63 @@
+-- PILOTS --
+
+CREATE TABLE IF NOT EXISTS pilots
+    (
+      id INT,
+      name VARCHAR(255),
+    );
+TRUNCATE TABLE pilots;
+
+INSERT INTO pilots VALUES (1, 'Luke S');
```
rishi-aga, line 10: you might need to write INSERTs like `INSERT INTO pilots SELECT 1, 'Luke S' WHERE NOT EXISTS(SELECT * FROM pilots WHERE id = 1 and name = 'Luke S');` to avoid multiple inserts for same row as this script will be called for each connection initialization.

jkusa: @rishi-aga good call out. We do `TRUNCATE TABLE` before inserting, however this could get hairy with many connections. Ultimately, this will be converted to a liquibase script.

PR #1114, opened 2020-11-10
initial e2e - liquibase plus int test

anupkumangodan:
Resolves #

## Use liquibase for starwars/demo model

Please note that this is not expected to build since liquibase task in gradle is failing. I am pushing this commit so someone can take a look.


packages/webservice/app/src/test/kotlin/com/yahoo/navi/ws/test/integration/StarWarsTest.kt
```diff
@@ -0,0 +1,20 @@
+package com.yahoo.navi.ws.test.integration
+
+import com.jayway.restassured.RestAssured.given
+import com.yahoo.navi.ws.test.framework.IntegrationTest
+import org.hamcrest.Matchers
+import org.junit.jupiter.api.Test
+
+class StarWarsTest : IntegrationTest() {
```
jkusa, line 8: Can we rename this file to `DemoDataSourceTest.kt`

packages/webservice/app/src/main/resources/demo-configs/db/sql/DemoConnection.hjson
```diff
@@ -2,7 +2,7 @@
   dbconfigs: [
     {
       name: DemoConnection
-      url: jdbc:h2:mem:SalesDB;DB_CLOSE_DELAY=-1;INIT=RUNSCRIPT FROM 'classpath:create-demo-data.sql'
+      url: jdbc:h2:file:./SalesDB;DB_CLOSE_DELAY=-1
```
jkusa, line 5: Can we not use an in memory db? Also can we change the name?

anupkumangodan: yea name can be demoDB better

anupkumangodan: I can change the name but using memory will need more work. I feel we could do with file based since its internally used and the actual management is through liquibase changelog.

packages/webservice/app/src/test/kotlin/com/yahoo/navi/ws/test/integration/StarWarsTest.kt
anupkumangodan: changed.

jkusa:
Closing for now as we are moving away from the inital demo data.

PR #1117, opened 2020-11-10
Update: Only request fact tables from elide

jkusa:
## Description

Only request fact tables from elide

## Proposed Changes

- Only request fact tables from elide

## License

### DIFF
--- a/packages/data/package.json
+++ b/packages/data/package.json
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@miragejs/graphql": "^0.1.8",
+    "@rsql/parser": "^1.0.2",
     "apollo-link-context": "^1.0.19",
     "broccoli-funnel": "^3.0.1",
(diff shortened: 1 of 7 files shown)


packages/data/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@miragejs/graphql": "^0.1.8",
+    "@rsql/parser": "^1.0.2",
```
kevinhinterlong, line 32: We could probably add something similar in the build for this
https://github.com/yahoo/navi/blob/master/packages/data/index.js

kevinhinterlong (approved):

packages/data/package.json
jkusa: Are you talking about removing it from the prod build? Since we remove the mirage directory for prod builds, auto import never brings it in. I validated this prior to the PR.

PR #1119, opened 2020-11-13
Support non time dim visualizations

jkusa:
## Description

Support non time dim visualizations

## Proposed Changes

- Support non time dim visualizations
- Fix render pie chart label on alias update
- Fix Pie undefined seriesConfig (teardown) case

## License

### DIFF
--- a/packages/core/addon/chart-builders/metric.ts
+++ b/packages/core/addon/chart-builders/metric.ts
@@ -24,6 +24,9 @@ import { BaseChartBuilder, C3Row, EmptyC3Data } from './base';
 import { tracked } from '@glimmer/tracking';
 import { MetricSeries } from 'navi-core/models/chart-visualization';
 import NaviFactResponse, { ResponseRow } from 'navi-data/models/navi-fact-response';
+import ColumnFragment from 'navi-core/models/bard-request-v2/fragments/column';
+
+const BLANK_X_VALUE = '';
 
 export default class MetricChartBuilder extends EmberObject implements BaseChartBuilder {
   @tracked byXSeries?: DataGroup<ResponseRow>;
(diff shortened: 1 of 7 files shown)


packages/core/addon/chart-builders/metric.ts
```diff
@@ -24,6 +24,9 @@ import { BaseChartBuilder, C3Row, EmptyC3Data } from './base';
 import { tracked } from '@glimmer/tracking';
 import { MetricSeries } from 'navi-core/models/chart-visualization';
 import NaviFactResponse, { ResponseRow } from 'navi-data/models/navi-fact-response';
+import ColumnFragment from 'navi-core/models/bard-request-v2/fragments/column';
+
+const BLANK_X_VALUE = '';
```
kevinhinterlong, line 29: Could probably put this in the base one and use it for the raw value of `EmptyC3Data`

packages/core/addon/chart-builders/metric.ts
```diff
@@ -48,59 +54,79 @@ export default class MetricChartBuilder extends EmberObject implements BaseChart
...
-     * Get all date buckets spanned by the data,
-     * and group data by date for easier lookup
-     */
-    const dates = interval.makeEndExclusiveFor(timeGrain).getDatesForInterval(timeGrain);
-    const byDate = new DataGroup(response.rows, (row: ResponseRow) =>
-      buildDateKey(row[timeGrainColumn] as MomentInput)
-    );
+    const { timeGrain, metricColumns } = request;
 
     const names = request.metricColumns.reduce((names: Record<string, string>, metric, index) => {
```
kevinhinterlong, line 63:
```suggestion
    const names = metricColumns.reduce((names: Record<string, string>, metric, index) => {
```

kevinhinterlong (approved):

PR #1121, opened 2020-11-13
Initial Test Fixes

holderb5:
## Description
Initial changes a few more modules tests fully passing.
Navi-Report testing module is closer to passing
## Proposed Changes

-

## License

### DIFF
--- a/packages/core/addon/mirage/fixtures/reports.js
+++ b/packages/core/addon/mirage/fixtures/reports.js
@@ -26,7 +26,8 @@ export default [
                 dimensions: [
                   { name: 'Property 1', values: { property: '114' } },
                   { name: 'Property 2', values: { property: '100001' } },
-                  { name: 'Property 3', values: { property: '100002' } }
+                  { name: 'Property 3', values: { property: '100002' } },
+                  { name: 'Property 4', values: { property: '101272' } }
                 ]
               }
             }
@@ -221,7 +222,29 @@ export default [
     updatedOn: '2015-07-01 00:00:00',
     authorId: 'navi_user',
     deliveryRuleIds: [],
-    visualization: null,
+    visualization: {
+      type: 'table',
+      version: 1,
+      metadata: {
+        columns: [
+          {
+            field: 'dateTime',
+            type: 'dateTime',
+            displayName: 'Date'
+          },
+          {
+            field: 'adClicks',
+            type: 'metric',
+            displayName: 'Ad Clicks'
+          },
+          {
+            field: 'navClicks',
+            type: 'metric',
+            displayName: 'Nav Clicks'
+          }
+        ]
+      }
+    },
     request: {
       logicalTable: {
         table: 'network',
@@ -250,22 +273,13 @@ export default [
     deliveryRuleIds: [],
     visualization: null,
     request: {
-      logicalTable: {
-        table: 'network',
-        timeGrain: 'day'
-      },
-      metrics: [],
-      dimensions: [],
+      table: '',
+      dataSource: '',
+      requestVersion: '2.0',
+      limit: null,
+      columns: [],
       filters: [],
-      intervals: [
-        {
-          end: '2015-11-16 00:00:00.000',
-          start: '2015-11-09 00:00:00.000'
-        }
-      ],
-      having: [{ metric: 'adClicks', operator: 'gt', values: [] }],
-      bardVersion: 'v1',
-      requestVersion: 'v1'
+      sorts: []
     }
   },
   {
@@ -614,6 +628,76 @@ export default [
       bardVersion: 'v1',
       requestVersion: 'v1',
       dataSource: 'bardTwo'
+    }
+  },
+  {
+    id: 13,
+    title: 'RequestV2 testing report',
+    createdOn: '2015-04-01 00:00:00',
+    updatedOn: '2015-04-01 00:00:00',
+    authorId: 'navi_user',
+    deliveryRuleIds: [],
+    visualization: {
+      type: 'table',
+      version: 1,
+      metadata: {
+        columns: [
+          {
+            field: 'dateTime',
+            type: 'dateTime',
+            displayName: 'Date'
+          },
+          {
+            field: 'property',
+            type: 'dimension',
+            displayName: 'Property'
+          },
+          {
+            field: 'adClicks',
+            type: 'metric',
+            displayName: 'Ad Clicks'
+          }
+        ]
+      }
+    },
+    request: {
+      table: 'network',
+      dataSource: 'bardOne',
+      limit: null,
+      requestVersion: '2.0',
+      filters: [
+        {
+          type: 'timeDimension',
+          dataSource: 'bardOne',
+          field: 'network.dateTime',
+          parameters: { grain: 'day' },
+          operator: 'bet',
+          values: []
+        }
+      ],
+      columns: [
+        {
+          cid: 'c1',
+          field: 'network.dateTime',
+          parameters: {
+            grain: 'day'
+          },
+          type: 'timeDimension'
+        },
+        {
+          cid: 'c2',
+          type: 'metric',
+          field: 'adClicks',
+          parameters: {}
+        },
+        {
+          cid: 'c3',
+          type: 'dimension',
+          field: 'property',
+          parameters: { field: 'id' }
+        }
+      ],
+      sorts: []
     }
   }
 ];
--- a/packages/core/addon/mirage/fixtures/reports.js
+++ b/packages/core/addon/mirage/fixtures/reports.js
@@ -639,25 +639,14 @@ export default [
     deliveryRuleIds: [],
     visualization: {
       type: 'table',
-      version: 1,
+      version: 2,
       metadata: {
-        columns: [
-          {
-            field: 'dateTime',
-            type: 'dateTime',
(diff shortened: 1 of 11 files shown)


kevinhinterlong:
https://cd.screwdriver.cd/pipelines/6102/builds/641199/steps/test-navi-reports
Navi reports silently failed linting

packages/core/tests/integration/components/chart-series-config-test.ts
```diff
@@ -53,7 +53,7 @@ module('Integration | Component | chart series config', function(hooks) {
 
     assert.deepEqual(
       findAll('.line-chart-config__series-config__item__content').map(el => el.textContent?.trim()),
-      ['Property 1', 'Property 2', 'Property 3'],
+      ['Property 1', 'Property 2', 'Property 3', 'Property 4'],
```
jkusa, line 56: Is the extra dimension/series necessary?

packages/core/addon/mirage/fixtures/reports.js
```diff
@@ -26,7 +26,8 @@ export default [
                 dimensions: [
                   { name: 'Property 1', values: { property: '114' } },
                   { name: 'Property 2', values: { property: '100001' } },
-                  { name: 'Property 3', values: { property: '100002' } }
+                  { name: 'Property 3', values: { property: '100002' } },
+                  { name: 'Property 4', values: { property: '101272' } }
```
jkusa, line 30: Why do we need to add an additional property?

holderb5: When running the tests I found that a Property 4 was showing up regardless of if I had one defined. It showed up with the value of 101272 so I added it and everything worked as expected with it defined properly.

packages/core/tests/integration/components/chart-series-config-test.ts
kevinhinterlong: This is a result of the rework around dimension series charts now always being invalid, so it pulls in the `Property 4` dimension on load since it wasn't filtered out

jkusa (approved):

packages/core/addon/mirage/fixtures/reports.js
```diff
@@ -615,5 +629,75 @@ export default [
...
+  {
+    id: 13,
+    title: 'RequestV2 testing report',
+    createdOn: '2015-04-01 00:00:00',
+    updatedOn: '2015-04-01 00:00:00',
+    authorId: 'navi_user',
+    deliveryRuleIds: [],
+    visualization: {
+      type: 'table',
+      version: 1,
```
jkusa, line 642: Can we make this a v2 with `cid`

packages/reports/tests/acceptance/report-test.js
```diff
@@ -47,20 +47,26 @@ module('Acceptance | Navi Report', function(hooks) {
...
     let errors = findAll('.navi-info-message__error-list-item').map(el => find(el).innerText.trim());
-    assert.deepEqual(errors, ['Operating System filter needs at least one value'], 'Form errors are displayed to user');
+    assert.deepEqual(errors, ['At least one column should be selected'], 'Form errors are displayed to user');
 
+    await clickItem('dimension', 'Date Time');
+    await clickItem('dimension', 'Property');
+    await clickItem('metric', 'Ad Clicks');
+    await clickItemFilter('dimension', 'Operating System');
     // Fix the errors and run report
     await click('.filter-collection__row:nth-of-type(2) .filter-collection__remove');
```
kevinhinterlong, line 66: I think here we are adding and then just deleting a filter so we can remove it
```suggestion
```

packages/reports/tests/acceptance/report-test.js
```diff
@@ -103,6 +110,16 @@ module('Acceptance | Navi Report', function(hooks) {
...
     await clickItemFilter('dimension', 'Operating System');
     await clickItem('metric', 'Ad Clicks');
+    await clickItem('dimension', 'Date Time');
+    await clickItemFilter('dimension', 'Date Time');
+    await click($('.filter-builder__operator:contains(Since) .filter-builder__select-trigger')[0]);
+    await click($('li.ember-power-select-option:contains(Between)')[0]);
+
+    await clickTrigger('.filter-values--dimension-date-range-input__low-value .ember-basic-dropdown-trigger');
+
+    await click('.ember-power-calendar-day[data-date="2020-11-05"]');
```
kevinhinterlong, line 120: This will probably fail when we're no longer in november so maybe try to make these values dynamic

packages/reports/tests/acceptance/report-test.js
```diff
@@ -121,6 +138,19 @@ module('Acceptance | Navi Report', function(hooks) {
...
+
+    //add date-dimension
+    await clickItem('dimension', 'Date Time');
+    await clickItemFilter('dimension', 'Date Time');
+
+    //set the filter interval
+    await click($('.filter-builder__operator:contains(Since) .filter-builder__select-trigger')[0]);
+    await click($('li.ember-power-select-option:contains(Between)')[0]);
+    await clickTrigger('.filter-values--dimension-date-range-input__low-value .ember-basic-dropdown-trigger');
+    await click('.ember-power-calendar-day[data-date="2020-11-05"]');
```
kevinhinterlong, line 150: same here

kevinhinterlong (commented):
small stuff

packages/reports/tests/acceptance/report-test.js
holderb5: Do we have any examples of this? would we just pick the first and last date of the calendar and how would we do that?

kevinhinterlong: We have this one where we pass in a moment and get an element selector for that date. So you could do something like `moment().startOf('month').add(4, 'days')` to `moment().startOf('month').add(5, 'days')`
https://github.com/yahoo/navi/blob/6a80ca7d51850f06b4d5a12b94538c1300f9a18f/packages/reports/tests/integration/components/navi-date-picker-test.js#L379-L383

or roughly the same version here

https://github.com/yahoo/navi/blob/9f2b5edede2134a93f16f307bdffaaee46ba5434/packages/reports/tests/acceptance/date-filter-test.js#L45-L50

holderb5: I chose the second version cause thats what I found initially in our code. Its pushed now

packages/reports/tests/acceptance/report-test.js
```diff
@@ -1061,6 +1105,16 @@ module('Acceptance | Navi Report', function(hooks) {
     await visit('/reports');
     await visit('/reports/new');
     await clickItem('metric', 'Ad Clicks');
+    await clickItem('dimension', 'Date Time');
+    await clickItemFilter('dimension', 'Date Time');
+    await click($('.filter-builder__operator:contains(Since) .filter-builder__select-trigger')[0]);
+    await click($('li.ember-power-select-option:contains(Between)')[0]);
```
kevinhinterlong, line 1113: Use this going forward
```suggestion
    await selectChoose('.filter-builder__select-trigger', 'Between');
```

packages/reports/tests/acceptance/report-test.js
```diff
@@ -76,12 +79,18 @@ module('Acceptance | Navi Report', function(hooks) {
...
+    await click('.navi-column-config-item__remove-icon[aria-label="delete dimension Property (id)"]');
+    await click('.navi-column-config-item__remove-icon[aria-label="delete time-dimension Date Time (day)"]');
     await click($('.navi-report__action-link:contains(Clone)')[0]);
 
-    assert.ok(currentURL().endsWith('edit'), 'An invalid new report transitions to the reports/:id/edit route');
+    //fails because of server error and stays at /reports/13/edit
+    assert.ok(
+      currentURL().endsWith('/edit'),
+      'After clicking the run button, the route transitions to the invalid route'
+    );
```
kevinhinterlong, line 93: This should move the user over to a new report, not stay at `/13/edit`, can we assert that it moves the user to a new report. Maybe assert the url does not include `/reports/13`?

holderb5: It moves to a new report edit, is that whats intended?

kevinhinterlong: Yea, the comment `//fails because of server error and stays at /reports/13/edit` indicated otherwise

kevinhinterlong: We should remove the comment^ and add an assert that the url does not contain `/reports/13`

kevinhinterlong (commented):
In the [last commit](https://github.com/yahoo/navi/pull/1121/commits/d3e94df460a52dd9000767f5111528a6d937f3d1) you can also remove all of the `await click($('.filter-builder__operator:contains(Since) .filter-builder__select-trigger')[0]);` calls

packages/reports/tests/acceptance/report-test.js
```diff
@@ -77,8 +77,6 @@ module('Acceptance | Navi Report', function(hooks) {
   });
 
   test('Clone invalid report', async function(assert) {
-    assert.expect(1);
```
kevinhinterlong, line 77: We typically leave in the assert count for async tests

kevinhinterlong (approved):

jkusa (approved):

PR #1130, opened 2020-11-19
Fix missing metric issues

jkusa:
## Description

Currently, the UI breaks when metric label or goal gauge try to render when a metric is missing. This gracefully handles the error, however more will need to be done to indicate to the user that the query needs to be re-executed.

## Proposed Changes

- Handle missing metric in goal gauge
- Handle missing metric in metric label

## License


kevinhinterlong (approved):
👍  could probably also add a test for a metricCid being defined and no column existing for that cid, but let's get this released

PR #1131, opened 2020-11-19
Disable Async Cleanup in Tests

moizarafat:
## Description
Disabling the Async Cleanup in test cases.

## Proposed Changes
A separate profile for `test` sets the flag to false.

## License

### DIFF
--- a/packages/webservice/app/src/main/resources/application.yaml
+++ b/packages/webservice/app/src/main/resources/application.yaml
@@ -47,10 +47,3 @@ server:
 
 logging:
   path: /tmp
-
----
-spring:
-  profiles: test
-  liquibase:
-    contexts: test
-    drop-first: true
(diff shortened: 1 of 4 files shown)


packages/webservice/app/src/main/resources/application.yaml
```diff
@@ -49,6 +49,10 @@ logging:
   path: /tmp
 
 ---
+elide:
+  async:
+    cleanupEnabled: false
```
jkusa, line 54: do we need a profile here?

moizarafat: no, anything below `---` gets tagged with test profile.

packages/webservice/app/src/main/resources/application.yaml
```diff
@@ -49,6 +49,10 @@ logging:
...
+
 spring:
   profiles: test
```
jkusa, line 57:
```suggestion
```

@moizarafat ??

jkusa (approved):

PR #1134, opened 2020-11-20
Updating the Netflix demo model to have a category per Dim/Metrics.

watfagithub:
Resolves #

## Proposed Changes

-

## License

### DIFF
--- a/packages/webservice/app/src/main/resources/demo-configs/models/tables/DemoTables.hjson
+++ b/packages/webservice/app/src/main/resources/demo-configs/models/tables/DemoTables.hjson
@@ -87,13 +87,13 @@
         }
         {
           name: total_seasons
-          category: TV Stats
+          category: Stats
           type: INTEGER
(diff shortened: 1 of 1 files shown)


watfagithub:
@kilroy please review and merge

packages/webservice/app/src/main/resources/demo-configs/models/tables/DemoTables.hjson
```diff
@@ -53,33 +61,39 @@
...
           name: description
+          category: Attributes
           type: TEXT
           definition: '{{description}}'
         }
       ]
       measures: [
         {
           name: count
+          category: Stats
```
jkusa, line 84: Should we have three categories with one metric each or one category with three metrics?

watfagithub: Based on this data and since it is only a demo, it is more clear (For this) to have 3 categories with one metic each, since season only applies to show and duration to movies ... but if you do not feel strong of it, we can have:
Stats -|
           |  (Change metic name to point to Movie or TV)

jkusa (approved):

PR #1133, opened 2020-11-20
Update: Elide time dimension filter value format

jkusa:
## Description

Update: Time dimension filter value format

## Proposed Changes

- Refactor time grains
  - Reorganize time grain and period types
  - Add `isoWeek` as grain
  - Normalize bard/fili `week` grain to `isoWeek`
  - Replace ISODatePeriod util methods
  - Remove get-date-range file
- Elide time dimension filter value format

## License

I confirm that this contribution is made under an MIT license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1136, opened 2020-11-24
Fix Elide `contains` Filter

jkusa:
## Description

Fix Elide `contains` Filter

**Needs** #1133 

## Proposed Changes

- Fix Elide `contains` filter building
- Ignore filters without filter values
- `null` & `notnull` operators now have a value of `true`

## License

### DIFF
--- a/packages/data/addon/adapters/facts/elide.ts
+++ b/packages/data/addon/adapters/facts/elide.ts
@@ -24,7 +24,7 @@ import { v1 } from 'ember-uuid';
 import moment, { Moment } from 'moment';
 import { Grain } from 'navi-data/utils/date';
 
-const escape = (value: string | number) => `${value}`.replace(/'/g, "\\\\'");
+const escape = (value: string) => value.replace(/'/g, "\\\\'");
 
 /**
  * Formats elide request field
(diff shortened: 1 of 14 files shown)


packages/data/addon/adapters/facts/elide.ts
```diff
@@ -71,12 +63,33 @@ export default class ElideFactsAdapter extends EmberObject implements NaviFactAd
...
   }
 
+  private filterBuilders: Record<FilterOperator, (field: string, value: string[]) => string> = {
+    eq: (f, v) => `${f}==('${v[0]}')`,
+    neq: (f, v) => `${f}!=('${v[0]}')`,
+    in: (f, v) => `${f}=in=(${v.map(e => `'${e}'`).join(',')})`,
+    notin: (f, v) => `${f}=out=(${v.map(e => `'${e}'`).join(',')})`,
+    contains: (f, v) => `${f}=in=(${v.map(e => `'*${e}*'`).join(',')})`,
+    null: (f, _v) => `${f}=isnull=true`,
+    notnull: (f, _v) => `${f}=isnull=false`,
```
kevinhinterlong, line 73: I think we should either adopt the single `isnull` type operator within navi with `[true]` or `[false]` as values or allow for values to be an empty array.

jkusa: Agreed. We will need to change a few things for that. Let's come back.

kevinhinterlong (approved):

PR #1137, opened 2020-11-24
More Test Fixes in Report Package

holderb5:
## Description
Down to 46 tests failing in reports, 115 assertions failing
## Proposed Changes

-

## License

### DIFF
--- a/packages/reports/tests/acceptance/multi-datasource-test.js
+++ b/packages/reports/tests/acceptance/multi-datasource-test.js
@@ -19,7 +19,7 @@ module('Acceptance | multi-datasource report builder', function(hooks) {
   });
 
   test('multi datasource report', async function(assert) {
-    assert.expect(13);
+    assert.expect(14);
 
     await visit('/reports/new');
 
@@ -67,10 +67,13 @@ module('Acceptance | multi-datasource report builder', function(hooks) {
     assert
       .dom('.report-builder__container--filters--collapsed')
       .containsText(
-        'Filters Container (id) equals 1 Date Time (day) between (<=>) 11/04/2020 and 11/05/2020 Used Amount greater than (>) 30',
+        'Filters Container (id) equals 1 Date Time (day) between (<=>)',
         'Collapsed filter contains right text'
       );
 
+    assert
+      .dom('.report-builder__container--filters--collapsed')
+      .containsText('Used Amount greater than (>) 30', 'Collapsed filter contains right text');
     //check visualizations are showing up correctly
     assert.deepEqual(
       findAll('.table-widget__table-headers .table-header-cell__title').map(el => el.textContent.trim()),
--- a/packages/reports/tests/acceptance/multi-datasource-test.js
+++ b/packages/reports/tests/acceptance/multi-datasource-test.js
@@ -27,7 +27,7 @@ module('Acceptance | multi-datasource report builder', function(hooks) {
 
     assert.deepEqual(
       findAll('.grouped-list__group-header-content').map(el => el.textContent.trim()),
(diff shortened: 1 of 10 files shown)


packages/reports/tests/integration/components/filter-builders/date-dimension-test.ts
```diff
@@ -57,6 +57,6 @@ module('Integration | Component | filter-builders/date-dimension', function(hook
 
     assert
       .dom('.filter-builder')
-      .hasText(`${this.filter.displayName} since (>=) Sep 08, 2020`, 'Rendered correctly when collapsed');
+      .hasText(`${this.filter.displayName} since (>=) 09/08/2020`, 'Rendered correctly when collapsed');
```
jkusa, line 60: Do we know why this changed?

holderb5: Nope, but across the board all the formatting was coming out in that format rather than the Month Day, year format

jkusa: Before we go forward with this change, we need to understand what is causing the behavior change. @kevinhinterlong any idea here? Will your calendar PR restore the formatting?

kevinhinterlong: Yea the calendar PR will restore this to the original behavior. It was updated to temporarily use `format-date-for-granularity` but it will go back so we could skip this change for now

packages/reports/tests/integration/components/power-select-bulk-import-trigger-test.js
```diff
@@ -19,7 +19,7 @@ module('Integration | Component | power select bulk import trigger', function(ho
       selected: [],
       onChange: () => null,
       extra: {
-        filter: { subject: { id: 'property', name: 'property' } }
+        filter: { subject: { id: 'property', name: 'property', source: 'bardOne' } }
```
kevinhinterlong, line 22: I think the `subject` part is leftover from the old filter builders stuff

kevinhinterlong (commented):
The integration tests for filter-values components aren't necessary or might need a rebase since I'll be updating those in my PR

kevinhinterlong (commented):
You can revert the changes to `filter-values/date-test.ts` and `filter-values/dimension-date-range-test.ts`. Other than that 👍

kevinhinterlong (approved):

packages/reports/tests/acceptance/multi-datasource-test.js
```diff
@@ -67,10 +67,13 @@ module('Acceptance | multi-datasource report builder', function(hooks) {
     assert
       .dom('.report-builder__container--filters--collapsed')
       .containsText(
-        'Filters Container (id) equals 1 Date Time (day) between (<=>) 11/04/2020 and 11/05/2020 Used Amount greater than (>) 30',
+        'Filters Container (id) equals 1 Date Time (day) between (<=>)',
```
jkusa, line 70: Are these dynamic dates? Can we calculate the expected values in the test?

holderb5: The dates will update to be the 4th and 5th of whatever month it is, so if we have a tool to calculate that then we likely could

kevinhinterlong: We can do something like `moment().startOf('month').add(4, 'days').format('MM/DD/YYY')` but it'll change format from my PR

holderb5: yeah thats true, wanna just leave it this way for now and update it when your pr comes through to be more dynamic?

jkusa (approved):

PR #1147, opened 2020-12-04
Fix sort consumer

kevinhinterlong:
## Description
Currently the table visualization allows users to insert or update a sort but not remove them

## Proposed Changes
- Fix the sort consumer so that it properly handles removing a sort

## License


packages/core/addon/components/navi-visualizations/table.ts
```diff
@@ -50,6 +50,7 @@ export type Args = {
 export type TableColumn = {
   fragment: ColumnFragment;
   attributes: TableColumnAttributes;
+  sortDirection: string | undefined;
```
jkusa, line 53:
```suggestion
  sortDirection?: SortDirection;
```
??

packages/reports/addon/consumers/report/table-visualization.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright 2018, Yahoo Holdings Inc.
```
jkusa, line 2:
```suggestion
 * Copyright 2020, Yahoo Holdings Inc.
```

jkusa, line 2:
```suggestion
 * Copyright 2020, Yahoo Holdings Inc.
```

packages/reports/addon/consumers/request/sort.ts
```diff
@@ -0,0 +1,146 @@
...
+        request.removeSort(sort);
+      }
+    },
+
+    /**
+     * @action REMOVE_COLUMN
+     * @param route - route that has a model that contains a request property
+     * @param columnMetadata - metadata model of column to remove
+     */
+    [RequestActions.REMOVE_COLUMN](this: SortConsumer, route: Route, columnMetadata: ColumnMetadataModels) {
```
jkusa, line 68: Do we need this one if we have `REMOVE_COLUMN_WITH_PARAMS`?

packages/reports/addon/consumers/request/sort.ts
```diff
@@ -0,0 +1,146 @@
...
+     */
+    [RequestActions.REMOVE_COLUMN_WITH_PARAMS](
+      this: SortConsumer,
+      route: Route,
+      columnMetadata: ColumnMetadataModels,
+      parameters: ColumnFragment['parameters']
+    ) {
+      const { routeName } = route;
+      const { request } = route.modelFor(routeName) as ReportModel;
+      const sorts = request.sorts.filter(
```
jkusa, line 93: Could we have more than one? Also, we use cname to find a column above, any reason not to do the same here?

kevinhinterlong: We don't have a column fragment here, so I could call canonicalizeMetric but either should work. I'm expecting that sorts will always have a length of 0 or 1, but it seemed right to use `.filter` instead of `.find`

packages/reports/addon/consumers/request/sort.ts
kevinhinterlong: I was planning to have another PR to slim down the consumer interface so that we could look at it as a whole

packages/reports/addon/consumers/request/sort.ts
jkusa: > but it seemed right to use .filter instead of .find
Agreeded, should we do the same for the other actions that use `find`?

jkusa (approved):

PR #1150, opened 2020-12-08
Upgrade elide version to 5.0.0-pr28

rishi-aga:
Upgrade elide version to 5.0.0-pr28
Set `isFact` as false for `Report` model so it doesn't show up as table.

## License

### DIFF
--- a/packages/webservice/models/build.gradle.kts
+++ b/packages/webservice/models/build.gradle.kts
@@ -30,7 +30,6 @@ allOpen {
 
 dependencies {
     implementation("com.yahoo.elide", "elide-core", "5.0.0-pr28")
-    implementation("com.yahoo.elide", "elide-datastore-aggregation", "5.0.0-pr28")
     implementation("javax.persistence", "javax.persistence-api", "2.2")
     implementation("org.hibernate", "hibernate-core", "5.4.15.Final")
 }
--- a/packages/webservice/models/build.gradle.kts
+++ b/packages/webservice/models/build.gradle.kts
@@ -29,7 +29,8 @@ allOpen {
 }
 
 dependencies {
-    implementation("com.yahoo.elide", "elide-core", "5.0.0-pr27")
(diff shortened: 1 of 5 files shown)


packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/Report.kt
```diff
@@ -25,6 +26,7 @@ import javax.persistence.Table
 @CreatePermission(expression = IS_AUTHOR)
 @UpdatePermission(expression = IS_AUTHOR)
 @DeletePermission(expression = IS_AUTHOR)
+@TableMeta(isFact = false)
```
rishi-aga, line 29: So that Navi can filter this.

moizarafat (approved):

packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/Report.kt
aklish: We should not need this

packages/webservice/models/build.gradle.kts
```diff
@@ -29,7 +29,8 @@ allOpen {
 }
 
 dependencies {
-    implementation("com.yahoo.elide", "elide-core", "5.0.0-pr27")
+    implementation("com.yahoo.elide", "elide-core", "5.0.0-pr28")
+    implementation("com.yahoo.elide", "elide-datastore-aggregation", "5.0.0-pr28")
```
aklish, line 33: We should not need this.  See other comment

aklish (approved):

packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/Report.kt
rishi-aga: removed.

PR #1164, opened 2020-12-12
Fix dashboard tests

kevinhinterlong:
## Description
Fixes failing dashboards tests

## Proposed Changes
- fixed a bug with dashboard-data service where we yielded a promise instead of a TaskInstance so cancelling did not bubble correctly
- removed date dimension filter
- catch errors when calling `getURL`
- Fili dimensions use `primaryKey` field as default value for dimension field parameter

## License

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -21,7 +21,7 @@ jobs:
       - test-navi-admin: npx lerna run test --scope navi-admin --stream;
       - test-navi-app: npx lerna run test --scope navi-app --stream;
       - test-navi-core: npx lerna run test --scope navi-core --stream;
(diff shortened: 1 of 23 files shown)


screwdriver.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       - test-navi-admin: npx lerna run test --scope navi-admin --stream;
       - test-navi-app: npx lerna run test --scope navi-app --stream;
       - test-navi-core: npx lerna run test --scope navi-core --stream;
-      - test-navi-dashboards: npx lerna run test --scope navi-dashboards --stream || true;
+      - test-navi-dashboards: npx lerna run test --scope navi-dashboards --stream;
```
jkusa, line 24: 🎉

jkusa (approved):

PR #1171, opened 2020-12-16
Consumer cleanup

### LINKED ISSUE #1170: Update filter grain in fili consumer to match column grain if it exists
Steps:
1. Add dateTime column
2. Update grain to non-default (week, month, etc)
3. Add dateTime filter

The filter grain uses the default and doesn't check if the column exists

kevinhinterlong:
Resolves #1170 

## Description
We have a lot of duplicate consumers left over from request v1 actions

## Proposed Changes
- remove uneeded consumers
- fix fili issue where time dimension filter vs column grain get out of sync

## License


packages/core/addon/models/bard-request-v2/request.ts
```diff
@@ -241,35 +247,25 @@ export default class RequestFragment extends Fragment.extend(Validations) implem
...
-   * Adds a new column with the parameters
-   */
-  addColumnFromMeta(columnMetadataModel: ColumnMetadataModels, parameters: Parameters): ColumnFragment {
-    return this.columns.pushObject(this.fragmentFactory.createColumnFromMeta(columnMetadataModel, parameters));
-  }
-
   /**
    * Adds a column with it's default parameters
    */
   addColumnFromMetaWithParams(columnMetadataModel: ColumnMetadataModels, parameters: Parameters = {}): ColumnFragment {
```
jkusa, line 253: should we rename this to `addColumnFromMeta`?

jkusa (approved):

packages/core/addon/models/bard-request-v2/request.ts
kevinhinterlong: Yea I think once we expect all tests to be passing we can start with some renaming, but maybe skip it for now

PR #1176, opened 2020-12-18
Update: Fix all but one test in reports package

kevinhinterlong:
## Description
Navi reports package is almost passing

## Proposed Changes
- Remove number dimension
- remove some v1 helpers
- fix v2 sort building (didn't show params)

## License


packages/reports/addon/templates/components/metric-selector.hbs
```diff
@@ -43,18 +43,10 @@
...
             >
               <EmberTooltip @side="right" @popperContainer="body" @effect="none">
-                {{#if (has-parameters metric)}}
-                  {{#if (has-unfiltered-parameters metric @request)}}
-                    Add next filter
-                  {{else}}
-                    Remove all
-                  {{/if}}
+                {{#if (get-shallow this.metricsFiltered metric.id)}}
+                  Remove Filter
```
jkusa, line 47: With parameterized columns, I wonder if this still makes sense

packages/reports/addon/templates/components/metric-selector.hbs
```diff
@@ -43,18 +43,10 @@
...
                 {{else}}
-                  {{#if (get-shallow this.metricsFiltered metric.id)}}
```
jkusa, line 53: Nice clean up here

jkusa (approved):
🔥

jkusa:
Looks like we have TS errors

packages/reports/addon/templates/components/metric-selector.hbs
kevinhinterlong: Yea I think at some point we talked about moving away from the toggle filter setup to just adding a new filter. It's also in a slightly weird state now since you can't change parameters from filters so you likely need to add/remove the filter from the column config

jkusa (approved):

PR #1187, opened 2020-12-29
New: Simple Request V2 Builder.

cythrawll:
## Description

In our projects we often use the request builder to make custom dashboards. Now that we moving onto RequestV2 would nice to have a builder for it.

## Proposed Changes

- Add a TypeScript RequestV2 builder
- Works a bit differently from the old one. Instead of attempting to do immutable operations. does a deep copy of the base request and then holds it as private state until you call build().

## License

### DIFF
--- a/packages/data/addon/builder/requestV2.ts
+++ b/packages/data/addon/builder/requestV2.ts
@@ -1,0 +1,176 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ *
+ * Helper object for building request object used by fact service
+ */
+import { RequestV2, Parameters, FilterOperator, SortDirection } from 'navi-data/adapters/facts/interface';
+import { ColumnType } from 'navi-data/models/metadata/column';
+import { getDefaultDataSourceName } from 'navi-data/utils/adapter';
+import { nanoid } from 'nanoid';
+import { assert } from '@ember/debug';
+
+/**
+ * Utility Class for building a request V2
+ */
+export default class RequestV2Builder {
+  /**
+   * @property The request being built
+   */
+  private request: RequestV2;
+
+  /**
+   * @param request Optional request to use as base.
+   */
+  constructor(request?: RequestV2) {
+    if (!request) {
+      this.request = {
+        filters: [],
+        columns: [],
+        table: '',
+        dataSource: getDefaultDataSourceName(),
+        sorts: [],
+        limit: null,
+        requestVersion: '2.0'
+      };
+    } else {
+      this.request = JSON.parse(JSON.stringify(request));
+    }
+  }
+
+  /**
+   * Adds a column to the request
+   * @param type - metric, dimension, timeDimension
+   * @param field - name of the metric/dimension or timeDimension
+   * @param parameters - Dictionary of parameters
+   * @param alias - Optoinal Display name for the field
+   * @param cid - column id, generated if not given
+   */
+  column(type: ColumnType, field: string, parameters: Parameters, alias?: string, cid?: string): RequestV2Builder {
+    this.request.columns.push({
+      type,
+      field,
+      parameters,
+      ...(alias ? { alias } : null),
+      cid: cid ?? nanoid(10)
+    });
+    return this;
+  }
+
+  /**
+   * Convenience for adding a metric column
+   * @param metricName Name of metric
+   * @param parameters Dictionary of parameters
+   */
+  metric(metricName: string, parameters: Parameters = {}): RequestV2Builder {
+    this.column('metric', metricName, parameters);
+    return this;
+  }
+
+  /**
+   * Convenience for adding a dimension column
+   * @param dimensionName Name of dimension
+   * @param parameters Dictionary of parameters
+   */
+  dimension(dimensionName: string, parameters: Parameters = {}): RequestV2Builder {
+    this.column('dimension', dimensionName, parameters);
+    return this;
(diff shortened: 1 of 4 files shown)


packages/data/addon/builder/requestV2.ts
```diff
@@ -0,0 +1,176 @@
...
+    return this;
+  }
+
+  /**
+   * Convenience method for adding an interval filter
+   * @param grain 'day','week','month' etc... whatever your datasource supports
+   * @param start Period, datestring or macro
+   * @param end Period, datestring or macro
+   */
+  interval(grain: string, start: string, end: string): RequestV2Builder {
```
jkusa, line 155: We are moving away from the concept of interval, should we rename this to something like this:

```suggestion
  addTimeRangeFilter(grain: Grain, start: string, end: string): RequestV2Builder {
```


Also, we can use the `Grain` type from packages `navi-data/addon/utils/date.ts`

packages/data/addon/builder/requestV2.ts
```diff
@@ -0,0 +1,176 @@
...
+    const table = this.request.table;
+    assert('Set the table before you set the interval', table.length !== 0);
+
+    this.column('timeDimension', `${table}.dateTime`, { grain }).filter(
+      'timeDimension',
+      `${table}.dateTime`,
```
jkusa, line 161: This is only really a Fili/Bard thing. Ideally, all the methods would take a column `id`, whatever it maybe. For Fili/Bard, the `id` for a timeDimension would be `tableA.dateTime`.

I guess this logic would make sense if this was a BardRequest(Builder).

packages/data/addon/builder/requestV2.ts
```diff
@@ -0,0 +1,176 @@
...
+    );
+
+    return this;
+  }
+
+  /**
+   * Returns the built request
+   */
+  build() {
+    return this.request;
```
jkusa, line 174: Are we planning to add significant logic in `build`? If not, can we make this an object that implements `RequestV2` with the convenient methods? This way we can pass the object without calling `build()`.

cythrawll: So the model is actually a bit overkill for this. The model comes with a lot of weight and dependencies that make building awkward, when all we need is a pojo that has the right interface to throw at the adapter.

packages/data/addon/builder/requestV2.ts
cythrawll: There isn't a concept of a Bard Request V2 and Elide Request V2, and probably better off that way.  I think the idea is that we do this particular thing enough in Bard/Fili that we really need a convenience method for this to make the builder worth it.  Those that are making an elide request without a date time filter simply not use this, or use raw addColumn/addFilter methods.  If there is a way to make this work conveniently for both I'm open to it. 

I could simply just add an optional column id to use here, and default to tableA.dateTime, Or I could add an option switch. Or I could just make this: `addBardInterval`.

packages/data/addon/builder/requestV2.ts
cythrawll: Also it wouldn't follow the builder pattern that way. The idea is to encapsulate the object while it's in an intermediate state as well as separate out building state concerns...

packages/data/addon/builder/requestV2.ts
jkusa: > There isn't a concept of a Bard Request V2 and Elide Request V2

Correct, but this is a higher level object than the RequestV2 interface that could conveniently normalize down to the required interface. However, this would not be my first choice 😄 

>  Those that are making an elide request without a date time filter simply not use this

If this is a general request object, I don't think we want a method that only works on specific data sources


I think the simplest approach would be to remove this logic and require users to pass the `id` as is. I agree it loses some convenience for bard users, but I think it is the most straight forward. Especially when the response column will be in the same form (`{table}.{timeDimension}`)

packages/data/addon/builder/requestV2.ts
jkusa: When you say `model` are you speaking to the ember data model/fragment? I am not suggesting that.

The `build()` method, as is, doesn't offer anything. It is not doing any validation, optimization, or heavy lifting to build the object (like Java's StringBuilder).

>The intent of the Builder design pattern is to separate the construction of a complex object from its representation. By doing so, the same construction process can create different representations.

I am just not sure we are really gaining anything from keeping the request as internal state and then requiring the user to call `build`. We can still have method-chaining without a `build` method. 

If you really wanted a terminal method, it would be cool it you could just call `fetch` (like in JOOQ). Although, that is a little more complicated.

packages/data/addon/builder/requestV2.ts
cythrawll: sure that's fine.

packages/data/addon/builder/requestV2.ts
cythrawll: Build simply does what most builders do, and unwraps the built object to be used. Basically it pulls the state out of the encapsulation. In most languages there would be a `new` here but in prototype languages that's building a pojo that's not really necessary. especially in TypeScript where we can check pojos against an interface like what we're doing here.

The gain we get is that we get an immutable state of each builder instance which is important for the use case as each dashboard usually creates a base query and then forks it in different ways to create a myriad of different queries.  I'm afraid without build() we would break that much needed encapsulation and make errors more likely.

I mean it kind of comes down to abandoning the builder pattern and replacing it with mutable class that is meant to be yet another implementation to requestV2. And also complicating the deep copy we have to do in the constructor.

PR #1188, opened 2021-01-04
friendly name changes

ayeswarya:
Resolves #

## Proposed Changes

-

## License


kevinhinterlong (approved):

packages/data/addon/mirage/factories/time-dimension.js
```diff
@@ -16,6 +16,10 @@ export default Factory.extend({
     return `Time Dimension ${this.index}`;
   },
 
+  friendlyName() {
+    return `Friendly Time Dimension ${this.index}`;
```
jkusa, line 20: Can we make this the same as the current `name()` value and change `name()` to be the `id()` value without the `table.` prefix?

jkusa (approved):

PR #1191, opened 2021-01-05
Report rerun

ayeswarya:
Resolves #

## Proposed Changes

The rerun window that appears when a new attribute is added to a report was not displaying fine. I have tried to fix it in this PR. Tested locally.

## License


jkusa:
As discussed offline, we do not want a modal experience when updating a request.

PR #1193, opened 2021-01-12
Convert to Isnull operator

kevinhinterlong:
## Description
Instead of having two operators `null` and `notnull` with `values = [true]` move to using `isnull` with `values = [true] | [false]`

## Proposed Changes
- `isnull` operator
- remove global replace null request decorator
- refactor some filter builder logic

## License


packages/data/addon/adapters/facts/bard.ts
```diff
@@ -58,12 +58,21 @@ function formatDimensionFieldName(field: string, parameters: Parameters, include
...
         .join(','); // comma to separate
       const formattedFieldName = formatDimensionFieldName(field, parameters, true);
+      if (operator === 'isnull') {
+        if (values[0] === true) {
+          operator = 'in';
+          serializedValues = '""';
+        } else if (values[0] == false) {
+          operator = 'notin';
+          serializedValues = '""';
+        }
```
jkusa, line 74:
```suggestion
        if (values[0] === true) {
          operator = 'in';
        } else if (values[0] == false) {
          operator = 'notin';
        } else {
          throw new FactAdapterError(`isnull operator can only have boolean values, found: ${value[0]}`);
        }
        serializedValues = '""';

```

packages/data/addon/adapters/facts/bard.ts
```diff
@@ -58,12 +58,23 @@ function formatDimensionFieldName(field: string, parameters: Parameters, include
...
         .map(v => `"${v}"`) // wrap each "value"
         .join(','); // comma to separate
       const formattedFieldName = formatDimensionFieldName(field, parameters, true);
+      if (operator === 'isnull') {
+        if (values[0] === true) {
+          operator = 'in';
+        } else if (values[0] === false) {
+          operator = 'notin';
+        } else {
+          throw new FactAdapterError(`isnull operator can only have boolean values, found: ${value[0]}`);
```
kevinhinterlong, line 73:
```suggestion
          throw new FactAdapterError(`isnull operator can only have boolean values, found: ${values[0]}`);
```

jkusa (approved):

PR #1201, opened 2021-01-15
Update: Use show clause for fili dimension fields

### LINKED ISSUE #958: Coalesce dimension fields on requested columns

kevinhinterlong:
Resolves #958 

## Description
Adds support for using the show clause on data requests to fili so instead of `/table/day/age/` -> `/table/day/age;show=id,key/`

## Proposed Changes
- Use show clause if dimension field is specified

## License


packages/data/addon/mirage/routes/bard-lite-parsers.ts
```diff
@@ -43,34 +51,35 @@ export function parseFilters(filtersParam?: string) {
   return filterStrings.map(parseSingleFilter);
 }
 
-type Having = { [key: string]: { operator: string; values: string[] } };
+export type HavingOperator = 'gt' | 'gte' | 'lt' | 'lte' | 'eq' | 'neq' | 'bet' | 'nbet';
```
jkusa, line 54: Would this be more useful in the adapter?

packages/data/addon/mirage/routes/bard-lite.ts
```diff
@@ -0,0 +1,308 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
```
jkusa, line 2:
```suggestion
 * Copyright 2021, Yahoo Holdings Inc.
```

jkusa (approved):

cythrawll (approved):

packages/data/addon/mirage/routes/bard-lite-parsers.ts
kevinhinterlong: It's only used in fili mirage, but we could export it from there. I don't think we'll get much use from putting it in the adapter though

PR #1217, opened 2021-01-27
Update: Remove all grain

### LINKED ISSUE #1207: Remove `all` time grain as a concept (only exists in fili adapter)

kevinhinterlong:
Resolves #1207 

## Description
The all time grain adds a lot of complexity because it's better modeled as a column that we are not grouping by

## Proposed Changes
- Remove all grain from date time parameters
- all grain is now indicated by the presence of a date time column
  - an error will be thrown if the table does not support it (only if table metadata loaded)
- Enhances the fili consumer so that timegrains can't get out of sync even with multiple date time columns
- some fixes for hour grain issues

## License

### DIFF
--- a/packages/data/addon/adapters/facts/bard.ts
+++ b/packages/data/addon/adapters/facts/bard.ts
@@ -22,6 +22,10 @@ import NaviFactAdapter, {
   Sort
 } from './interface';
 import { omit } from 'lodash-es';
+import NaviMetadataService from 'navi-data/services/navi-metadata';
+import BardTableMetadataModel from 'navi-data/models/metadata/bard/table';
+import { GrainWithAll } from 'navi-data/serializers/metadata/bard';
+import { Grain } from 'navi-data/utils/date';
 
 export type Query = RequestOptions & Dict<string | number | boolean>;
 export class FactAdapterError extends Error {
@@ -96,18 +100,12 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
    */
   namespace = 'v1/data';
 
-  /**
-   * @property {Service} ajax
-   */
   @service ajax: TODO;
+
+  @service naviMetadata!: NaviMetadataService;
 
   /**
    * Builds the dimensions path for a request
-   *
-   * @method _buildDimensionsPath
-   * @private
-   * @param request
-   * @return dimensions path
    */
   _buildDimensionsPath(request: RequestV2 /*options*/): string {
     const dimensionToFields = request.columns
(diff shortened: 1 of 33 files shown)


packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -278,10 +318,16 @@ export default class BardMetadataSerializer extends NaviMetadataSerializer {
...
+  private createTimeGrainColumnFunction(
+    table: RawTablePayload,
+    timeGrainInfo: TableTimeGrainInfo,
+    dataSourceName: string
+  ): ColumnFunctionMetadataModel {
+    const { hasAllGrain } = timeGrainInfo;
+    const grainIds = timeGrainInfo.timeGrains.map(g => g.name);
+
+    const grains = grainIds.join(',');
+    const columnFunctionId = `${this.namespace}:timeGrain(table=${table.name};grains=${grains};hasAllGrain=${hasAllGrain})`;
```
jkusa, line 330: What is the case when you need `hasAllGrain=${hasAllGrain}` to distinguish between different column function?

packages/reports/addon/components/filter-values/time-dimension/base.ts
```diff
@@ -23,16 +23,11 @@ export default class BaseTimeDimensionFilter extends Component<TimeDimensionFilt
   @readOnly('args.filter.parameters.grain') dateTimePeriod!: Grain;
 
   /**
-   * the active dateTimePeriod which does not include 'all'
+   * the active dateTimePeriod
    */
   @computed('dateTimePeriod')
   get calendarDateTimePeriod() {
```
jkusa, line 29: Do we need this property anymore, can we just use the arg?

packages/reports/addon/consumers/request/fili.ts
```diff
@@ -29,6 +32,31 @@ export default class FiliConsumer extends ActionConsumer {
   }
 
   actions = {
+    /**
+     * @action UPDATE_COLUMN_FRAGMENT_WITH_PARAMS
+     * @param route - route that has a model that contains a request property
+     * @param columnFragment - data model fragment of the column
+     * @param parameterKey - the name of the parameter to update
+     * @param parameterValue - the value to update the parameter with
```
jkusa, line 40:
```suggestion
     * @param changeset - filter updates
``````

File: packages/reports/addon/components/filter-values/time-dimension/base.ts (hunk 2 above)
Comment (kevinhinterlong) [reply packages/reports/addon/components/filter-values/time-dimension/base.ts:29]: I was planning on keeping this for now because I think it could be useful for sub day grains

File: packages/data/addon/serializers/metadata/bard.ts (hunk 1 above)
Comment (kevinhinterlong) [reply packages/data/addon/serializers/metadata/bard.ts:330]: Ah good call I'll remove this it shouldn't make a difference

File: packages/data/addon/adapters/facts/bard.ts [hunk 4]
```diff
@@ -237,9 +236,7 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
     const grains = [...new Set(dateTimeColumns.map(c => c.parameters.grain))] as Grain[];
     if (grains.length > 1) {
       throw new FactAdapterError(
-        `Requesting more than one '${request.table}.dateTime' grain is not supported. You requested [${grains.join(
-          ','
-        )}]`
+        `Requesting more than multiple 'Date Time' grains is not supported. You requested [${grains.join(',')}]`
```
jkusa, line 239:
```suggestion
        `Requesting multiple 'Date Time' grains is not supported. You requested [${grains.join(',')}]`
```

jkusa (approved):
Just the one comment, otherwise 👍

PR #1235, opened 2021-02-09
Update: Convert intervals to [inclusive, inclusive]

kevinhinterlong:
## Description
Converts [inclusive, exclusive) v1 filters to [inclusive, inclusive] for v2

## Proposed Changes
- Use ISO String to store date filter values

## License

### DIFF
--- a/packages/data/addon/adapters/facts/bard.ts
+++ b/packages/data/addon/adapters/facts/bard.ts
@@ -24,7 +24,8 @@ import { omit } from 'lodash-es';
 import NaviMetadataService from 'navi-data/services/navi-metadata';
 import BardTableMetadataModel from 'navi-data/models/metadata/bard/table';
 import { GrainWithAll } from 'navi-data/serializers/metadata/bard';
-import { Grain } from 'navi-data/utils/date';
+import { getPeriodForGrain, Grain } from 'navi-data/utils/date';
+import moment from 'moment';
 
 export type Query = RequestOptions & Dict<string | number | boolean>;
 export class FactAdapterError extends Error {
@@ -140,7 +141,7 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
       );
     }
     const timeFilter = dateTimeFilters[0];
-    const filterGrain = timeFilter?.parameters?.grain;
+    const filterGrain = timeFilter?.parameters?.grain as Grain;
 
     const timeColumn = request.columns.filter(isDateTime)[0];
     const columnGrain = timeColumn?.parameters?.grain;
@@ -150,8 +151,15 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
       );
     }
 
-    const [start, end] = timeFilter.values;
+    let [start, end] = timeFilter.values as string[];
 
+    const endMoment = end ? moment.utc(end) : undefined;
+    if (endMoment?.isValid()) {
+      end = endMoment.add(1, getPeriodForGrain(filterGrain)).toISOString();
+    }
+
+    start = start.replace('T00:00:00.000Z', '');
+    end = end.replace('T00:00:00.000Z', '');
     return `${start}/${end}`;
   }
 
--- a/packages/data/addon/adapters/facts/bard.ts
+++ b/packages/data/addon/adapters/facts/bard.ts
@@ -158,8 +158,9 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
       end = endMoment.add(1, getPeriodForGrain(filterGrain)).toISOString();
     }
 
-    start = start.replace('T00:00:00.000Z', '');
-    end = end.replace('T00:00:00.000Z', '');
+    // Removing Z to strip off time zone if it's there
+    start = start.replace('Z', '');
+    end = end.replace('Z', '');
     return `${start}/${end}`;
   }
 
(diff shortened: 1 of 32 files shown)


packages/data/addon/adapters/facts/bard.ts
```diff
@@ -150,8 +151,15 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
...
 
-    const [start, end] = timeFilter.values;
+    let [start, end] = timeFilter.values as string[];
 
+    const endMoment = end ? moment.utc(end) : undefined;
+    if (endMoment?.isValid()) {
+      end = endMoment.add(1, getPeriodForGrain(filterGrain)).toISOString();
+    }
+
+    start = start.replace('T00:00:00.000Z', '');
```
jkusa, line 161: How would this work with Fili if you had `T00:00:00.000Z` in the value with an hourly grain?

packages/data/addon/mirage/routes/bard-lite.ts
```diff
@@ -187,6 +187,15 @@ export default function(
 
     // Get date range from query params + grain
     const [start, end] = request.queryParams.dateTime.split('/');
+    if (start === end) {
+      return new Response(
```
jkusa, line 191: 👍

packages/data/addon/utils/classes/interval.ts
```diff
@@ -127,24 +127,44 @@ export default class Interval {
    * @param makeEndInclusiveIfSame - add an extra time period to include end date if it's equal to start date
    * @returns object with start and end properties
    */
-  asMomentsForTimePeriod(grain: Grain, makeEndInclusiveIfSame = true): SerializedWithEnd<Moment> {
```
jkusa, line 130: Can we update the docs do we know we are getting inclusive, inclusive?

Also, line 127 can be removed.

packages/data/addon/utils/classes/interval.ts
```diff
@@ -196,7 +216,7 @@ export default class Interval {
   diffForTimePeriod(grain: Grain): number {
     const moments = this.asMomentsForTimePeriod(grain);
     const period = getPeriodForGrain(grain);
-    return moments.end.diff(moments.start, period);
+    return moments.end.diff(moments.start, period) + 1;
```
jkusa, line 219: Does an `Interval` still hold `inclusive/exclusive` internally?

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -197,6 +197,7 @@ module('Unit | Adapter | facts/elide', function(hooks) {
       'Request with "not between" filter operator splits the filter into two correctly'
     );
 
+    debugger;
```
jkusa, line 200:
```suggestion
```

packages/data/tests/unit/utils/classes/interval-test.ts
```diff
@@ -80,19 +86,19 @@ module('Unit | Utils | Interval Class', function() {
     );
 
     assert.equal(
-      new Interval(moment('2014-10-10'), moment('2015-10-10')).diffForTimePeriod('day'),
+      new Interval(moment('2014-10-10'), moment('2015-10-09')).diffForTimePeriod('day'),
```
jkusa, line 89: Should `Interval` still be  inclusive/exclusive as defined?

packages/data/addon/adapters/facts/bard.ts
kevinhinterlong: It wouldn't (I was planning to come back to fix `hourly` as a whole), but it looks like we can update this to just remove the `Z` because [timeZone has to be passed in as a param](https://github.com/yahoo/fili/blob/7c022c7368f7c99a83f4fe5b36731783af48a8b5/docs/end-user-api.md#time-zone)

packages/reports/addon/components/filter-values/time-dimension/base-interval.ts
```diff
@@ -45,43 +50,39 @@ export default class BaseIntervalComponent extends BaseTimeDimensionFilter {
...
+    const { interval, grain } = this;
     if (interval) {
-      let { end } = interval.asMomentsForTimePeriod(calendarDateTimePeriod, false);
-      end = end
-        .clone()
-        .subtract(1, getPeriodForGrain(calendarDateTimePeriod))
-        .utc(true);
-      return end;
+      let { end } = interval.asMomentsForTimePeriod(grain);
+      return end.subtract(1, getPeriodForGrain(grain));
```
jkusa, line 58: Since we do `end.subtract(...` quite a bit, would it make sense to add that logic to `asMomentsForTimePeriod`?
```js
const inclusiveEnd = true;
let { end } = interval.asMomentsForTimePeriod(grain, inclusiveEnd);
```

packages/reports/addon/consumers/request/fili.ts
```diff
@@ -47,8 +77,10 @@ export default class FiliConsumer extends ActionConsumer {
       const { routeName } = route;
       const { request } = route.modelFor(routeName) as ReportModel;
       const { dateTimeFilter } = request;
+      debugger;
```
jkusa, line 80:
```suggestion
```

jkusa (approved):
Just the `debugger` cleanup, otherwise 👍

PR #1240, opened 2021-02-13
Update: Extra TS files, remove unneeded services

kevinhinterlong:
## Description
Moves some files to TS

## Proposed Changes
- Remove metric name service
- Remove metric parameter service

## License


packages/data/addon/mirage/factories/metric.ts
```diff
@@ -1,13 +1,14 @@
...
-import { Factory } from 'ember-cli-mirage';
+import { Factory } from 'miragejs';
 
 export default Factory.extend({
-  index: i => i,
+  index: (i: number) => i,
 
-  id() {
+  id(): string {
+    //@ts-ignore
```
jkusa, line 11: What's the issue here? Accessing properties on `this`?

jkusa (approved):
Just the question about `//@ts-ignore`, otherwise 👍

packages/data/addon/mirage/factories/metric.ts
kevinhinterlong: I couldn't figure out how to use the `'miragejs'` types to get the definition of `'table'` (we only define the factory and I was struggling to figure out how to extract the instance definition) so I just ignored it for now since I didn't think it was worth investigating

jkusa (approved):

PR #1252, opened 2021-02-19
Update: Added relative database file for saves across service restarts

aklish:
## Description

Adds a permanent database file to persist changes across service restarts.

## License

### DIFF
--- a/packages/webservice/app/src/main/resources/application.yaml
+++ b/packages/webservice/app/src/main/resources/application.yaml
@@ -36,7 +36,7 @@ spring:
       jdbc:
         use_scrollable_resultset: true
   datasource:
-    url: 'jdbc:h2:mem:db1;DB_CLOSE_DELAY=-1'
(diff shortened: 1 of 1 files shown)


packages/webservice/app/src/main/resources/application.yaml
```diff
@@ -36,7 +36,7 @@ spring:
       jdbc:
         use_scrollable_resultset: true
   datasource:
-    url: 'jdbc:h2:mem:db1;DB_CLOSE_DELAY=-1'
+    url: 'jdbc:h2:file:./yavindb;DB_CLOSE_DELAY=-1'
```
jkusa, line 39:
```suggestion
    url: 'jdbc:h2:file:./yavin.db;DB_CLOSE_DELAY=-1'
```
??

aklish: actual filename is : yavindb.h2.db

aklish: Should we just call it yavin?

jkusa (approved):

PR #1263, opened 2021-02-25
Update: Visualization blueprints

kevinhinterlong:
## Description
Modernize visualization blueprints

## License


jkusa (approved):
Looks good. My only thought is, should we expect TS?

PR #1269, opened 2021-03-02
Update: Better UX for elide filters

kevinhinterlong:
## Description
Better flow for dimension filters in elide. 

Need some better documentation around filter values being populated without searching if the dimension has a small cardinality (and that this must be configured in the semantic model)

## Proposed Changes
- Expand filter dropdown to show all operators
- Case insensitive search for filter values
- Configure dimension cardinality for elide

## License


packages/data/addon/models/metadata/dimension.ts
```diff
@@ -60,22 +59,17 @@ export default class DimensionMetadataModel
...
-
-    if (type === 'field') return this._cardinality;
-
-    // TODO: get cardinality for ref and formula type dimensions
-    return undefined;
+  get cardinality(): Cardinality | undefined {
+    return this._cardinality;
   }
-  set cardinality(cardinality: Cardinality) {
+  set cardinality(cardinality: Cardinality | undefined) {
```
jkusa, line 72: Maybe we can just use TS to manage the values

jkusa (approved):
Just the one comment

jkusa (approved):

PR #1272, opened 2021-03-05
Table Export Queries

ayeswarya:
Resolves #

## Proposed Changes

-

## License


moizarafat (approved):

packages/webservice/app/src/main/resources/db/changelog/changelog.xml
```diff
@@ -108,4 +108,27 @@
         );
         </sql>
     </changeSet>
+    <changeSet author="admin" id="2">
+        <sql>
+          CREATE TABLE TableExport (
+             id varchar(36) not null,
+             createdOn timestamp,
+             principalName varchar(255),
```
jkusa, line 116: How does this relate to the user table?

aklish: the principal name is the user principal name (should ideally match the user table).

jkusa: Could we add a foreign key to the user table?

moizarafat: Is there an entry in the `User` table available for No-Authentication scenario? When we use it for a demo, `principalName` is going to be null.

jkusa: yes, we add "admin" in the demo case

moizarafat: Verified that the `principalName` is not null for AsyncQuery in Yavin App, its set to `admin` in case of No Auth scenario. Adding a Foreign Key for AsyncQuery and TableExport should be fine.

moizarafat: https://github.com/yahoo/yavin/blob/5e5aa18cabb5bd3ee63a8cbf0436d2a3ecf96be7/packages/webservice/app/src/main/kotlin/com/yahoo/navi/ws/filters/AuthFilter.kt#L24

moizarafat (approved):

packages/webservice/app/src/main/resources/db/changelog/changelog.xml
```diff
@@ -84,28 +84,55 @@
...
+            createdOn timestamp,
+            principalName varchar(255),
+            query varchar(255),
+            queryType integer,
+            requestId varchar(36),
+            resultType varchar(255),
+            status varchar(255),
+            updatedOn timestamp,
+            contentLength integer,
+            responseBody text,
```
jkusa, line 127: @moizarafat do store a query result here? Should we make this `MEDIUMTEXT like we do in `AsyncQuery`?
```suggestion
            responseBody MEDIUMTEXT,
```

packages/webservice/app/src/main/resources/db/changelog/changelog.xml
```diff
@@ -84,28 +84,55 @@
...
+            completedOn timestamp,
+            httpStatus integer,
+            recordCount integer,
+            message text,
```
jkusa, line 131:
```suggestion
            message varchar(255),
```
??

packages/webservice/app/src/main/resources/db/changelog/changelog.xml
moizarafat: Good catch. Its a bug on our end. We don't need responseBody in the TableExport table. We need to fix the elide example.

packages/webservice/app/src/main/resources/db/changelog/changelog.xml
moizarafat: 255 is not enough. This is used for storing the error message returned by the query and will vary. In elide example we kept it same as responsebody.

jkusa (approved):

PR #1285, opened 2021-03-12
Fix: Elide enum value search

kevinhinterlong:
## Description
We used `colN` aliasing convention for elide columns but didn't implement this for our mock enum response

## Proposed Changes
- use 'col0' as field instead of elideField for enum responses

## License

### DIFF
--- a/packages/data/addon/adapters/dimensions/elide.ts
+++ b/packages/data/addon/adapters/dimensions/elide.ts
@@ -1,18 +1,19 @@
 /**
- * Copyright 2020, Yahoo Holdings Inc.
+ * Copyright 2021, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  *
  * Description: The adapter for the Elide dimension model.
  */
-import { getOwner } from '@ember/application';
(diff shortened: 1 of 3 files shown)


packages/data/addon/adapters/dimensions/elide.ts
```diff
@@ -30,8 +31,8 @@ export default class ElideDimensionAdapter extends EmberObject implements NaviDi
   private factAdapter: ElideFactsAdapter = getOwner(this).lookup('adapter:facts/elide');
 
   private formatEnumResponse(dimension: DimensionColumn, values: (string | number)[]): AsyncQueryResponse {
-    const { id, tableId } = dimension.columnMetadata;
-    const field = getElideField(id, dimension.parameters);
+    const { tableId } = dimension.columnMetadata;
+    const field = 'col0';
     const nodes = values.map((value) => `{"node":{"${field}":"${value}"}}`);
```
jkusa, line 36:
```suggestion
    const nodes = values.map((value) => `{"node":{"col1":"${value}"}}`);
```

jkusa (approved):
Small thing, otherwise 👍

PR #1291, opened 2021-03-16
Update: Sort dimension values in filter dropdown

kevinhinterlong:
## Description
The dimension select currently shows values directly as they are received from the backend

## Proposed Changes
- Sort the results so they are easier to scroll through
- add a check for sorting a list of string numbers to sort them numerically instead

## License


packages/reports/addon/components/filter-values/dimension-select.ts
```diff
@@ -1,46 +1,57 @@
...
 
 interface DimensionSelectComponentArgs {
   filter: FilterFragment;
   request: RequestFragment;
   onUpdateFilter(changeSet: Partial<FilterFragment>): void;
 }
 
+function isNumeric(num: string) {
+  //@ts-ignore
+  return !isNaN(num);
```
jkusa, line 32:
```suggestion
  return !isNaN(num as any);
```

packages/reports/addon/components/power-select-collection-options.ts
```diff
@@ -4,13 +4,26 @@
...
+    groupKey?: string;
+    sortKey?: string;
+    sortFn?: (options: ItemOptions) => ItemOptions;
+  };
+};
+
+export type IndexedOptions<Option = ItemOptions[number]> = { option: Option; idx: number };
+
 export default class PowerSelectCollectionOptions extends Options {
+  declare args: Args;
```
jkusa, line 25: Interesting, does something like this not work 
```
PowerSelectCollectionOptions<Args> extends Options
```
```
(PowerSelectCollectionOptions extends Options)<Args>

```

jkusa (approved):

packages/reports/addon/components/power-select-collection-options.ts
kevinhinterlong: Nope, easiest way is to override it like this unfortunately

packages/reports/addon/components/filter-values/dimension-select.ts
kevinhinterlong: Updated to use any instead of ts-ignore

PR #1300, opened 2021-03-18
Update: Build main SD jobs in parallel

kevinhinterlong:
## Description
Parallelize the sd build for the main job

We should also look into ways to speed up the WS build because it takes awhile to boot up for each test

## License

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -79,7 +79,6 @@ jobs:
     steps:
       - *install-node
       - *npm-ci
-      - *install-chrome
       - test-navi-webservice: cd packages/webservice && ./gradlew check && cd -;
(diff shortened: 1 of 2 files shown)


screwdriver.yaml
```diff
@@ -1,51 +1,115 @@
...
+      PACKAGE: navi-reports
+  main-search:
+    <<: *main-npm-test-job
+    environment:
+      PACKAGE: navi-search
+  main-webservice:
+    steps:
+      - *install-node
+      - *npm-ci
+      - *install-chrome
```
jkusa, line 82: do we need these steps for ws?

jkusa (approved):

PR #1333, opened 2021-03-23
Update: Use Ember concurrency for data interfaces/adapters

kevinhinterlong:
## Description
Update facts/dimension adapters and services to use ember-concurrency tasks. Cancels elide queries we no longer use

## Proposed Changes
- Convert most data interfaces/adapters to be task based
- dimension select local search if all results are fetched
- update typescript/ember-concurrency

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1349, opened 2021-04-01
Update: Dimension Service `all` should get ALL

kevinhinterlong:
## Description
This fixes a bug that was introduced by #1333 where the dimension select would fetch `all` values using the dimension service, but it wouldn't page through results (e.g. an elide dimension with 700 results but the pageSize is 500 you would only get the first 500 values).

## Proposed Changes
- The dimension service will page through dimensions until all values have been fetched
- Rename the `searchThresholds` config option to `cardinalities: { small, medium }`

## License

### DIFF
--- a/packages/data/addon/serializers/facts/interface.ts
+++ b/packages/data/addon/serializers/facts/interface.ts
@@ -1,11 +1,10 @@
 /**
- * Copyright 2020, Yahoo Holdings Inc.
+ * Copyright 2021, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
-
-import { RequestV1, RequestV2 } from 'navi-data/adapters/facts/interface';
-import NaviFactResponse from 'navi-data/models/navi-fact-response';
-import NaviAdapterError from 'navi-data/errors/navi-adapter-error';
+import type { RequestOptions, RequestV2 } from 'navi-data/adapters/facts/interface';
+import type NaviFactResponse from 'navi-data/models/navi-fact-response';
+import type NaviAdapterError from 'navi-data/errors/navi-adapter-error';
 
 export interface ResponseV1 {
   readonly rows: Array<Record<string, unknown>>;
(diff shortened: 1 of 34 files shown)


packages/data/addon/services/navi-dimension.ts
```diff
@@ -39,15 +40,40 @@ export default class NaviDimensionService extends Service {
...
     const { type: dataSourceType } = getDataSource(dimension.columnMetadata.source);
     const adapter = this.adapterFor(dataSourceType);
-    const payload: unknown = yield taskFor(adapter.all).perform(dimension, options);
-    return this.serializerFor(dataSourceType).normalize(dimension, payload);
+    const serializer = this.serializerFor(dataSourceType);
+    let moreResults = true;
+    let paginationOptions: Pick<ServiceOptions, 'page' | 'perPage'> = {};
+    let values: NaviDimensionModel[] = [];
+    while (moreResults) {
+      debugger;
```
jkusa, line 55:
```suggestion
```

packages/data/addon/serializers/facts/interface.ts
```diff
@@ -25,12 +23,12 @@ export default interface NaviFactSerializer {
    * @param payload - payload to normalize
    * @param request - request for response payload
    */
-  normalize(payload: unknown, request: RequestV1 | RequestV2): NaviFactResponse | undefined;
+  normalize(payload: unknown, request: RequestV2, options?: RequestOptions): NaviFactResponse | undefined;
```
jkusa, line 26: 🎉

packages/webservice/app/src/main/resources/application.yaml
```diff
@@ -1,5 +1,5 @@
 elide:
-  pageSize: 10000
+  pageSize: 1000
```
jkusa, line 2: This also impacts the fact query, do we want to keep this at `10000`?

jkusa (approved):
Just the one question, otherwise 👍

PR #1354, opened 2021-04-02
Update: Use vertical-collection in dimension-select

kevinhinterlong:
## Description
The dimension select component used ember-collection and assumed the height of each entry was the same causing issues when strings needed to wrap

## Proposed Changes
- Use vertical-collection instead of ember-collection

## Screenshots
Screen Shot 2021-04-02 at 10 16 32 AM
Screen Shot 2021-04-02 at 10 20 46 AM

## License


packages/reports/addon/components/power-select-collection-options.ts
```diff
@@ -24,45 +25,23 @@ export type IndexedOptions<Option = ItemOptions[number]> = { option: Option; idx
 export default class PowerSelectCollectionOptions extends Options {
   declare args: Args;
 
-  /**
-   * @property {Number} maxDisplayedItems - max number of item to show at once
-   */
-  maxDisplayedItems = 6;
+  guid = guidFor(this);
```
jkusa, line 28: Is this being used?

kevinhinterlong: 👍  good catch, the `id` was being overwritten by the `...attributes` so this is no longer needed

jkusa (approved):

PR #1370, opened 2021-04-24
Update: Helpers and small TS updates

kevinhinterlong:
## Description
Update some helpers to TS/remove observers/etc

## License


packages/reports/addon/helpers/model-for.ts
```diff
@@ -0,0 +1,29 @@
...
+   * in a route hierarchy. If the ancestor route's model was a promise,
+   * its resolved result is returned.
+   */
+  compute([name]: [string]) {
+    const route = getOwner(this).lookup(`route:${name}`) as Route | undefined;
+    if (!route) {
+      return;
+    }
+    if (this.route) {
+      this.route.controller.removeObserver('model', this, this.recompute);
```
jkusa, line 23: do we know if model is `tracked`?

jkusa (approved):
just the one question, otherwise 👍

packages/reports/addon/helpers/model-for.ts
kevinhinterlong: It does not appear to be https://github.com/emberjs/ember.js/blob/v3.26.1/packages/%40ember/controller/lib/controller_mixin.js#L41-L56

PR #1374, opened 2021-04-27
Fix: Dimensions should obey sorting from service

cythrawll:
## Description

Issue where sortFn was messing with searchAPI sorting of results.

## Proposed Changes

- Obey sorting from dimensionService when cardinality is MEDIUM or above.
- Testing was fruitless because ordering of sortFn is done visually only and does not reflect in DOM. (matrix display magic).

## License

### DIFF
--- a/packages/reports/addon/components/filter-values/dimension-select.js
+++ b/packages/reports/addon/components/filter-values/dimension-select.js
@@ -134,7 +134,6 @@ export default class DimensionSelectComponent extends Component {
     if (isNumericDimensionArray(dimensions)) {
       return sortBy(dimensions, [d => Number(d.option.value)]);
     }
-    console.log('odering by description');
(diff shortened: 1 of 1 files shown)


packages/reports/addon/components/filter-values/dimension-select.js
```diff
@@ -120,9 +127,14 @@ export default class DimensionSelectComponent extends Component {
...
   @action
   sortValues(dimensions) {
+    //don't sort if searching on server
+    if (!this.isSmallCardinality) {
+      return dimensions;
+    }
     if (isNumericDimensionArray(dimensions)) {
       return sortBy(dimensions, [d => Number(d.option.value)]);
     }
+    console.log('odering by description');
```
kevinhinterlong, line 137:
```suggestion
```

kevinhinterlong (approved):

PR #1378, opened 2021-05-05
Update: Use different compression library

kevinhinterlong:
## Description
Replace compression library to one with a more compatible license

## Proposed Changes
- Also fixed the ModelFor helper which had a `tracked` property that did not need to be tracked (created a lot of deprecation warnings)

## License


packages/core/addon/services/compression.ts
```diff
@@ -7,47 +7,39 @@
 import Service, { inject as service } from '@ember/service';
 import { assert } from '@ember/debug';
 //@ts-ignore
```
jkusa, line 9: Can we remove this with: https://www.npmjs.com/package/@types/lz-string

jkusa (approved):
Just the one comment otherwise 👍

PR #1392, opened 2021-05-14
Update: Make fili search feature configurable

jkusa:
Resolves #

## Proposed Changes

- Update: Make fili search feature configurable

**TODO** add `options` to server-generated-config model

## License

### DIFF
--- a/packages/data/addon/adapters/dimensions/bard.ts
+++ b/packages/data/addon/adapters/dimensions/bard.ts
@@ -19,7 +19,6 @@ import type { DimensionFilter } from './interface';
 import type { ServiceOptions } from 'navi-data/services/navi-dimension';
 import type DimensionMetadataModel from 'navi-data/models/metadata/dimension';
 import type { DimensionColumn } from 'navi-data/models/metadata/dimension';
-import type { FiliConfigOptions } from 'navi-config';
 
 const SUPPORTED_FILTER_OPERATORS = ['in', 'notin', 'startswith', 'contains'];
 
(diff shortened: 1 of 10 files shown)


packages/data/config/navi-config.d.ts
```diff
@@ -7,8 +7,14 @@ declare module 'navi-config' {
     description?: string;
     uri: string;
     type: string;
+    options?: FiliConfigOptions | ElideConfigOptions;
```
kevinhinterlong, line 10: Can we do something like

```ts
type BaseDataSource<Type, Options> = {
  name: string;
  displayName: string;
  description?: string;
  uri: string;
  type: Type;
  options?: Options;
};
export interface FiliConfigOptions {
  enableDimensionSearch?: boolean;
}
export interface ElideConfigOptions {
  // elide options here
}

type NaviDataSource = BaseDataSource<'bard', FiliConfigOptions> | BaseDataSource<'elide', ElideConfigOptions>
```

So that only the options for the current dataSource can be used

packages/data/addon/adapters/dimensions/bard.ts
```diff
@@ -154,8 +155,22 @@ export default class BardDimensionAdapter extends EmberObject implements NaviDim
     query: string,
     options: ServiceOptions = {}
   ): TaskGenerator<FiliDimensionResponse> {
-    const url = this._buildUrl(dimension, 'search');
-    const data: Record<string, string> = query ? { query } : {};
-    return yield this._find(url, data, options);
+    const filiOptions = getDataSource(dimension.columnMetadata.source).options as FiliConfigOptions;
```
kevinhinterlong, line 158: Then you could probably do something like

```ts
const { source } = dimension.columnMetadata;
assert('This is a bard dataSource', source === 'bard')
const filiOptions = getDataSource(source).options;
```
and not have to cast

kevinhinterlong (approved):

PR #1393, opened 2021-05-17
Update: Dimension Suggestion columns

kevinhinterlong:
## Description
Adds support for suggestion columns for dimensions

## Proposed Changes
- add all fili dimension fields as suggestions 
- convert models to not extend from EmberObject (removes the extra public interfaces, and just relies on the models themselves)

## Screenshots
dimension suggestion columns

## License

### DIFF
--- a/packages/data/tests/unit/models/metadata/dimension-test.ts
+++ b/packages/data/tests/unit/models/metadata/dimension-test.ts
@@ -5,8 +5,11 @@ import Pretender from 'pretender';
 import { TestContext } from 'ember-test-helpers';
 //@ts-ignore
 import metadataRoutes from 'navi-data/test-support/helpers/metadata-routes';
+import { Factory } from 'navi-data/models/native-with-create';
 
-let Payload: DimensionMetadataPayload, Dimension: DimensionMetadataModel;
+let Payload: DimensionMetadataPayload,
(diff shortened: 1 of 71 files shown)


packages/data/tests/unit/models/metadata/dimension-test.ts
```diff
@@ -125,12 +129,14 @@ module('Unit | Metadata Model | Dimension', function (hooks) {
 
     assert.deepEqual(Dimension.primaryKeyFieldName, 'id', 'primaryKeyFieldName returned `id` as the primary key field');
 
-    let nonId = DimensionMetadataModel.create({
+    //@ts-expect-error
```
jkusa, line 132: Why do we have errors here? Missing fields?

jkusa (commented):
Are we going to add suggestion fields to the filter clause in another pass?

packages/data/tests/unit/models/metadata/dimension-test.ts
kevinhinterlong: Yea now that `.create` expects the correct payload to be passed in this gives a type error because of the missing fields

kevinhinterlong:
> Are we going to add suggestion fields to the filter clause in another pass?

Yep, we'll have to think more about the filtering strategy so a second pass would be good

PR #1402, opened 2021-06-14
Fix: Use options fetching metricFn from everything

jkusa:
Resolves #

## Proposed Changes

- When fetching a non default bard datasource metadata, default datasource metric functions where being fetched. This PR fixes that.

## License

### DIFF
--- a/packages/data/addon/adapters/metadata/bard.ts
+++ b/packages/data/addon/adapters/metadata/bard.ts
@@ -68,7 +68,7 @@ export default class BardMetadataAdapter extends EmberObject implements NaviMeta
 
   async fetchEverything(options?: MetadataOptions): Promise<TODO> {
     const fullViewReq = this.query('table', '', { query: { format: 'fullview' }, ...options });
-    const metricFunctionsReq = this.fetchAll('columnFunction');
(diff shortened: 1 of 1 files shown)


packages/data/addon/adapters/metadata/bard.ts
```diff
@@ -68,7 +68,7 @@ export default class BardMetadataAdapter extends EmberObject implements NaviMeta
 
   async fetchEverything(options?: MetadataOptions): Promise<TODO> {
     const fullViewReq = this.query('table', '', { query: { format: 'fullview' }, ...options });
-    const metricFunctionsReq = this.fetchAll('columnFunction');
+    const metricFunctionsReq = this.fetchAll('columnFunction', options);
```
kevinhinterlong, line 71: Come back to add a test?

kevinhinterlong (approved):

packages/data/addon/adapters/metadata/bard.ts
jkusa: Yeah 👍

jkusa: https://github.com/yavin-dev/framework/issues/1403

PR #1420, opened 2021-06-30
Fix: Maven Central Migration

moizarafat:
## Description
Migrate to Maven Central for publishing "models" jar instead of Bintray. Bintray was EOL

## Proposed Changes
With the new changes the group changes to "dev.yavin" instead of "com.yahoo.navi" in maven central

## License


packages/webservice/models/build.gradle.kts
```diff
@@ -62,54 +67,45 @@ publishing {
...
+                    url.set("http://yavin.dev")
+                }
+            }
+        }
+    }
+    repositories {
+        maven {
+            val releasesRepoUrl = uri("https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/")
+            val snapshotsRepoUrl = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")
+            url = if (project.hasProperty("isSnapshot") && project.property("isSnapshot").toString().equals("true")) snapshotsRepoUrl else releasesRepoUrl
```
moizarafat, line 82: This line is evaluated before `gradle.taskGraph.whenReady` so the version did not contain "SNAPSHOT" in name so it was using releaseUrl instead of snapshotUrl. introduced a new property isSnapshot to control this.

jkusa: Maybe we should leave a comment in the code for this

kevinhinterlong (approved):

jkusa (approved):

PR #1422, opened 2021-07-02
Update: Add bard metadata model extension point

jkusa:
Resolves #

## Proposed Changes

- Add protected methods for metadata model creation as a extension point for custom logic

## License


packages/data/addon/serializers/metadata/bard.ts
```diff
@@ -582,10 +590,14 @@ export default class BardMetadataSerializer extends NaviMetadataSerializer {
         partialData: isNone(description),
         columnFunctionId: metricFunctionId,
       };
-      return this.metricFactory.create(payload);
+      return this.createMetricModel(payload);
     });
   }
 
+  protected createMetricModel(payload: MetricMetadataPayload): MetricMetadataModel {
```
kevinhinterlong, line 597: It would be nice to have this in the base class and maybe have the factories be private there so both bard/elide serializers only call these functions

kevinhinterlong (approved):
small comment

kevinhinterlong (approved):

PR #1430, opened 2021-07-14
Update: Rename `author` to `owner` in models

jkusa:
Resolves #

## Description

This is changes the model field `author` to `owner`

## Proposed Changes

- Cleaned up db table names

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1432, opened 2021-07-16
Delivery rule models

jkusa:
Resolves #

## Description

Needs #1430 

## Proposed Changes

- Add delivery rules

## License


jkusa:
@kevinhinterlong we might want to bump from 2.0 since this change could break users that have the old FB schema.

kevinhinterlong (approved):

PR #1442, opened 2021-08-10
Fix: Serializing __rollupMask is a dimension

cythrawll:
## Description

Missed in spec. since __rollupMask  is a dimension, get serialized as a string

## Proposed Changes

- Handle _rollupMask as a string.

## License


kevinhinterlong (approved):
👍 if it passes

PR #1453, opened 2021-08-25
New: Subtotal models and consumers.

cythrawll:
## Description

Port of alpha branch but is simplified. Instead of holding dimension model copies in rollup columns, we simply just hold a reference cid.

## Proposed Changes

- add rollup to request model and add consumers to modify request.

## License

### DIFF
--- a/packages/core/addon/models/bard-request-v2/fragments/rollup.ts
+++ b/packages/core/addon/models/bard-request-v2/fragments/rollup.ts
@@ -14,10 +14,10 @@ const Validations = buildValidations({
 
 export default class RollupFragment extends Fragment.extend(Validations) implements Rollup {
   @attr({ defaultValue: () => [] })
-  columns!: Rollup['columns'];
+  declare columns: Rollup['columns'];
 
   @attr('boolean', { defaultValue: false })
-  grandTotal!: Rollup['grandTotal'];
+  declare grandTotal: Rollup['grandTotal'];
 }
 
 declare module 'navi-core/models/registry' {
--- a/packages/core/addon/models/bard-request-v2/fragments/rollup.ts
+++ b/packages/core/addon/models/bard-request-v2/fragments/rollup.ts
@@ -1,0 +1,27 @@
+/**
+ * Copyright 2021, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import { attr } from '@ember-data/model';
+import Fragment from 'ember-data-model-fragments/fragment';
+import { Rollup } from 'navi-data/adapters/facts/interface';
+import { validator, buildValidations } from 'ember-cp-validations';
+
+const Validations = buildValidations({
+  columns: [validator('has-many')],
+  grandTotal: validator('presence', true),
+});
+
+export default class RollupFragment extends Fragment.extend(Validations) implements Rollup {
+  @attr({ defaultValue: () => [] })
+  columns!: Rollup['columns'];
+
+  @attr('boolean', { defaultValue: false })
+  grandTotal!: Rollup['grandTotal'];
+}
+
+declare module 'navi-core/models/registry' {
+  export interface FragmentRegistry {
+    'bard-request-v2/fragments/rollup': RollupFragment;
+  }
+}
--- a/packages/core/addon/models/bard-request-v2/request.ts
+++ b/packages/core/addon/models/bard-request-v2/request.ts
@@ -356,6 +356,8 @@ export default class RequestFragment extends Fragment.extend(Validations) implem
    * @param column - dimension column to add a rollup
    */
   pushRollupColumn(column: ColumnFragment) {
+    const columnExists = this.columns.toArray().some((requestColumn) => requestColumn.cid === column.cid);
+    assert(`Rollup column with cid: ${column.cid} must exist in request`, columnExists);
     this.removeRollupColumn(column); //remove duplicate
     this.rollup.columns.push(column.cid);
   }
--- a/packages/core/addon/models/bard-request-v2/request.ts
+++ b/packages/core/addon/models/bard-request-v2/request.ts
@@ -7,7 +7,7 @@ import { inject as service } from '@ember/service';
 import { attr } from '@ember-data/model';
 import Fragment from 'ember-data-model-fragments/fragment';
 //@ts-ignore
-import { fragmentArray } from 'ember-data-model-fragments/attributes';
+import { fragmentArray, fragment } from 'ember-data-model-fragments/attributes';
 import { validator, buildValidations } from 'ember-cp-validations';
 //@ts-ignore
 import { isEqual } from 'lodash-es';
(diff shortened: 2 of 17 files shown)


packages/core/addon/models/bard-request-v2/fragments/rollup.ts
```diff
@@ -0,0 +1,27 @@
...
+  columns: [validator('has-many')],
+  grandTotal: validator('presence', true),
+});
+
+export default class RollupFragment extends Fragment.extend(Validations) implements Rollup {
+  @attr({ defaultValue: () => [] })
+  columns!: Rollup['columns'];
+
+  @attr('boolean', { defaultValue: false })
+  grandTotal!: Rollup['grandTotal'];
```
kevinhinterlong, line 20:
```suggestion
  declare columns: Rollup['columns'];

  @attr('boolean', { defaultValue: false })
  declare grandTotal: Rollup['grandTotal'];
```

packages/reports/addon/consumers/request/rollup.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright 2021, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import { inject as service } from '@ember/service';
+import ActionConsumer from 'navi-core/consumers/action-consumer';
+import Route from '@ember/routing/route';
+import RequestActionDispatcher, { RequestActions } from 'navi-reports/services/request-action-dispatcher';
+import ColumnFragment from 'navi-core/addon/models/bard-request-v2/fragments/column';
+import ReportModel from 'navi-core/models/report';
```
kevinhinterlong, line 10: Check for other `/addon/` in imports

```suggestion
import RequestActionDispatcher, { RequestActions } from 'navi-reports/services/request-action-dispatcher';
import type Route from '@ember/routing/route';
import type ColumnFragment from 'navi-core/models/bard-request-v2/fragments/column';
import type ReportModel from 'navi-core/models/report';
```

packages/reports/tests/unit/consumers/request/rollup-test.ts
```diff
@@ -0,0 +1,113 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+// @ts-ignore
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { RequestActions } from 'navi-reports/services/request-action-dispatcher';
+import { run } from '@ember/runloop';
+import StoreService from '@ember-data/store';
+import RollupConsumer from 'navi-reports/consumers/request/rollup';
+import RequestFragment from 'navi-core/models/bard-request-v2/request';
+import { Column } from 'navi-data/adapters/facts/interface';
```
kevinhinterlong, line 10:
```suggestion
import type StoreService from '@ember-data/store';
import type RollupConsumer from 'navi-reports/consumers/request/rollup';
import type RequestFragment from 'navi-core/models/bard-request-v2/request';
import type { Column } from 'navi-data/adapters/facts/interface';
```

packages/core/tests/unit/services/compression-test.ts
```diff
@@ -59,8 +59,8 @@ module('Unit | Service | compression', function (hooks) {
     const compressedString = await Service.compressModel(report);
 
     assert.equal(
-      compressedString,
-      encodeURIComponent(compressedString),
+      compressedString.replace(/[+/=]/g, ''),
+      encodeURIComponent(compressedString.replace(/[+/=]/g, '')),
```
kevinhinterlong, line 63: Maybe we should update the service to return `encodeURIComponent(str)`?

kevinhinterlong (approved):

packages/reports/addon/consumers/request/rollup.ts
```diff
@@ -4,10 +4,10 @@
  */
 import { inject as service } from '@ember/service';
 import ActionConsumer from 'navi-core/consumers/action-consumer';
-import Route from '@ember/routing/route';
 import RequestActionDispatcher, { RequestActions } from 'navi-reports/services/request-action-dispatcher';
-import ColumnFragment from 'navi-core/addon/models/bard-request-v2/fragments/column';
-import ReportModel from 'navi-core/models/report';
+import type Route from '@ember/routing/route';
+import type ColumnFragment from 'navi-core/addon/models/bard-request-v2/fragments/column';
```
kevinhinterlong, line 9: `/addon/` import

packages/core/tests/unit/services/compression-test.ts
cythrawll: I'm not sure what that would break tbh, afraid of double encoding.
this was just a bad test that assumed all base64 characters wouldn't get uri encoded. probably should just be removed.

cythrawll: So looks like lz-string actually does do something like this already, it just doesn't use `encodeURIComponent` and has it's own safe list of uri characters that includes `+ and -`

So switched to `encodeURI` in tests instead to better match what the lz-string lib actually does.

packages/core/addon/models/bard-request-v2/request.ts
```diff
@@ -347,6 +351,31 @@ export default class RequestFragment extends Fragment.extend(Validations) implem
     this.sorts.pushObject(this.fragmentFactory.createSort(type, source, field, parameters, direction));
   }
 
+  /**
+   * Pushes column cid to end of rollup list, removing dupes.
+   * @param column - dimension column to add a rollup
+   */
+  pushRollupColumn(column: ColumnFragment) {
```
jkusa, line 358: What if the `cid` doesn't exist in columns?

cythrawll: typescript error.
or you suggesting we should generate one here if it's empty?

packages/reports/addon/consumers/request/column.ts
```diff
@@ -59,6 +59,7 @@ export default class ColumnConsumer extends ActionConsumer {
       const { routeName } = route;
       const { request } = route.modelFor(routeName) as ReportModel;
       request.removeColumn(column);
+      this.requestActionDispatcher.dispatch(RequestActions.REMOVE_ROLLUP_COLUMN, route, column);
```
jkusa, line 62: Can we move this to a handler in `consumers/request/rollup.ts` that listens to `[RequestActions.REMOVE_COLUMN_FRAGMENT]` as done in the past:

https://github.com/yavin-dev/framework/blob/23a6258192842bd01d5507574267c9a2b5170c77/packages/reports/addon/consumers/request/sort.ts#L92

packages/core/addon/models/bard-request-v2/request.ts
jkusa: I was thinking along the lines of asserting the column fragment exists in the `columns`  array

packages/reports/addon/consumers/request/rollup.ts
```diff
@@ -0,0 +1,50 @@
...
+import RequestActionDispatcher, { RequestActions } from 'navi-reports/services/request-action-dispatcher';
+import type Route from '@ember/routing/route';
+import type ColumnFragment from 'navi-core/models/bard-request-v2/fragments/column';
+import type ReportModel from 'navi-core/models/report';
+
+export default class RollupConsumer extends ActionConsumer {
+  @service
+  declare requestActionDispatcher: RequestActionDispatcher;
+
+  actions = {
```
jkusa, line 16: How will handle reorder? Will that come in another PR?

cythrawll: the order of columns does not effect rollups. it's always the order the person pushes the rollup buttons.

jkusa: I believe we decided that behavior was sufficient for the alpha branch, but wanted to utilize the column reorder feature for rollup ordering in beta

cythrawll: We are only adding behavior to equal to alpha branch for now.

packages/core/addon/models/bard-request-v2/request.ts
cythrawll: we could add that assertion yes.

jkusa (approved):

PR #1455, opened 2021-08-26
New: Subtotal bard adapter.

cythrawll:
## Description

Port of subtotal adapter from alpha branch

## Proposed Changes

- Since reqv2 uses CIDs rather than dimensions, we have to normalize to dimension names in the adapter.

## License

### DIFF
--- a/packages/data/addon/adapters/facts/bard.ts
+++ b/packages/data/addon/adapters/facts/bard.ts
@@ -267,6 +267,27 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
   }
 
   /**
+   * Builds rollup query params
+   * @param {RequestV1} request
+   * @return {String}
+   */
+  _buildRollupParam(request: RequestV2): string {
+    const columns = request.columns;
+    const rollupColumns = request.rollup?.columns || [];
+    return rollupColumns.length
+      ? array(
+          rollupColumns.map((rollupColumn) => {
+            const column = columns.find((col) => rollupColumn === col.cid);
+            return column && isDateTime(column) ? 'dateTime' : column?.field;
+          })
+        )
+          .uniq()
+          .filter((x) => !!x)
+          .join(',')
+      : '';
+  }
+
+  /**
    * Builds a URL path for a request
    */
   _buildURLPath(request: RequestV2, options: RequestOptions = {}): string {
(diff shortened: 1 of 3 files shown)


packages/data/addon/adapters/facts/bard.ts
```diff
@@ -266,6 +266,27 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
...
+  /**
+   * Builds rollup query params
+   * @param {RequestV1} request
+   * @return {String}
+   */
+  _buildRollupParam(request: RequestV2): string {
+    const columns = request.columns;
+    const rollupColumns = request.rollup?.columns || [];
+    return rollupColumns.length
+      ? array(
```
kevinhinterlong, line 278: Can we avoid ember array and make cid -> field map
```js
const cidToField = request.columns.reduce((cidToField, column) => {
  const field = column && isDateTime(column) ? 'dateTime' : column?.field;
  cidToField.set(column.cid, field);
  return cidToField
}, new Map());
```

kevinhinterlong (commented):
small stuff

packages/data/addon/adapters/facts/bard.ts
```diff
@@ -266,6 +266,24 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
...
+   * Builds rollup query params
+   * @param {RequestV1} request
+   * @return {String}
+   */
+  _buildRollupParam(request: RequestV2): string {
+    const columns = request.columns;
+    const rollupColumnCids = request.rollup?.columns || [];
+    return [
+      ...columns.reduce((colSet, requestColumn) => {
+        if (rollupColumnCids.includes(requestColumn.cid ?? '')) {
```
daniellevinson, line 279: can `rollupColumnCids` actually include `''` or is it just to return false?

packages/data/addon/adapters/facts/bard.ts
```diff
@@ -266,6 +266,24 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
     return undefined;
   }
 
+  /**
+   * Builds rollup query params
+   * @param {RequestV1} request
```
daniellevinson, line 271:
```suggestion
   * @param {RequestV2} request
```

daniellevinson (approved):

packages/data/addon/adapters/facts/bard.ts
cythrawll: just to return false.

PR #1461, opened 2021-09-07
Fix: Base-json-serializer javascript

100starnight:


daniellevinson:
I think @kevinhinterlong has this change in #1460

PR #1469, opened 2021-09-11
Update: Dimension value context layout

jkusa:
Resolves #

## Proposed Changes

-

## License


packages/reports/addon/components/filter-values/dimension-select.hbs
```diff
@@ -26,7 +26,10 @@
       @dropdownClass="filter-values--dimension-select__dropdown"
       @triggerClass="is-medium filter-values--dimension-select__trigger {{if @filter.validations.attrs.values.isInvalid "filter-values--dimension-select--error"}}"
       as |dim|>
-        <span title={{dim.displayValue}}>{{dim.displayValue}}</span>
+        <div class="filter-values--dimension-select__option-value" title={{dim.displayValue}}>{{dim.displayValue}}</div>
+        {{#each-in dim.suggestions as |col val|}}
```
cythrawll, line 30: `|field value|`

kevinhinterlong (approved):

packages/data/addon/models/navi-dimension.ts
```diff
@@ -30,17 +30,13 @@ export default class NaviDimensionModel extends NativeWithCreate {
...
   /**
    * Dimension value for display purposes
    */
   get displayValue() {
-    const { value, suggestions } = this;
-    if (suggestions?.length) {
-      return `${value} (${suggestions.join(', ')})`;
-    }
-    return `${value}`;
+    return `${this.value}`;
```
kevinhinterlong, line 39: The dimension-select component expects this to contain the suggestions so that it performs the local search correctly. Can we try to add a test for that to show searching by suggestion doesn't work for small cardinality and then add a fix there to explicitly search the suggestions

kevinhinterlong (approved):

PR #1489, opened 2021-10-04
Update: Configurable Fili end for since operator

cythrawll:
## Description

Default end date for dateTime `since` operator for fili was  9999, this actually causes an issue with fili to search way far out into the future for date buckets that don't exist.

## Proposed Changes

- Allow a configurable value per datasource to control how far out `since` will search for.

## License

### DIFF
--- a/packages/data/addon/adapters/facts/bard.ts
+++ b/packages/data/addon/adapters/facts/bard.ts
@@ -10,7 +10,7 @@ import { inject as service } from '@ember/service';
 import { A as array } from '@ember/array';
 import EmberObject from '@ember/object';
 import { canonicalizeMetric } from '../../utils/metric';
-import { configHost } from '../../utils/adapter';
+import { configHost, getDataSource } from '../../utils/adapter';
 import NaviFactAdapter, {
   Filter,
   RequestOptions,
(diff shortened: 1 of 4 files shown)


packages/data/addon/adapters/facts/bard.ts
```diff
@@ -180,7 +181,9 @@ export default class BardFactsAdapter extends EmberObject implements NaviFactAda
       if (!moment.utc(start).isValid()) {
         throw new FactAdapterError(`Since operator only supports datetimes, '${start}' is invalid`);
       }
-      end = moment.utc('9999-12-31').startOf(filterGrain).toISOString();
+      const dataSourceConfig = getDataSource(request.dataSource) as FiliDataSource;
+      const sinceOperatorEnd = dataSourceConfig.options?.sinceOperatorEndPeriod;
```
kevinhinterlong, line 185: This will cause issues because of the grain attached to the period. Say we have a week grain for the query, then no period of days/months/years will match up, only something like `P4W` would. Maybe we can do something similar to the next line where we cut it off to the closest grain that aligns

kevinhinterlong (approved):

daniellevinson (approved):

jkusa (approved):

PR #1506, opened 2021-10-25
Fix: Don't client side sort dim search endpoint

cythrawll:
## Description

Lucene based dimension searching endpoint has better result sorting than client side provides. So in the case of that dimension sort we do not want to resort the result

## Proposed Changes

- Do not client side sort if cardinality is larger than "Small" and we're using the enhanced dimension search.
- TODO: add caching to `all` in the adapter will be in a followup PR.

## License

### DIFF
--- a/packages/data/addon/adapters/dimensions/bard.ts
+++ b/packages/data/addon/adapters/dimensions/bard.ts
@@ -91,6 +91,14 @@ export default class BardDimensionAdapter extends EmberObject implements NaviDim
     return andQueries.length ? { filters: serializeFilters(requestV2Filters, dimension.columnMetadata.source) } : {};
   }
 
+  _searchDimensions(dimensions: FiliDimensionResponse, query: string): FiliDimensionResponse {
+    const results = SearchUtils.searchDimensionRecords(arr(dimensions.rows), query, 5000, 1) as SearchUtilResult[];
+    return {
+      rows: results.map((result: SearchUtilResult) => result.record),
+      ...(dimensions.meta ? { meta: dimensions.meta } : {}),
+    };
+  }
+
   _find(
     url: string,
     data: Record<string, string | number | boolean>,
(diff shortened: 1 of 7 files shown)


packages/data/addon/adapters/dimensions/bard.ts
```diff
@@ -39,6 +42,11 @@ export type FiliDimensionResponse = {
   meta?: Record<string, unknown>;
 };
 
+type SearchUtilResult = {
+  relevance: number;
+  record: Record<string, string>;
```
jkusa, line 47:
```suggestion
  record: ResponseRow;
```
??

jkusa, line 47:
```suggestion
  record: ResponseRow;
```
??

packages/data/addon/utils/search.ts
```diff
@@ -74,20 +75,25 @@ export default {
...
 
     // Filter, map, and sort records based on how close each record is to the search query
     for (let i = 0; i < records.length; i++) {
-      record = records.objectAt(i);
+      record = records.objectAt(i) as Record<string, string>;
 
       // Determine relevance based on string match weight
       let descriptionMatchWeight = this.getPartialMatchWeight(
-          (record.description || '').toLowerCase(),
+          (record?.description || '').toLowerCase(),
```
jkusa, line 93: Look like we only are only considering `id` and `description` fields. We need search against the dimension value plus any `suggestion` fields.

jkusa: @cythrawll here is an example of the suggstionColumns: https://github.com/yavin-dev/framework/blob/master/packages/reports/addon/components/filter-values/dimension-select.ts#L131

cythrawll: Since we don't really have access to that in the search utl, changed so it weights all fields returned by the dimension search.

packages/data/addon/utils/search.ts
```diff
@@ -74,25 +75,31 @@ export default {
...
 
     // Filter, map, and sort records based on how close each record is to the search query
     for (let i = 0; i < records.length; i++) {
-      record = records.objectAt(i);
+      record = records.objectAt(i) as Record<string, string>;
 
       // Determine relevance based on string match weight
-      let descriptionMatchWeight = this.getPartialMatchWeight(
-          (record.description || '').toLowerCase(),
+      const { id, ...rest } = record;
```
kevinhinterlong, line 92: Even in just the fili context `id` does not have always exist. Can we maybe pass `'id', ['desc', 'other']` as parameters based on the columnMetadata?

cythrawll: Unfortunately we don't really have access to meta data here, we'll need to probably come up with an enhanced version later.
But this code does handle the fact some dimensions don't have 'id' It will get picked up and weighted with the ...rest fields and id field weighting will be ignored.

kevinhinterlong (approved):

jkusa (approved):

PR #1508, opened 2021-10-29
Update: Added file attribute for dropdown export

utsabpoud:
## Description
Added attribute_name for csv download in mult-format-export

## License

### DIFF
--- a/packages/reports/addon/components/report-actions/export.ts
+++ b/packages/reports/addon/components/report-actions/export.ts
@@ -13,6 +13,7 @@ import type ReportModel from 'navi-core/models/report';
 import { RequestV2 } from 'navi-data/adapters/facts/interface';
 import Ember from 'ember';
 import { dasherize } from '@ember/string';
+import moment from 'moment';
 
 interface Args {
   model: ReportModel;
@@ -43,7 +44,8 @@ export default class ReportActionExport extends Component<Args> {
    * filename for the downloaded file
    */
   get filename() {
-    return this.args.model.title;
+    const dateString = moment().format('YYYYMMDDTHHmmss');
+    return `${this.args.model.title}-${dateString}.csv`;
   }
 
   /**
@@ -59,9 +61,10 @@ export default class ReportActionExport extends Component<Args> {
     const serializedRequest = this.args.model.request.serialize() as RequestV2;
 
     try {
-      const url: string = yield taskFor(this.facts.getDownloadURL).perform(serializedRequest, {
+      let url: string = yield taskFor(this.facts.getDownloadURL).perform(serializedRequest, {
         format: 'csv',
         dataSourceName: serializedRequest.dataSource,
+        fileName: this.filename,
       });
       this.downloadURLLink(url);
     } catch (e) {
--- a/packages/reports/addon/components/report-actions/export.ts
+++ b/packages/reports/addon/components/report-actions/export.ts
@@ -45,7 +45,7 @@ export default class ReportActionExport extends Component<Args> {
    */
   get filename() {
     const dateString = moment().format('YYYYMMDDTHHmmss');
-    return `${this.args.model.title}-${dateString}.csv`;
+    return `${this.args.model.title}-${dateString}`;
   }
 
   /**
@@ -64,7 +64,7 @@ export default class ReportActionExport extends Component<Args> {
       let url: string = yield taskFor(this.facts.getDownloadURL).perform(serializedRequest, {
         format: 'csv',
         dataSourceName: serializedRequest.dataSource,
-        fileName: this.filename,
+        fileName: `${this.filename}.csv`,
       });
       this.downloadURLLink(url);
     } catch (e) {
--- a/packages/reports/addon/components/report-actions/export.ts
+++ b/packages/reports/addon/components/report-actions/export.ts
@@ -41,7 +41,7 @@ export default class ReportActionExport extends Component<Args> {
   }
 
   /**
-   * filename for the downloaded file
+   * filename along with the format for the downloaded file
    */
   get filename() {
     const dateString = moment().format('YYYYMMDDTHHmmss');
--- a/packages/reports/addon/components/report-actions/multiple-format-export.ts
+++ b/packages/reports/addon/components/report-actions/multiple-format-export.ts
@@ -21,6 +21,10 @@ import { taskFor } from 'ember-concurrency-ts';
 import type ReportModel from 'navi-core/models/report';
 import type GsheetExportService from 'navi-core/services/gsheet-export';
 import { htmlSafe } from '@ember/template';
+import { RequestV2 } from 'navi-data/addon/adapters/facts/interface';
+import { getDataSource } from 'navi-data/utils/adapter';
+import { dasherize } from '@ember/string';
+import moment from 'moment';
 
 export default class MultipleFormatExport extends ReportActionExport {
   /**
@@ -122,7 +126,14 @@ export default class MultipleFormatExport extends ReportActionExport {
 
     try {
       if (exportType === 'CSV') {
-        yield taskFor(super.downloadTask).perform();
+        const serializedRequest = this.args.model.request.serialize() as RequestV2;
+        let url = yield taskFor(super.downloadTask).perform();
+        if (getDataSource(serializedRequest.dataSource).type === 'bard') {
+          const dateString = moment().format('YYYYMMDDTHHmmss');
+          const filename = `${dasherize(this.args.model.title)}_${dateString}.csv`;
+          url = `${url}&filename=${filename}`;
+        }
+        this.downloadURLLink(url);
       } else if (exportType === 'PDF' || exportType === 'PNG') {
         let url: string = yield this.exportHref;
         if (exportType === 'PNG') {
--- a/packages/reports/addon/components/report-actions/multiple-format-export.ts
+++ b/packages/reports/addon/components/report-actions/multiple-format-export.ts
@@ -119,6 +119,7 @@ export default class MultipleFormatExport extends ReportActionExport {
 
   /**
    * @override
+   * Added datastring format in filename
    */
   get filename(): string {
     const dateString = moment().format('YYYYMMDDTHHmmss');
--- a/packages/reports/addon/components/report-actions/multiple-format-export.ts
+++ b/packages/reports/addon/components/report-actions/multiple-format-export.ts
@@ -23,7 +23,6 @@ import type GsheetExportService from 'navi-core/services/gsheet-export';
 import { htmlSafe } from '@ember/template';
 import { RequestV2 } from 'navi-data/addon/adapters/facts/interface';
 import { getDataSource } from 'navi-data/utils/adapter';
-import { dasherize } from '@ember/string';
 import moment from 'moment';
 
 export default class MultipleFormatExport extends ReportActionExport {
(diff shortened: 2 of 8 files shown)


packages/reports/addon/components/report-actions/multiple-format-export.ts
```diff
@@ -122,7 +134,12 @@ export default class MultipleFormatExport extends ReportActionExport {
 
     try {
       if (exportType === 'CSV') {
-        yield taskFor(super.downloadTask).perform();
+        const serializedRequest = this.args.model.request.serialize() as RequestV2;
+        let url = yield taskFor(super.downloadTask).perform();
+        if (getDataSource(serializedRequest.dataSource).type === 'bard') {
```
kevinhinterlong, line 139: We don't want to add datasource specific logic in the components, we should update the `downloadTask` task in `export.ts` to pass a `filename: ...` attribute
https://github.com/yavin-dev/framework/blob/166b2a4a4139c9575003ec0074857164a330e6eb/packages/reports/addon/components/report-actions/export.ts#L62-L65
then update the options type of `getDownloadURL` to allow for `filename: string` so that each adapter can handle it
https://github.com/yavin-dev/framework/blob/e167311615defcf91944df3a9699a8b454c9f870/packages/data/addon/services/navi-facts.ts#L77-L81

utsabpoud: Modified the logic into export.ts, I have added fileName in optionsType.

packages/reports/tests/integration/components/report-actions/multiple-format-export-test.js
```diff
@@ -118,7 +119,7 @@ module('Integration | Component | report actions - multiple-format-export', func
     await click($('.menu-content a:contains("CSV")')[0]);
     assert.equal(
       find('.export__download-link').getAttribute('download'),
```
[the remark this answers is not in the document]
utsabpoud: I have updated with test case wit href attribute.

packages/reports/addon/components/report-actions/multiple-format-export.ts
kevinhinterlong: Did you forget to push your changes?

packages/reports/addon/components/report-actions/export.ts
```diff
@@ -43,7 +44,8 @@ export default class ReportActionExport extends Component<Args> {
    * filename for the downloaded file
    */
   get filename() {
-    return this.args.model.title;
+    const dateString = moment().format('YYYYMMDDTHHmmss');
+    return `${this.args.model.title}-${dateString}.csv`;
```
daniellevinson, line 48: since `filename` might be used for other formats as well, can we remove the `.csv` here and just add it in `downloadTask`?

packages/reports/tests/integration/components/report-actions/multiple-format-export-test.js
```diff
@@ -43,14 +44,17 @@ module('Integration | Component | report actions - multiple-format-export', func
...
+    const expectedHref = factService.getURL(this.report.request.serialize(), {
+      format: 'csv',
+      fileName: `Hyrule News-${dateString}.csv`,
+    });
     await triggerEvent('.menu-trigger', 'mouseenter');
     await click($('.menu-content a:contains("CSV")')[0]);
     assert
       .dom('.export__download-link')
-      .hasAttribute('href', expectedHref, 'The href attribute is set correctly for CSV');
+      .hasAttribute('href'.trim(), expectedHref, 'The href attribute is set correctly for CSV');
```
daniellevinson, line 57: why do we need the `trim()` here?

packages/reports/tests/acceptance/multi-datasource-test.js
```diff
@@ -186,13 +186,11 @@ module('Acceptance | multi-datasource report builder', function (hooks) {
...
-      .hasAttribute(
-        'href',
-        'https://data2.naviapp.io/v1/data/inventory/day/container;show=desc/displayCurrency;show=desc/?dateTime=P3D%2Fcurrent&metrics=usedAmount%2Crevenue(currency%3DGIL)&filters=container%7Cid-in%5B%222%22%5D&having=usedAmount-gt%5B50%5D&format=csv',
-        'uses csv export from right datasource'
-      );
+    const csv = document.querySelector('.export__download-link')?.getAttribute('href') ?? '';
+    assert.ok(
+      /(?=.*?data2.naviapp.io)(?=.*format=csv)(?=.*dateTime=)(?=.*csv&filename=)/.test(csv),
+      'csv export url is generated for the report'
+    );
```
daniellevinson, line 193: do we test the filename itself or only that it exists?
to be more specific we can do `&filename=the_filename-\d\d\d\d-\d\d-\d\d` or whatever the date format is

also, would this work? (here and in the other asserts below)

```suggestion
    assert
      .dom('.export__download-link')
      .hasAttribute(
        'href',
        /(?=.*?data2.naviapp.io)(?=.*format=csv)(?=.*dateTime=)(?=.*csv&filename=)/,
        'csv export url is generated for the report'
      );
```

packages/reports/addon/components/report-actions/export.ts
```diff
@@ -59,9 +61,10 @@ export default class ReportActionExport extends Component<Args> {
     const serializedRequest = this.args.model.request.serialize() as RequestV2;
 
     try {
-      const url: string = yield taskFor(this.facts.getDownloadURL).perform(serializedRequest, {
+      let url: string = yield taskFor(this.facts.getDownloadURL).perform(serializedRequest, {
```
daniellevinson, line 64: not sure if we want to pass it to `this.facts.getDownloadURL` or just do
```this.downloadURLLink(`${url}&filename=${this.filename}`)```
I'm fine with either

Comment (utsabpoud) [reply packages/reports/addon/components/report-actions/export.ts:64]: I kept in filename becoz i thought that we can have common across everywhere we can pass it as attribute in stead of appending in the url

File: screwdriver.yaml [hunk 7]
```diff
@@ -9,7 +9,7 @@ shared:
     steps:
       - .: &install-node
           install-node: |
-            sd-cmd exec screwdriver/install-nodejs@latest $NODE_VERSION
+            sd-cmd exec screwdriver/install-nodejs@1.0.4 $NODE_VERSION
```
kevinhinterlong, line 12: 1.0.4 is latest according to https://cd.screwdriver.cd/commands/screwdriver/install-nodejs
```suggestion
            sd-cmd exec screwdriver/install-nodejs@latest $NODE_VERSION
```

utsabpoud: Updated

packages/reports/tests/acceptance/multi-datasource-test.js
utsabpoud: Added date format in regex

packages/reports/tests/integration/components/report-actions/multiple-format-export-test.js
```diff
@@ -109,18 +113,20 @@ module('Integration | Component | report actions - multiple-format-export', func
...
-      'The download attribute is set correctly'
-    );
+
+    assert
+      .dom('.export__download-link')
+      .hasAttribute(
+        'href',
+        /(?=.*?data.naviapp.io)(?=.*format=csv)(?=.*dateTime=)(?=.*csv&filename=Hyrule%20News-\d\d\d\d\d\d\d*T\d\d\d\d\d\d)/,
+        'csv export url is generated for the report'
+      );
```
daniellevinson, line 129: We test the url in the previous test, this test is for the `download` attr so maybe it should be changed to something like 
```suggestion
    assert.ok(
      /hyrule-news-\d\d\d\d\d\d\d*T\d\d\d\d\d\d)/.test(find('.export__download-link').getAttribute('download')),
      'The download attribute is set correctly'
    );
```

packages/reports/addon/components/report-actions/export.ts
daniellevinson:
```suggestion
      const url: string = yield taskFor(this.facts.getDownloadURL).perform(serializedRequest, {
```

packages/reports/tests/acceptance/multi-datasource-test.js
```diff
@@ -190,8 +190,8 @@ module('Acceptance | multi-datasource report builder', function (hooks) {
       .dom('.export__download-link')
       .hasAttribute(
         'href',
-        'https://data2.naviapp.io/v1/data/inventory/day/container;show=desc/displayCurrency;show=desc/?dateTime=P3D%2Fcurrent&metrics=usedAmount%2Crevenue(currency%3DGIL)&filters=container%7Cid-in%5B%222%22%5D&having=usedAmount-gt%5B50%5D&format=csv',
-        'uses csv export from right datasource'
+        /(?=.*?data2.naviapp.io)(?=.*format=csv)(?=.*dateTime=)(?=.*csv&filename=Report%20for%20different%20datasource-\d\d\d\d\d\d\d*T\d\d\d\d\d\d)/,
```
daniellevinson, line 193: can we still test that the entire url is correct regardless of the filename (`https://data2.naviapp.io/v1/data/inventory/day/container;show=desc/displayCurrency;show=desc/?dateTime=P3D%2Fcurrent&metrics=usedAmount%2Crevenue(currency%3DGIL)&filters=container%7Cid-in%5B%222%22%5D&having=usedAmount-gt%5B50%5D&format=csv`) and just add the filename?

packages/reports/tests/integration/components/report-actions/export-test.js
```diff
@@ -56,11 +64,13 @@ module('Integration | Component | report actions - export', function (hooks) {
...
-      'hyrule-news',
-      'The download attribute is set correctly'
-    );
+    assert
+      .dom('.export__download-link')
+      .hasAttribute(
+        'href',
+        /(?=.*?data.naviapp.io)(?=.*format=csv)(?=.*dateTime=)(?=.*csv&filename=Hyrule%20News-\d\d\d\d\d\d\d*T\d\d\d\d\d\d)/,
+        'csv export url is generated for the report'
+      );
```
daniellevinson, line 73: we test the url in the previous test, this is for the `download` attr so maybe it should be something like 
```suggestion
    assert
      .dom('.export__download-link')
      .hasAttribute(
        'download',
        /hyrule-news-\d\d\d\d\d\d\d*T\d\d\d\d\d\d)/,
        'The download attribute is set correctly'
      );
```

packages/reports/tests/integration/components/report-actions/multiple-format-export-test.js
```diff
@@ -43,9 +44,12 @@ module('Integration | Component | report actions - multiple-format-export', func
...
     await render(TEMPLATE);
 
+    const dateString = moment().format('YYYYMMDDTHHmmss');
     const factService = this.owner.lookup('service:navi-facts');
-    const expectedHref = factService.getURL(this.report.request.serialize(), { format: 'csv' });
-
+    const expectedHref = factService.getURL(this.report.request.serialize(), {
+      format: 'csv',
+      fileName: `Hyrule News-${dateString}.csv`,
+    });
```
daniellevinson, line 52: can we do the regex test here like in export-test.js?

packages/reports/tests/integration/components/report-actions/export-test.js
utsabpoud: Before I have used filename attribute, So because of Url Encoding Special Characters getting added. Now i have appended in the url itself instead of passing into the attribute.

In Previous also we have tested by appending into the url. Let me change into appending into the url in` export.ts`

utsabpoud: let me know if I can go with filename attribute and url appending.

packages/reports/tests/acceptance/multi-datasource-test.js
utsabpoud: Since we are having the date format I have added regex

cythrawll (approved):

daniellevinson (approved):

PR #1509, opened 2021-10-29
Elide namespace support

daniellevinson:
## Description

Add support for Elide namespaces https://elide.io/pages/guide/v6/04-analytics.html#namespaces

#### WS:
- add optional `namespaces` list to the `DataSource` model
- add a namespace to datasources config
- new Hjson config file for `DemoNamespace` and a new table `TrendingNow` that belongs to that namespace (same physical base table)

#### UI:
- Add a namespace filter to the GraphQL metadata queries `GetTables` and `GetTable`
- add the namespace connection to the schema and table fragment
- normalizer: use `friendlyName` as `name`
- Show an ellipsis at the front of the column id so table namespace prefix is hidden

## Screenshots

Screen Shot 2021-11-01 at 11 10 18 AM

Screen Shot 2021-11-01 at 11 10 11 AM

Screen Shot 2021-11-01 at 11 10 33 AM

## License


kevinhinterlong (commented):
It would be nice to have more of a service around dataSources instead of relying directly on `config.navi.dataSources`. Then we could do stuff like fetch all the namespaces from an elide datasource automatically rather than configuring each namespace manually

kevinhinterlong (approved):

packages/app/config/environments.json
```diff
@@ -23,6 +23,14 @@
         "uri": "http://localhost:8080/graphql/api/v1",
         "type": "elide",
         "suggestedDataTables": []
+      },
```
jkusa, line 26: Can we model namespaces like this:

```json
      {
        "name": "corpAnalytics",
        "displayName": "Corp Analytics - General",
        "uri": "http://localhost:8080/graphql/api/v1",
        "type": "elide",
        "suggestedDataTables": [] // for default namespace
        "namespaces": [{
          "name": "rev",
          "displayName": "Revenue",
          "suggestedDataTables": ['allRevenue']
        }]        
      },
```

Then the datasource id for namespaces can be something like:
- `corpAnalytics.default` 
- `corpAnalytics.rev`

daniellevinson:
> It would be nice to have more of a service around dataSources instead of relying directly on `config.navi.dataSources`. Then we could do stuff like fetch all the namespaces from an elide datasource automatically rather than configuring each namespace manually

Yeah @jkusa had similar feedback and said we can address that in a separate PR. I thought we wanted to explicitly configure which namespaces we want to use.

packages/data/addon/utils/adapter.ts
```diff
@@ -20,11 +21,25 @@ export function getDataSource<T extends keyof DataSourceRegistry>(dataSourceName
...
-    return dataSource as DataSourceRegistry[T];
+  const [lookupName, lookupNamespace] = dataSourceName.split('.');
+  const dataSource = dataSources.find(({ name }) => name === lookupName);
+  if (!dataSource) {
+    throw new Error(`Datasource "${lookupName}" should be configured in the navi environment`);
   }
-  throw new Error(`Datasource ${dataSourceName} should be configured in the navi environment`);
+  if (lookupNamespace) {
+    const namespace = dataSource.namespaces?.find(({ name }) => name === lookupNamespace);
+    if (!namespace) {
```
jkusa, line 31: Do we store `elideOne.default` as a data source?

daniellevinson: no, just `elideOne`

packages/data/config/navi-config.d.ts
```diff
@@ -1,12 +1,16 @@
...
     displayName: string;
     description?: string;
+  };
+
+  export type BaseDataSource<Type, Options = void> = DataSourceNamespace & {
+    type: Type;
     uri: string;
     options?: Options;
+    namespaces?: DataSourceNamespace[];
     suggestedDataTables?: string[];
```
kevinhinterlong, line 14: Maybe move `suggestedDataTables` to `DataSourceNamespace`?

kevinhinterlong (commented):
Just the one comment

packages/reports/addon/components/report-builder/sidebar.ts
```diff
@@ -120,14 +120,36 @@ export default class ReportBuilderSidebar extends Component<Args> {
...
+
+      fetchNamespacesTasks.push(taskFor(this.fetchDataSourceNamespaces).perform(dataSource));
+    });
+
+    const namespaces = (yield all(fetchNamespacesTasks)).flat();
+    sources = [...sources, ...namespaces];
+    return yield sortBy(sources, ['name']);
+  }
+
+  @task *fetchDataSourceNamespaces(dataSource: NaviDataSource): TaskGenerator<SourceItem<NaviDataSource>[]> {
```
jkusa, line 141: Long term it would be good to have this layer not know about namespaces. We can just process the original data sources with nested namespaces and generate a single flat list of `data sources`.

packages/data/config/navi-config.d.ts
```diff
@@ -1,12 +1,16 @@
 import { Grain } from 'navi-data/utils/date';
 declare module 'navi-config' {
-  export type BaseDataSource<Type, Options = void> = {
-    type: Type;
+  export type DataSourceNamespace = {
     name: string;
     displayName: string;
     description?: string;
```
jkusa, line 6:
```suggestion
    description?: string;
    suggestedDataTables?: string[];
```

packages/data/config/navi-config.d.ts
jkusa:
```suggestion
```

jkusa (approved):
Just the one comment otherwise 👍

packages/data/config/navi-config.d.ts
daniellevinson: That's what I did initially, I just bumped it to the root level because in the future we're not going to have `namespaces` at all and we'll have to somehow define all suggested tables at the root level
But yeah maybe for now it makes more sense to have it in `DataSourceNamespace`

packages/reports/addon/components/report-builder/sidebar.ts
daniellevinson: I agree, like you and @kevinhinterlong suggested we should maybe have a service around that to discover namespaces and flatten

Issue #1338, opened 2021-03-25
I think we should sort these alphabetically, case insensitive with `Uncategorized` at the end


jkusa:
I think we should sort these alphabetically, case insensitive with `Uncategorized` at the end

_Originally posted by @kevinhinterlong in https://github.com/yavin-dev/framework/pull/1335#r601794109_

PR #1510, opened 2021-11-01
don't show categories for columns with only one category

100starnight:
## License

### DIFF
--- a/packages/reports/addon/components/column-selector.ts
+++ b/packages/reports/addon/components/column-selector.ts
@@ -10,6 +10,7 @@ import RequestFragment from 'navi-core/models/request';
 //@ts-ignore
 import { searchRecords } from 'navi-core/utils/search';
 import ColumnMetadataModel from 'navi-data/models/metadata/column';
+import { groupBy } from 'lodash-es';
 
 interface ColumnSelectorArgs {
   title: string;
@@ -32,5 +33,10 @@ export default class ColumnSelector extends Component<ColumnSelectorArgs> {
     } = this;
 
     return query ? searchRecords(allColumns, query, 'name') : allColumns;
+  }
+
+  get isSingleCategory() {
+    const grouped = groupBy(this.args.allColumns, (row) => row.category?.split(',')[0] || `Uncategorized`);
+    return Object.keys(grouped).length === 1;
   }
 }
--- a/packages/reports/addon/components/column-selector.ts
+++ b/packages/reports/addon/components/column-selector.ts
@@ -36,7 +36,7 @@ export default class ColumnSelector extends Component<ColumnSelectorArgs> {
   }
 
   get isSingleCategory() {
-    const grouped = groupBy(this.args.allColumns, (row) => row.category?.split(',')[0] || `Uncategorized`);
+    const grouped = groupBy(this.args.allColumns, (row) => row.category?.split(',')[0]);
     return Object.keys(grouped).length === 1;
   }
 }
--- a/packages/reports/addon/components/grouped-list.hbs
+++ b/packages/reports/addon/components/grouped-list.hbs
@@ -7,22 +7,28 @@
     @bufferSize={{10}}
     @containerSelector="#{{this.guid}}"
     as |item idx|>
-    {{#if item.isGroup}}
-      <li
-        class="grouped-list__group-header"
-        role="button"
-        {{on "click" (fn this.toggleOpen item.name)}}
-        >
-        <DenaliIcon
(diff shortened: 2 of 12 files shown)


packages/reports/addon/components/grouped-list.ts
```diff
@@ -17,9 +17,19 @@ interface Args {
...
+    super(...arguments);
+    if (Object.keys(this.groupedItems).length === 1) {
+      this.isSingleCategory = true;
+    }
+  }
+
   @tracked
   groupConfigs: Record<string, boolean> = {};
 
+  isSingleCategory = false;
```
jkusa, line 31:
```suggestion
  get isSingleCategory() {
    return Object.keys(this.groupedItems).length === 1
  }
```

packages/reports/addon/components/grouped-list.ts
```diff
@@ -17,9 +17,19 @@ interface Args {
...
 
 export default class GroupedListComponent extends Component<Args> {
+  constructor() {
+    //@ts-ignore
+    super(...arguments);
+    if (Object.keys(this.groupedItems).length === 1) {
+      this.isSingleCategory = true;
+    }
+  }
+
```
jkusa, line 27:
```suggestion
```
This is better done as a getter so it can be reactive to new `groupedItems` that are passed in.

packages/reports/addon/components/grouped-list.hbs
```diff
@@ -7,22 +7,28 @@
...
-        >
-        <DenaliIcon
-          class="grouped-list__group-header-icon grouped-list__group-header-icon--{{if item.isOpen "open" "closed"}}"
-          @icon="arrowhead-{{if item.isOpen "down" "right"}}"
-        />
-        <span class="grouped-list__group-header-content">{{item.name}} ({{item.length}})</span>
-      </li>
-    {{else}}
-      <li class="grouped-list__item m-l-18 m-t-2">
+    {{#if this.isSingleCategory}}
```
jkusa, line 10: Do we need this `if` can we use lines 28-30 to cover this case with a check on `isSingleCategory` for the indent class?

packages/search/tests/unit/services/navi-search/navi-definition-search-provider-test.js
```diff
@@ -39,7 +39,7 @@ module('Unit | Service | navi-definition-search-provider', function (hooks) {
 
     const makeDateTime = (table) => ({ id: `${table}.dateTime`, name: 'Date Time' });
     const expectedResults = [
-      ...['network', 'tableA', 'tableB', 'protected', 'tableC'].map((table) => makeDateTime(table)),
+      ...['network', 'tableA', 'tableB', 'protected', 'tableC', 'tableD'].map((table) => makeDateTime(table)),
```
jkusa, line 42: Do we have any tests that assert the new behavior?

packages/reports/addon/components/grouped-list.ts
100starnight: if I make this into a getter it will respond to the search function - is that what we want to happen?

kevinhinterlong: I think that should be alright, we just want to prevent unecessary clicks. We could probably add a getter for `shouldOpenAllGroups` to check if there are only a few (maybe 5?) items.

Could we also pick up #1338

packages/reports/addon/components/column-selector.ts
```diff
@@ -33,4 +34,9 @@ export default class ColumnSelector extends Component<ColumnSelectorArgs> {
 
     return query ? searchRecords(allColumns, query, 'name') : allColumns;
   }
+
+  get isSingleCategory() {
+    const grouped = groupBy(this.args.allColumns, (row) => row.category?.split(',')[0] || `Uncategorized`);
```
kevinhinterlong, line 39:
```suggestion
    const grouped = groupBy(this.args.allColumns, (row) => row.category?.split(',')[0]);
```

packages/reports/addon/components/grouped-list.ts
```diff
@@ -49,12 +50,27 @@ export default class GroupedListComponent extends Component<Args> {
...
+        return -1;
+      } else {
+        return a[0].localeCompare(b[0]);
+      }
+    });
+
+    // flatten category headers and items into single list
+    return sortedCategories.reduce((items, ele) => {
+      const groupItems = ele[1];
+      const name = ele[0];
```
kevinhinterlong, line 68: Let's make this a forEach
```suggestion
    return sortedCategories.forEach(([name, groupItems]) => {
```

kevinhinterlong (approved):

jkusa (approved):

PR #1521, opened 2021-11-18
Fix: Tooltip pass the row correctly

kevinhinterlong:
## Description
The buildTooltip function keeps a reference to the chart builder to access row information, but we lookup the same instance instead of creating new ones so it gets overwritten if used more than once

## License

### DIFF
--- a/packages/core/tests/acceptance/line-chart-test.ts
+++ b/packages/core/tests/acceptance/line-chart-test.ts
@@ -14,20 +14,40 @@ module('Acceptance | line chart', function (hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
 
+  test('row count passed to tooltip formatter', async function (assert) {
+    assert.expect(2);
+
+    await visit('/line-chart');
+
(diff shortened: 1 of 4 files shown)


packages/core/tests/acceptance/line-chart-test.ts
```diff
@@ -14,20 +14,40 @@ module('Acceptance | line chart', function (hooks) {
...
+
+    await visit('/line-chart');
+
+    const metricCharts = findAll('.line-chart-widget.type-metric');
+    // check text of the tooltip container
+    await showTooltip(getEmberComponent(metricCharts[0].id), 1);
+    assert
+      .dom(metricCharts[0].querySelector('.chart-tooltip__value'))
+      .includesText('rc=6', 'The row count is correct for the first metric chart.');
+    // await hideTooltip(0);
```
jkusa, line 28:
```suggestion
```

jkusa (approved):
Just the one cleanup item

cythrawll (approved):

PR #1543, opened 2021-12-28
New: Adding overwriteFile rule.

cythrawll:
## Description

Adds overwrite file to delivery rule for cloud based delivery systems where previous data exports can be reasonably overwritten.

## Proposed Changes

- Adds overwriteFile to scheduling rules.

## License

### DIFF
--- a/packages/core/addon/models/fragments/scheduling-rules.ts
+++ b/packages/core/addon/models/fragments/scheduling-rules.ts
@@ -8,6 +8,9 @@ import Fragment from 'ember-data-model-fragments/fragment';
 export default class SchedulingRuleFragment extends Fragment {
   @attr('boolean')
   mustHaveData!: boolean;
+
+  @attr('boolean')
+  overwriteFile!: boolean;
 }
 
 declare module 'navi-core/models/registry' {
--- a/packages/core/tests/unit/models/delivery-rule-test.js
+++ b/packages/core/tests/unit/models/delivery-rule-test.js
@@ -14,6 +14,7 @@ const ExpectedDeliveryRule = {
   frequency: 'week',
   schedulingRules: {
     mustHaveData: false,
+    overwriteFile: false,
   },
   format: {
     type: 'csv',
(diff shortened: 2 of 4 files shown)


kevinhinterlong (approved):
Update gsheet poll in separate pr?
https://github.com/yavin-dev/framework/blob/87438183a2b5ac177d50d2c9ed5fd9c5a9e596cd/packages/core/addon/services/gsheet-export.ts#L77

cythrawll:
Yes we will update the poll mechanism in another PR. For export it's less necessary. But want to make sure we have good full integration test before we switch over to that mechanism.

kevinhinterlong (approved):

packages/core/addon/models/fragments/scheduling-rules.ts
```diff
@@ -8,6 +8,9 @@ import Fragment from 'ember-data-model-fragments/fragment';
 export default class SchedulingRuleFragment extends Fragment {
   @attr('boolean')
   mustHaveData!: boolean;
+
+  @attr('boolean')
+  overwriteFile!: boolean;
```
jkusa, line 13: IMHO, I think this could be better placed as an option of the format, as not all formats will support it (at least initially)
```ts
{
  "format": {
    "type": "gsheet",
    "options": {
      "overwriteFile": true
    }
  }
}
```

This could trail blaze the ability to add more options for particulars format in the future.

Another thought that comes to mind is that the original idea behind `scheduling-rules` is that it contains the rules for when to kick off a delivery. Not sure `overwriteFile` fits that idea.

packages/core/addon/models/fragments/format-options.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright 2022, Yahoo Holdings Inc.
+ * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
+ */
+import { attr } from '@ember-data/model';
+import Fragment from 'ember-data-model-fragments/fragment';
+
+export default class FormatOptions extends Fragment {
+  @attr('boolean')
+  overwriteFile!: boolean;
```
jkusa, line 10:
```suggestion
  declare overwriteFile: boolean;
```

jkusa (approved):

PR #1545, opened 2022-01-06
Elide params

jkusa:
Resolves #

## Proposed Changes

-

## License


holderb5 (approved):

packages/data/tests/unit/models/metadata/function-parameter-test.ts
```diff
@@ -68,70 +58,44 @@ module('Unit | Metadata Model | Function Parameter', function (hooks) {
...
-    };
+      'table value source type returns table values'
+    );
 
-    const TrendFunctionArgument = FunctionParameterFactory.create(trendArgPayload);
-    const trendValues = await TrendFunctionArgument.values;
+    //Test NONE
+    FunctionParameter.valueSourceType = ValueSourceType.NONE;
+    const noneValues = await FunctionParameter.values;
+    assert.deepEqual(noneValues, [], 'none value source type returns an empty array');
```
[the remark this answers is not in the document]
jkusa: We could, but then we would need to add more null checks.

kevinhinterlong (approved):

packages/data/addon/mirage/routes/graphql.js
```diff
@@ -488,7 +487,7 @@ export default function () {
                 throw new Error(`filtering on a related model is not currently supported in mirage mocks: ${filter}`);
               } else {
                 // `namespace.id` doesn't match any attribute of the records, so we want to change it to `namespaceIds`
```
cythrawll, line 489:
```suggestion
                // `namespace.id` doesn't match any attribute of the records, so we want to change it to `namespaceId`
```

daniellevinson (approved):

packages/reports/tests/integration/components/parameter-list-item-test.ts
```diff
@@ -2,40 +2,23 @@ import { module, test } from 'qunit';
...
-
-    this.set('allParams', [
-      { groupName: 'one', options: [{ name: 'foo' }, { name: 'bar' }] },
-      { groupName: 'two', options: [{ name: 'baz' }, { name: 'bam' }] },
-    ]);
-    assert.dom(this.element).hasText('foo', 'Shows name when there is no duplicate in grouped param list');
+    const arg1: ParameterListItemArgs['argument'] = { name: '', id: 'idField' };
+    this.set('arg', arg1);
+    await render(hbs`<ParameterListItem @argument={{this.arg}} />`);
+    assert.dom(this.element).hasText(`${arg1.id}`, 'it should show name field when name is not available');
```
kevinhinterlong, line 14:
```suggestion
    assert.dom(this.element).hasText(`${arg1.id}`, 'it should show id field when name is not available');
```

kevinhinterlong (approved):

PR #1554, opened 2022-01-21
Update: Add global date filters to dashboards

daniellevinson:
## Description

Add global date filters to dashboards

## Proposed Changes

- Include `timeDimension`s  in the dimension selector (**desirable TODO: Consolidate date dimensions of multiple datasources/tables into one option** while applying correctly to each request)
- Set default operator and filter values according to filter type
- Add filter type to dash filter model, serialize `grain` as `field` for `timeDimension`s
- Parse correctly elide datasources with namespace (`elide.namespace`)
- **Remove fact adapter error when grains don't match and let ws return errors**
- Display request errors in widgets

## Screenshots

Jan-26-2022 22-49-40

## License

### DIFF
--- a/packages/core/addon/serializers/dashboard.js
+++ b/packages/core/addon/serializers/dashboard.js
@@ -1,41 +1,46 @@
 /**
- * Copyright 2020, Yahoo Holdings Inc.
+ * Copyright 2022, Yahoo Holdings Inc.
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 
 import AssetSerializer from 'navi-core/serializers/asset';
 import { getDefaultDataSourceName } from 'navi-data/utils/adapter';
-import { getOwner } from '@ember/application';
 import { canonicalizeMetric } from 'navi-data/utils/metric';
 
-function v1ToV2Filter(filter, metadataService) {
+function v1ToV2Filter(filter) {
   let source;
-  let dimension; // the filter.dimension might be stored as dataSource.dimension
-  if (filter.dimension.includes('.')) {
-    [source, ...dimension] = filter.dimension.split('.');
-    dimension = dimension.join('.');
+  let dimension;
+  const dotOccurences = (filter.dimension.match(/\./g) ?? []).length;
+  if (dotOccurences > 0) {
+    /**
+     * filter.dimension might be stored as:
+     * dataSource.dimension (fili)
+     * dataSource.table.dimension (elide or dateTime in fili)
+     * dataSource.namespace.table.dimension (elide)
+     */
+    const split = filter.dimension.split('.');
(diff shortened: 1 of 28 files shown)


packages/core/addon/serializers/dashboard.js
```diff
@@ -84,9 +86,10 @@ export default AssetSerializer.extend({
     dashboard.data.attributes.filters = dashboard.data.attributes.filters.map((filter) => {
       const source = filterSources[buildKey(filter)];
       return {
+        type: filter.type,
```
kevinhinterlong, line 89: Can we run the elide NetflixTitles demo and try saving a dashboard with a time filter. I thought we were stuck with the existing fields because elide wouldn't save them
https://github.com/yavin-dev/framework/blob/aebf77433f79c9136579fa0f951f303b3756107c/packages/webservice/models/src/main/kotlin/com/yahoo/navi/ws/models/beans/fragments/request/DashboardFilter.kt#L7-L12

daniellevinson: image

packages/dashboards/addon/services/dashboard-data.ts
```diff
@@ -179,9 +179,19 @@ export default class DashboardDataService extends Service {
...
+    const validFilters = this.getValidGlobalFilters(dashboard, request).filter((filter) => filter.values.length > 0);
+
+    validFilters.forEach((filter) => {
+      if (filter.type === 'timeDimension') {
+        requestClone.filters
+          .toArray()
+          .filter((f) => f.field === filter.field)
+          .forEach((f) => {
+            requestClone.removeFilter(f);
+          });
```
daniellevinson, line 191:
```suggestion
          .forEach((f) => requestClone.removeFilter(f));
```

cythrawll (approved):

packages/dashboards/addon/services/dashboard-data.ts
```diff
@@ -179,9 +179,17 @@ export default class DashboardDataService extends Service {
...
-    this.getValidGlobalFilters(dashboard, request)
-      .filter((filter) => filter.values.length > 0)
-      .forEach((filter) => requestClone.addFilter(filter));
+    const validFilters = this.getValidGlobalFilters(dashboard, request).filter((filter) => filter.values.length > 0);
+
+    validFilters.forEach((filter) => {
+      if (filter.type === 'timeDimension') {
+        requestClone.filters
+          .toArray()
+          .filter((f) => f.field === filter.field)
```
kevinhinterlong, line 188: Can we add the `f.source` check here too

daniellevinson, line 188:
```suggestion
          .filter((f) => f.source === field.source && f.field === filter.field)
```

kevinhinterlong (approved):

PR #1560, opened 2022-01-27
Fix: Update start or end when they become invalid

kevinhinterlong:
## Description
When updating the start past the end or the end before the start, update both so the interval stays valid

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "navi",
       "version": "0.4.0",
       "hasInstallScript": true,
       "license": "MIT",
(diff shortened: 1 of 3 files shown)


kevinhinterlong:
> should the lock file be part of this PR?

🤷  looks like it keeps fighting with adding and removing the `"name": "navi"` line. I can remove it if you want

daniellevinson:
> > should the lock file be part of this PR?
> 
> 🤷 looks like it keeps fighting with adding and removing the `"name": "navi"` line. I can remove it if you want

I don't mind, let's stay out of this fight 😆

cythrawll (approved):

PR #1564, opened 2022-02-01
Fix: Sorts removed when params are changed.

cythrawll:
## Description

The original intent was sorts matching the base metric would be removed after a parameter is changed. However that behavior was lost as removal functions became more strict matching parameters. Causing invalid requests.

## Proposed Changes

- Change back to original intent where changing a parameter will remove all base metrics from sort.
- Future update will make this a bit better as we can map the sort to the intended column and update the sort instead of removing it.

## License

### DIFF
--- a/packages/reports/tests/unit/consumers/request/sort-test.ts
+++ b/packages/reports/tests/unit/consumers/request/sort-test.ts
@@ -22,6 +22,12 @@ const metric = {
   parameters: {},
   source: 'bardOne',
 };
+const metricWithParam = {
+  type: 'metric',
+  field: 'revenue',
+  parameters: { currency: 'USD' },
+  source: 'bardOne',
+};
 const dimension = {
   type: 'dimension',
   field: 'age',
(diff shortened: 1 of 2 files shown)


packages/reports/tests/unit/consumers/request/sort-test.ts
```diff
@@ -132,4 +138,26 @@ module('Unit | Consumer | request sort', function (hooks) {
...
+
+  test('UPDATE_COLUMN_FRAGMENT_WITH_PARAMS', function (assert) {
+    const request: RequestFragment = Store.createFragment('request', {
+      table: 'network',
+      limit: null,
+      dataSource: 'bardOne',
+      requestVersion: '2.0',
+      columns: [metricWithParam],
+      filters: [],
+      sorts: [{ ...metricWithParam, direction: 'desc' }],
```
jkusa, line 150: Can we add a few more sorts here? One with the same base but different param and another with a different base metric?

daniellevinson (approved):
can we open an issue around this to address it in the future?

cythrawll:
#1565

jkusa (approved):

PR #1566, opened 2022-02-02
Caching dimension filter values

100starnight:
## License

### DIFF
--- a/packages/data/addon/services/navi-dimension.ts
+++ b/packages/data/addon/services/navi-dimension.ts
@@ -100,7 +100,10 @@ export default class NaviDimensionService extends Service {
         cacheResponse = cache[dimId];
         if (cacheResponse) {
           // sort the results
-          results = SearchUtils.searchDimensionRecords(A(cacheResponse?.dimResponse.values), query, 5000);
+          let searchResults = SearchUtils.searchNaviDimensionRecords(A(cacheResponse.dimResponse.values), query);
+          results = {
+            values: searchResults.map((val) => val.record),
+          };
         }
       }
     } else {
@@ -123,6 +126,12 @@ export default class NaviDimensionService extends Service {
    */
   private _getCacheId(dimension: DimensionColumn, query?: string): string {
     let cacheId = dimension.columnMetadata.id;
+    const params = dimension.parameters;
+    if (params) {
+      for (const [key, value] of Object.entries(params)) {
+        cacheId = cacheId + '(' + key + ':' + value + ')';
+      }
+    }
     if (query) {
       cacheId = cacheId + '.' + query;
     }
--- a/packages/data/addon/services/navi-dimension.ts
+++ b/packages/data/addon/services/navi-dimension.ts
@@ -15,8 +15,13 @@ import type { DimensionColumn } from 'navi-data/models/metadata/dimension';
 import NaviDimensionResponse from 'navi-data/models/navi-dimension-response';
 import NaviDimensionModel from 'navi-data/models/navi-dimension';
 import CARDINALITY_SIZES from 'navi-data/utils/enums/cardinality-sizes';
+import Cache from 'navi-data/utils/classes/cache';
+import { canonicalizeMetric } from 'navi-data/utils/metric';
 import SearchUtils from 'navi-data/utils/search';
 import { A } from '@ember/array';
+
+const MAX_CACHE_SIZE = 10;
+const CACHE_TIMEOUT = 60 * 60 * 1000;
 
 export type ServiceOptions = {
   timeout?: number;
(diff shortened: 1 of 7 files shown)


packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+   * @param dimResponse NaviDimensionResponse
+   */
+  private _addToCache(cacheId: string, dimResponse: NaviDimensionResponse) {
+    if (!cacheId) {
+      return;
+    }
+    const cache = this._dimensionCache;
+    cache[cacheId] = {
+      time: Date.now(),
+      dimResponse: dimResponse,
```
daniellevinson, line 78:
```suggestion
      dimResponse,
```

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+  }
+
+  /**
+   * generates the _dimensionCache ID matching a particular dimensions call
+   * @param dimension DimensionColumn
+   * @returns string
+   */
+  private _getCacheId(dimension: DimensionColumn, query?: string): string {
+    let cacheId = dimension.columnMetadata.id;
+    const params = dimension.parameters;
```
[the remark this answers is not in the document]
kevinhinterlong:
```
dimension.columnMetadata.getCanonicalName(dimension.parameters)
```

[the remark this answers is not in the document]
daniellevinson: yeah that one's better 🙃

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+   *   2) removing the oldest item in the cache (if cache is over 5 entries)
+   */
+  private _trimCache(): void {
+    const cache = this._dimensionCache;
+    const now = Date.now();
+    let oldestDate = now;
+    let oldestId = '';
+    for (const [id, record] of Object.entries(cache)) {
+      // remove stale items (over an hour old)
+      if (now - record.time > 60 * 60 * 1000) {
```
jkusa, line 51: Can we make the stale threshold a cost above the service definition?

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+        delete cache[id];
+      }
+      // find oldest non-stale item
+      else if (record.time < oldestDate) {
+        oldestDate = record.time;
+        oldestId = id;
+      }
+    }
+    // if cache still too large, remove oldest item
+    if (Object.keys(cache).length > 5) {
```
jkusa, line 61: Can we make the max size a const?

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+    let results = undefined;
+
+    // for 'search' type requests, check for cached 'all' requests
+    if (!cacheResponse) {
+      const [dimId, query] = cacheId.split('.');
+      if (query) {
+        cacheResponse = cache[dimId];
+        if (cacheResponse) {
+          // sort the results
+          let searchResults = SearchUtils.searchNaviDimensionRecords(A(cacheResponse.dimResponse.values), query);
```
jkusa, line 103: Can we move this operation outside the cache?

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+          };
+        }
+      }
+    } else {
+      results = cacheResponse.dimResponse;
+    }
+
+    // if something was found, update record access time
+    if (cacheResponse) {
+      cacheResponse.time = Date.now();
```
jkusa, line 115: 👍 for LRU, but how does this fit in with the 1 hour expiration?

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +25,119 @@ export type ServiceOptions = {
...
+    }
+
+    this._trimCache();
```
jkusa, line 118: Do we want to do this at the beginning to check for expired cache entries?

packages/data/addon/services/navi-dimension.ts
```diff
@@ -45,6 +160,13 @@ export default class NaviDimensionService extends Service {
    * @param options - method options
    */
   @task *all(dimension: DimensionColumn, options: ServiceOptions = {}): TaskGenerator<NaviDimensionResponse> {
+    // check cache
+    const cacheId = this._getCacheId(dimension);
+    const cacheResponse = this._checkCache(cacheId);
```
jkusa, line 165: Can we call `getCacheId` inside `_checkCache` so `all` doesn't need to know the details of the cache?

packages/data/addon/utils/search.ts
```diff
@@ -115,4 +117,51 @@ export default {
...
+
+    // Filter, map, and sort records based on how close each record is to the search query
+    records.forEach((record) => {
+      // Determine relevance based on string match weight
+      const { value, suggestions } = record;
+      const searchString = suggestions ? (Object.values(suggestions).join(' ') || '').toLowerCase() : '';
+      let nonIdMatchWeight = this.getPartialMatchWeight(searchString, query.toLowerCase()),
+        //@ts-ignore
+        idMatchWeight = this.getExactMatchWeight((value || '').toLowerCase(), query.toLowerCase()),
+        relevance = nonIdMatchWeight || idMatchWeight;
```
jkusa, line 147:
```suggestion
      const valueMatchWeight = this.getExactMatchWeight((value || '').toLowerCase(), query.toLowerCase());
      const suggestionsMatchWeight = this.getPartialMatchWeight(searchString, query.toLowerCase());
      const relevance = valueMatchWeight || suggestionsMatchWeight;
```

packages/data/addon/services/navi-dimension.ts
100starnight: If you access a record, this resets the timer on that cache entry (records are removed 1 hour from when they were last accessed, not necessarily 1 hour from their creation)

packages/data/addon/services/navi-dimension.ts
100starnight: Depends; moving it up wouldn't change much in practice, but it would mean that the max cache size would be one larger than whatever it's manually set to. We'd also run the risk of removing records and then looking for them in the same service call, but the odds of that edge-case actually happening are kinda low

packages/data/addon/services/navi-dimension.ts
100starnight: Yes, but then getCacheId will be called twice for any instance where something isn't in the cache (once to check the cache at the beginning, and then again when adding the new records to the cache). If that's nbd, though, I can change it

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+  records: T; // the actual records
+};
+
+/**
+ * a cache
+ * to use, give it a unique string (cacheId) and whatever you want stored alongside that key
+ * returns only exact matches
+ */
+export default class Cache<T> {
+  private _maxCacheSize;
```
cythrawll, line 18: Preference here, but I think caches should have a `Map` like interface. (with a few extra methods for cache stuff). So my suggestions will be based in that preference.

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+      if (now - cacheResult.creationTime >= this._cacheTimeout) {
+        delete cache[id];
+      }
+    }
+  }
+
+  /**
+   * removes the least recently accessed items until cache is at or below max size
+   */
+  private _trimExtraItems(): void {
```
cythrawll, line 51: consider returning number of deletes (could be useful for creating cache stats later)

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  /**
+   * adds a given key/value pair to the cache
+   * @param {string} cacheId - the unique identifier associated with the thing you're putting into the cache (result of getCacheId function)
+   * @param {any} records - the thing you want to store in the cache
+   * @returns void
+   */
+  addToCache(cacheId: string, records: T): void {
```
cythrawll, line 72: `set`

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+  constructor(cacheIdFunction: (args: any) => string, maxCacheSize = 10, cacheTimeout = 60 * 60 * 1000) {
+    this.getCacheId = cacheIdFunction;
+    this._maxCacheSize = maxCacheSize;
+    this._cacheTimeout = cacheTimeout;
+  }
+
+  /**
+   * removes items that were created too long ago from the cache
+   */
+  private _removeStaleItems(): void {
```
cythrawll, line 38: consider returning number of deletes (could be useful for creating cache stats later)

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    this._trimExtraItems();
+  }
+
+  /**
+   * looks for and returns a specific record in the cache
+   * @param cacheId - the unique identifier associated with the thing you're looking for (result of getCacheId function)
+   * @returns the thing you're looking for (or undefined, if it's not in the cache)
+   */
+  checkCache(cacheId: string): T | undefined {
```
cythrawll, line 93: `get`, what if it took an optional function that returns `T`.  If cache misses it can run that function and store the results in given cache key. (then returns the result of the function as well). Could cleanup some of the using the cache.

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+    let results = undefined;
+
+    // if something was found, update record access time
+    if (cacheResponse) {
+      cacheResponse.lastAccessed = Date.now();
+      results = cacheResponse.records;
+    }
+
+    return results;
+  }
```
cythrawll, line 110: Consider implementing `has(cacheId: string): boolean`.  simply checks for presence of cache entry.

cythrawll (approved):

packages/data/addon/utils/classes/cache.ts
```diff
@@ -0,0 +1,124 @@
...
+          oldestDate = value.lastAccessed;
+          oldestKey = key;
+        }
+      });
+      cache.delete(oldestKey);
+
+      removedCount = removedCount + 1;
+    }
+    return removedCount;
+  }
```
kevinhinterlong, line 70: Since the interface only allows you to set one item at a time, I think we should be able to make something simpler like

```js
_makeSpaceFor(cacheKey) {
    const cache = this._cache;
    // if the key is not in the cache and there is no space left
    if(!cache.has(cacheKey) && cache.size === this._maxCacheSize) {
        // find and remove least recently accessed
   }
}

setItem(cacheKey, records) {
    this._removeStaleItems();
    if (!cacheKey) {
      return;
    }
	this._makeSpaceFor(cacheKey);
    //....
}
```

packages/data/addon/services/navi-dimension.ts
```diff
@@ -22,7 +30,74 @@ export type ServiceOptions = {
...
+}
+
 export default class NaviDimensionService extends Service {
+  /**
+   * @property {DimensionCache} _dimensionCache - local cache for dimensions
+   */
+  private _dimensionCache = new DimensionCache();
+
+  clearCache(): void {
+    this._dimensionCache = new DimensionCache();
```
daniellevinson, line 98: can the `Cache` class itself have a `clear` method or is it more than just creating a new map?

packages/data/addon/utils/search.ts
```diff
@@ -115,4 +117,47 @@ export default {
...
+   *
+   * @method searchDimensionRecords
+   * @param {Array} records - collection of records to search
+   * @param {String} query - search query used to filter and rank assets
+   * @param {Number} resultLimit - maximum number of results
+   * @param {Number} [page] - page number starting from page 1
+   * @returns {values: NaviDimensionModel}
+   */
+  searchNaviDimensionRecords(records: NaviDimensionModel[], query: string): { values: NaviDimensionModel[] } {
+    let results: MutableArray<{ record: NaviDimensionModel; relevance: number }> = A(); // = [] as;
```
kevinhinterlong, line 132:
```suggestion
    let results: MutableArray<{ record: NaviDimensionModel; relevance: number }> = A();
```

packages/data/addon/services/navi-dimension.ts
```diff
@@ -14,6 +14,14 @@ import type { DimensionFilter } from 'navi-data/adapters/dimensions/interface';
...
 import NaviDimensionResponse from 'navi-data/models/navi-dimension-response';
 import NaviDimensionModel from 'navi-data/models/navi-dimension';
+import CARDINALITY_SIZES from 'navi-data/utils/enums/cardinality-sizes';
+import Cache from 'navi-data/utils/classes/cache';
+import { canonicalizeMetric } from 'navi-data/utils/metric';
+import SearchUtils from 'navi-data/utils/search';
+import { A } from '@ember/array';
+
+const MAX_CACHE_SIZE = 10;
+const CACHE_TIMEOUT = 60 * 60 * 1000;
```
kevinhinterlong, line 24: Can we move these to the navi config

packages/data/addon/services/navi-dimension.ts
```diff
@@ -47,7 +44,7 @@ class DimensionCache extends Cache<NaviDimensionResponse> {
       });
       return cannonicalName;
     }
-    super(getDimensionCacheId, MAX_CACHE_SIZE, CACHE_TIMEOUT);
```
kevinhinterlong: We should add these to the config
https://github.com/yahoo/yavin/blob/018103098cbb47f7b932a38edda5630271e89714/packages/data/config/environment.js#L7
and declare them here
https://github.com/yahoo/yavin/blob/04e324c3a0d93737c7fb5f533209aac1bb65cbc9/packages/data/config/navi-config.d.ts#L42

```
dimensionCache: {
	size: 10,
	timeoutMs: 60 * 60 * 1000
}
```

kevinhinterlong (approved):

PR #1567, opened 2022-02-02
Build: Add pr builds for vis-v2 branch

jkusa:
Resolves #

## Proposed Changes

-

## License

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -29,7 +29,6 @@ shared:
           - ~pr
           - ~commit
           - ~pr:vis-v2
-          - ~commit:vis-v2
   environment:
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -28,6 +28,8 @@ shared:
         requires:
           - ~pr
           - ~commit
+          - ~pr:vis-v2
+          - ~commit:vis-v2
```
kevinhinterlong, line 32: Just for pr builds?
```suggestion
```
Could add webservice pr build too

jkusa: :+1: If we need/want this we can add it later

PR #1582, opened 2022-02-16
Update: Hide datasources from UI

100starnight:
## License

### DIFF
--- a/packages/reports/addon/components/report-builder/sidebar.ts
+++ b/packages/reports/addon/components/report-builder/sidebar.ts
@@ -127,7 +127,7 @@ export default class ReportBuilderSidebar extends Component<Args> {
       sources.push({
         name: dataSource.displayName,
         description: dataSource.description,
-        isHidden: dataSource.isHidden,
+        hide: dataSource.hide,
         source: dataSource,
       });
 
@@ -144,11 +144,11 @@ export default class ReportBuilderSidebar extends Component<Args> {
       return yield [];
     }
 
-    return yield dataSource.namespaces.map(({ name, displayName, description, isHidden }) => {
+    return yield dataSource.namespaces.map(({ name, displayName, description, hide }) => {
       return {
         name: displayName,
         description,
-        isHidden,
+        hide,
         source: getDataSource(`${dataSource.name}.${name}`),
       };
     });
--- a/packages/reports/addon/components/report-builder/sidebar.ts
+++ b/packages/reports/addon/components/report-builder/sidebar.ts
@@ -135,7 +135,7 @@ export default class ReportBuilderSidebar extends Component<Args> {
     });
 
     const namespaces = (yield all(fetchNamespacesTasks)).flat();
-    sources = [...sources, ...namespaces];
+    sources = [...sources, ...namespaces].filter((source) => !source.hide);
     return yield sortBy(sources, ['name']);
   }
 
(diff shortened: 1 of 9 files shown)


packages/reports/tests/dummy/config/environment.js
```diff
@@ -64,11 +64,24 @@ module.exports = function (environment) {
           type: 'elide',
         },
         {
-          name: 'elideTwo',
-          displayName: 'Elide Two',
-          description: 'Elide Two Description',
+          name: 'elideTwoDefault',
+          displayName: 'Elide Two Default',
+          isHidden: true,
```
daniellevinson, line 69: might be a little confusing that `isHidden` is true at the root level yet some of the namespaces are still shown.
but it's either a flag (`hideDefaultNamespace` is probably too elide-specific) or flatten everything and add namespace id for each entry
anyway I guess it's good for what we need now 👍 and we can refactor in the future

packages/reports/tests/dummy/config/environment.js
```diff
@@ -64,11 +64,24 @@ module.exports = function (environment) {
           type: 'elide',
         },
         {
-          name: 'elideTwo',
-          displayName: 'Elide Two',
-          description: 'Elide Two Description',
+          name: 'elideTwoDefault',
+          displayName: 'Elide Two Default',
+          hide: true,
```
daniellevinson, line 69: can we add a test that the source list is displayed correctly with this config?

daniellevinson: oh ok it's fine

packages/reports/tests/dummy/config/environment.js
```diff
@@ -64,11 +64,24 @@ module.exports = function (environment) {
...
+            {
+              name: 'elideTwo',
+              displayName: 'Elide Two',
+              description: 'Elide Two Description',
+            },
+            {
+              name: 'elideTwoCopy',
+              displayName: 'Elide Two Copy',
+              description: 'Copy of Elide Two Description',
+              hide: true,
```
jkusa, line 82: What is the case for telling the UI about the namespace but then hiding it? Could removing then namespace entry be just as effective?

100starnight: Theoretically yes, you could just remove the namespace, but IFF the data source you're trying to hide is actually a namespace. This feature is more specifically targeted at the case of trying to hide a default data source while keeping the namespaces under it (like we're trying to do with Digits+).
As a bonus, though, this should also make it easy to suspend data sources in the future - if you wanted to stop people from making new reports/dashboards, but didn't necessarily want to migrate the items already relying on it, you could just hide the source and basically pause production on it without having to worry about backwards compatibility for already existing things

packages/reports/addon/components/report-builder/sidebar.ts
```diff
@@ -143,10 +144,11 @@ export default class ReportBuilderSidebar extends Component<Args> {
       return yield [];
     }
 
-    return yield dataSource.namespaces.map(({ name, displayName, description }) => {
+    return yield dataSource.namespaces.map(({ name, displayName, description, hide }) => {
```
jkusa, line 147: Instead of filtering the data source in the template, could we filter out the data sources at a lower level?

jkusa (approved):

PR #1585, opened 2022-03-01
Fix: Update sorts on request parameter changes

100starnight:
## Description
When column parameters are updated, sorts update with them instead of being removed.

## License

### DIFF
--- a/packages/core/addon/serializers/request.js
+++ b/packages/core/addon/serializers/request.js
@@ -53,19 +53,13 @@ export default class RequestSerializer extends JSONSerializer {
     // add cid to sorts
     const map = new Map();
     normalized.columns?.forEach(({ type, field, parameters, cid }) => {
-      const canonicalName =
-        type === 'timeDimension'
-          ? canonicalizeMetric({ metric: 'dateTime', parameters })
-          : canonicalizeMetric({ metric: field, parameters });
+      const canonicalName = canonicalizeMetric({ metric: field, parameters });
       assert('columns must have cid', cid);
       map.set(`${type}|${canonicalName}`, cid);
     });
     normalized.sorts?.forEach((sort) => {
       const { type, field, parameters } = sort;
-      const canonicalName =
-        type === 'timeDimension'
-          ? canonicalizeMetric({ metric: 'dateTime', parameters })
-          : canonicalizeMetric({ metric: field, parameters });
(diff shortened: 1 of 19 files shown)


cythrawll:
I don't believe this solves the problem and introduces some bugs.  I'll  make a quick screen cap of what I was trying to prevent with refactoring toward CIDs.

holderb5 (dismissed):

cythrawll (dismissed):

packages/core/tests/unit/models/report-test.js
```diff
@@ -145,6 +145,15 @@ module('Unit | Model | report', function (hooks) {
...
       });
 
+      // check and remove cid (b/c they're non-deterministic)
+      const sortCids = [];
+      serialized.data.attributes.request.sorts.forEach((sort, idx) => {
+        sortCids.push(sort.cid);
+        unset(serialized, `data.attributes.request.sorts[${idx}].cid`);
+      });
+      assert.equal(sortCids.length, 1, 'there is one sort');
+      assert.deepEqual(sortCids[0], cids[3], 'sort has same cid as matching column');
```
kevinhinterlong, line 155: We need to add a serializer for sort because this should not show up in the serialized request (or update the persistence ws to support it but I think it's ok to keep client side only for now)

```js
import BaseJsonSerializer from 'navi-core/serializers/base-json-serializer';

export default class SortSerializer extends BaseJsonSerializer {
  attrs: {
    cid: { serialize: false },
  }
}
```

packages/core/addon/serializers/request.js
```diff
@@ -48,6 +50,27 @@ export default class RequestSerializer extends JSONSerializer {
       normalized[attr].forEach((fragment) => (fragment.source = normalized.dataSource))
     );
 
+    // add cid to sorts
+    const map = new Map();
+    normalized.columns?.forEach(({ type, field, parameters, cid }) => {
+      const canonicalName =
+        type === 'timeDimension'
+          ? canonicalizeMetric({ metric: 'dateTime', parameters })
+          : canonicalizeMetric({ metric: field, parameters });
```
kevinhinterlong, line 59: We should be able to update it to
```suggestion
      const canonicalName = canonicalizeMetric({ metric: field, parameters })
```

PR #1627, opened 2022-05-13
Update: Scroll to and highlight last added filter

daniellevinson:
Resolves #

## Proposed Changes

-

## License

### DIFF
--- a/packages/reports/addon/components/filter-collection.ts
+++ b/packages/reports/addon/components/filter-collection.ts
@@ -1,18 +1,22 @@
 /**
- * Copyright 2020, Yahoo Holdings Inc.
+ * Copyright 2022, Yahoo Holdings Inc.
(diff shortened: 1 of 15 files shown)


packages/reports/addon/components/filter-collection.ts
```diff
@@ -26,6 +30,8 @@ type FilterItem = {
 export default class FilterCollection extends Component<FilterCollectionArgs> {
   @service requestConstrainer!: RequestConstrainer;
 
+  componentElement!: HTMLElement;
```
kevinhinterlong, line 33: Can we rename this to lastAddedFilterElement or something similar? I feel like we typically use componentElement to mean the top level element in the template
```suggestion
  componentElement: HTMLElement | undefined;
```

kevinhinterlong (approved):

PR #1634, opened 2022-05-23
Update: Manually configure dependency lookup

kevinhinterlong:
## Description
Allows a generic `injector` to be passed in to models so that they are no longer tied to using the ember service lookup methods

## License


packages/c3/tests/dummy/app/routes/goal-gauge.ts
```diff
@@ -23,7 +27,7 @@ export default class GoalGaugeRoute extends Route {
       dataSource: 'bardOne',
       requestVersion: '2.0',
     });
-    const response = NaviFactResponse.create({ rows: [{ pageViews: 3060000000 }] });
+    const response = new NaviFactResponse(this.clientInjector, { rows: [{ pageViews: 3060000000 }] });
```
jkusa, line 30: Did you consider adding the injector to a NaviFactResponse class/subclass, instead of passing it as a constructor arg?

packages/c3/tests/unit/chart-builders/dimension-test.ts
```diff
@@ -33,7 +34,8 @@ const DATA = NaviFactResponse.create({
   ]
 });
 // prettier-ignore
-const DATA2 = NaviFactResponse.create({
+//@ts-ignore - skip injector
```
jkusa, line 37: why skip it?

packages/core/tests/dummy/app/routes/pie-chart.ts
```diff
@@ -1,110 +0,0 @@
-import { A } from '@ember/array';
```
jkusa, line 1: 👍

jkusa (approved):
A few minor comments otherwise 👍

packages/c3/tests/unit/chart-builders/dimension-test.ts
kevinhinterlong: This was skipped since it was in a scope that it couldn't do a service lookup. I've moved it inside the test so it now passes it in

PR #1635, opened 2022-05-23
Fix: Update cache

100starnight:
## Description
When checking for item in cache, remove stale items from cache prior to check
Remove parameter getCacheKey from class constructor

## License

### DIFF
--- a/packages/data/addon/utils/classes/cache.ts
+++ b/packages/data/addon/utils/classes/cache.ts
@@ -31,6 +31,17 @@ export default class Cache<T> {
   }
 
   /**
+   * checks if an item was created too long ago, if so removes it from the cache
+   * @param cacheKey - the unique identifier you're checking the age of
+   */
+  private _removeStaleItem(cacheKey: string): void {
+    const item = this._cache.get(cacheKey);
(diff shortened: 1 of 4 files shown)


packages/data/addon/utils/classes/cache.ts
```diff
@@ -133,4 +131,12 @@ export default class Cache<T> {
 
     return results;
   }
+
+  /**
+   * removes a given key/record pair from the cache
+   * @param {string} cacheKey - the unique identifier you want to remove from the cache
+   */
+  removeItem(cacheKey: string): void {
```
kevinhinterlong, line 139: Can we add a test for this

kevinhinterlong (approved):

packages/data/addon/utils/classes/cache.ts
```diff
@@ -83,12 +80,13 @@ export default class Cache<T> {
    * @returns {boolean} boolean indicating presence of given key in cache
    */
   has(cacheKey: string): boolean {
+    this._removeStaleItems();
```
jkusa, line 83: Do we need to clean the entire cache for a lookup? Is it possible to just remove the single entry if it is stale?

jkusa (approved):

PR #1637, opened 2022-05-24
Update: Move some utils to client, use latest moment

kevinhinterlong:
## Description
Move some utils to client and upgrade to the latest moment

## License


packages/client/src/utils/classes/interval.ts
```diff
@@ -154,7 +155,7 @@ export default class Interval {
     const period = getPeriodForGrain(grain);
 
     let { start, end } = this._asMoments(period);
-    assert('when making the end of an interval exclusive, the end should exist', end);
+    invariant(end, 'when making the end of an interval exclusive, the end should exist');
```
daniellevinson, line 158: do we still need this? might have forgotten to remove it in my PR

packages/client/src/utils/date.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Copyright 2021, Yahoo Holdings Inc.
```
daniellevinson, line 2:
```suggestion
 * Copyright 2022, Yahoo Holdings Inc.
```

packages/client/src/utils/duration-utils.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright 2020, Yahoo Holdings Inc.
```
daniellevinson, line 2:
```suggestion
 * Copyright 2022, Yahoo Holdings Inc.
```

daniellevinson (approved):
one tiny thing, not a blocker

PR #1649, opened 2022-06-01
Update: Move graphql queries to client package

kevinhinterlong:
## Description
Our typescript eslint has been broken on most packages because our parser was too old. It's now been updated and some lint errors have been fixed

## License


packages/dashboards/addon/components/dashboard-filters.ts
```diff
@@ -28,7 +28,7 @@ export default class DashboardFiltersComponent extends Component<Args> {
   @tracked
   isAddingMode = false;
 
-  @computed('args.dashboard.filters.[]')
+  @computed('args.dashboard.filters.[]', 'store')
```
daniellevinson, line 31: maybe it's just me but it's unfortunate we have to add it here

daniellevinson (approved):

packages/dashboards/addon/components/dashboard-filters.ts
kevinhinterlong: Yea some of the lints around computed don't seem to work very well so I figured I'll just fix it until we can move away from it

jkusa (approved):

PR #1660, opened 2022-06-08
Fix: Make dashboard view match report view when no data present

100starnight:
## Description
The report view doesn't render visualizations when there is no data available; this PR fixes the dashboard view match the report view.

## Screenshots
Report view:
Screen Shot 2022-06-08 at 3 33 50 PM
Dashboard view:
Screen Shot 2022-06-08 at 3 44 27 PM

## License

### DIFF
--- a/packages/dashboards/addon/components/navi-widget.ts
+++ b/packages/dashboards/addon/components/navi-widget.ts
@@ -97,7 +97,7 @@ export default class NaviWidget extends Component<NaviWidgetArgs> {
    * whether or not there is data to display
    */
   get hasNoData() {
-    return this.data?.firstObject?.response?.meta.pagination?.numberOfResults === 0;
+    return this.data?.firstObject?.response?.rows?.length === 0;
   }
 
   @action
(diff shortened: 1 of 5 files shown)


cythrawll (commented):
Can we add a test when that metaData doesn't exist. Seems to be missing a lot of the time in responses.

packages/dashboards/addon/components/navi-widget.ts
```diff
@@ -93,6 +93,13 @@ export default class NaviWidget extends Component<NaviWidgetArgs> {
     return isSuccessful ? arr(value as WidgetData) : null;
   }
 
+  /**
+   * whether or not there is data to display
+   */
+  get hasNoData() {
+    return this.data?.firstObject?.response?.meta.pagination?.numberOfResults === 0;
```
jkusa, line 100: I'm not sure `numberOfResults` will always be present. Are we populating this in Elide responses? We might be better off testing `response.rows.length`

100starnight: Sure we can use row length - do you want me to update report-view so that it also uses row length? https://github.com/yavin-dev/framework/blob/master/packages/reports/addon/components/report-view.ts#L92

cythrawll: *nod*

kevinhinterlong (approved):

jkusa (approved):

PR #1663, opened 2022-06-13
Update: Move bard fact adapter to client

kevinhinterlong:
## Description
Moves the bard fact adapter to the client

- `GrainWithAll` is copied from `navi-data/serializers/metadata/bard` and should be removed later
- `@waitFor` is needed for tests to keep the current behavior (added to every service but only facts needs it at this point), but it achieves this by wrapping promises so the metadata tests directly comparing promises needed to be updated
- replace `ember-ajax` with `cross-fetch` in fact adapter
- need to set up msw and move the bard test over to client

## License


packages/client/src/plugins/bard/adapter/facts.ts
```diff
@@ -0,0 +1,544 @@
...
+import type RequestDecoratorService from '../../../services/interfaces/request-decorator.js';
+
+/**
+ * @deprecated TODO import this from the bard metadata serializer
+ */
+type GrainWithAll = Grain | 'all';
+
+export type Query = Record<string, string | number | boolean>;
+
+export class FetchError extends Error {
```
jkusa, line 36: Is this unique to fili?

packages/client/src/plugins/bard/adapter/facts.ts
```diff
@@ -0,0 +1,544 @@
...
+ * @param column the column to check if it is a dateTime
+ */
+function isDateTime(column: Column | Filter | Sort) {
+  return column.type === 'timeDimension' && column.field.endsWith('.dateTime');
+}
+
+export type FilterOperator = 'in' | 'notin' | 'contains';
+export type HavingOperator = 'gt' | 'gte' | 'lt' | 'lte' | 'eq' | 'neq' | 'bet' | 'nbet';
+
+export default class BardFactsAdapter extends NativeWithCreate implements NaviFactAdapter {
```
jkusa, line 213: Is this a good time to rename to Fili?

packages/core/addon/helpers/is-forbidden.ts
```diff
@@ -6,13 +6,13 @@
  */
 import { helper as buildHelper } from '@ember/component/helper';
 import NaviAdapterError from '@yavin/client/errors/navi-adapter-error';
-import { AjaxError, isForbiddenError } from 'ember-ajax/errors';
+import { FetchError } from '@yavin/client/plugins/bard/adapter/facts';
```
jkusa, line 9: Should `FetchError` be data source agnostic?

jkusa (approved):

PR #1665, opened 2022-06-14
Update: Move fili fact serializer to client

kevinhinterlong:
## Description
Moves the fili fact serializer to the client package and setups up msw for node/web so that we can move over tests which mock http requests

## License


cythrawll (approved):

packages/client/src/plugins/fili/serializers/facts.ts
```diff
@@ -0,0 +1,117 @@
...
+type BardError = {
+  description: string;
+  druidQuery: Record<string, unknown>;
+  reason: string;
+  requestId: string;
+  status: number;
+  statusName: string;
+};
+
+export default class FiliFactsSerializer extends NativeWithCreate implements NaviFactSerializer {
```
jkusa, line 27: Is this a good time to start calling these normalizers?

packages/client/src/plugins/fili/serializers/facts.ts
```diff
@@ -0,0 +1,117 @@
+/**
+ * Copyright 2021, Yahoo Holdings Inc.
```
jkusa, line 2:
```suggestion
 * Copyright 2022, Yahoo Holdings Inc.
```

jkusa (approved):

packages/client/src/plugins/fili/serializers/facts.ts
kevinhinterlong: Pushing this off for now

PR #1668, opened 2022-06-17
Update: Move fili fact adapter test to client

kevinhinterlong:
## Description
Move the fili fact adapter tests to client and fix msw setup to only create a single instance for the whole test suite

## License


packages/client/test/tests/unit/plugins/fili/adapters/facts-test.ts
```diff
@@ -1,67 +1,1491 @@
...
+  dataSource: 'bardOne',
+  requestVersion: '2.0',
+  limit: null,
+  columns: [],
+  filters: [],
+  sorts: [],
+};
+
+const TestRequest: RequestV2 = {
+    table: 'table1',
```
jkusa, line 28: Is the indent off here?

jkusa (approved):

PR #1676, opened 2022-06-22
Update: Move fact service to client and make ember wrapper

kevinhinterlong:
## Description
Moves the fact service to the client package. We currently do not do any cancellation for fact queries from the UI, so we should refactor that later, but I've verified by adding a delay and calling `cancel()` on the task wrapper instance that it cascades down properly

## License

### DIFF
--- a/packages/client/src/adapters/facts/interface.ts
+++ b/packages/client/src/adapters/facts/interface.ts
@@ -3,6 +3,7 @@
  * Licensed under the terms of the MIT license. See accompanying LICENSE.md file for terms.
  */
 import type { Request } from '../../request.js';
+import type { Task } from 'effection';
 
 export type RequestOptions = {
   clientId?: string;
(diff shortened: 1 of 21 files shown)


packages/client/src/adapters/facts/interface.ts
```diff
@@ -83,7 +84,7 @@ export interface TableExportResult {
   message: string;
 }
 export default interface NaviFactAdapter {
-  fetchDataForRequest(request: Request, options: RequestOptions): Promise<unknown>;
+  fetchDataForRequest(request: Request, options: RequestOptions): Promise<unknown> | Task<unknown>;
   urlForFindQuery(request: Request, options: RequestOptions): string;
-  urlForDownloadQuery(request: Request, options: RequestOptions): Promise<string>;
+  urlForDownloadQuery(request: Request, options: RequestOptions): Promise<string> | Task<unknown>;
```
jkusa, line 89: What is the Task generic here? Should it be `string` like the promise?

kevinhinterlong: Yep, it should be string
```suggestion
  urlForDownloadQuery(request: Request, options: RequestOptions): Promise<string> | Task<string>;
```

jkusa (approved):

daniellevinson (approved):

PR #1680, opened 2022-06-22
Update: Move metadata service to client

kevinhinterlong:
## Description
Moves the metadata service to client and adds an ember concurrency wrapper for test waiting/task cancelling (although the metadata service is promise based so it does not yet support it)

## License

### DIFF
--- a/packages/data/addon/services/navi-facts.ts
+++ b/packages/data/addon/services/navi-facts.ts
@@ -35,9 +35,10 @@ export default class NaviFactsService extends FactService {
   @task *taskWrapper<T>(taskLike: () => Promise<T>): TaskGenerator<T> {
     const result = taskLike();
     try {
-      return yield waitForPromise(result);
+      // Wrap promise value since waitForPromise assumes they are unique
+      return yield waitForPromise(result.then((v) => Promise.resolve(v)));
     } finally {
-      maybeHalt(result);
+      yield waitForPromise(maybeHalt(result));
     }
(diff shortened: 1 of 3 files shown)


packages/data/addon/services/navi-facts.ts
```diff
@@ -35,9 +35,10 @@ export default class NaviFactsService extends FactService {
   @task *taskWrapper<T>(taskLike: () => Promise<T>): TaskGenerator<T> {
     const result = taskLike();
     try {
-      return yield waitForPromise(result);
+      // Wrap promise value since waitForPromise assumes they are unique
+      return yield waitForPromise(result.then((v) => Promise.resolve(v)));
```
jkusa, line 39: why is promise chaining required? `result.then((v) => Promise.resolve(v))`

kevinhinterlong: For facts it is not required, only the metadata service `loadMetadata` call uses it since we store the promise and prevent people from loading a datasource twice. But I figured it's cheap and could be more confusing to fix later if we use caching by promises

Without it, for metadata we run into this error
https://github.com/emberjs/ember-test-waiters/blob/ef29761477e2a397bc1ebf08725399e1a23f7f8b/addon/%40ember/test-waiters/build-waiter.ts#L38

daniellevinson (approved):

holderb5 (approved):

PR #1705, opened 2022-08-31
Fix: visualization-toggle crash

cythrawll:
## Description

crashes if selected category is no longer has any valid visualizations

## Proposed Changes

- Pass empty list to `DenaliToggle` instead of `undefined` in this case.

## License

### DIFF
--- a/packages/reports/addon/components/visualization-toggle.ts
+++ b/packages/reports/addon/components/visualization-toggle.ts
@@ -56,7 +56,7 @@ export default class VisualizationToggle extends Component<Args> {
   }
 
   get categoryVisualizations() {
-    let visualizations = this.valid.categoryToVisualizations[this.selectedCategory];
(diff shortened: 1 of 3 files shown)


packages/reports/addon/components/visualization-toggle.ts
```diff
@@ -55,6 +55,11 @@ export default class VisualizationToggle extends Component<Args> {
     return category ?? this.valid.categories[0];
   }
 
+  get categoryVisualizations() {
+    let visualizations = this.valid.categoryToVisualizations[this.selectedCategory];
```
kevinhinterlong, line 59:
```suggestion
    const visualizations = this.valid.categoryToVisualizations[this.selectedCategory];
```

kevinhinterlong (approved):

jkusa (approved):

PR #1709, opened 2022-09-21
Row limit

cythrawll:
## Description

Allow users to change the row limit passed the UI. Shows warning on large page limits.

## Proposed Changes

- Add UI for changing row limit above filter box. 
- Adjust pagination logic on adapters: `request.limit` will override `perPage` options if `request.limit is <= perPage`

## Screenshots
Screen Shot 2022-09-21 at 10 28 18 AM

## License

### DIFF
--- a/packages/client/test/tests/unit/plugins/fili/adapters/facts-test.ts
+++ b/packages/client/test/tests/unit/plugins/fili/adapters/facts-test.ts
@@ -1229,7 +1229,7 @@ module('Unit | Plugins | Fili | Adapters | facts', function (hooks) {
         page: 1,
         perPage: 22,
       },
-      'the request limit overrides the perPage option'
+      'request limit override adapter options if limit is less than page size'
     );
   });
 
--- a/packages/client/test/tests/unit/plugins/fili/adapters/facts-test.ts
+++ b/packages/client/test/tests/unit/plugins/fili/adapters/facts-test.ts
@@ -1227,9 +1227,9 @@ module('Unit | Plugins | Fili | Adapters | facts', function (hooks) {
         having: 'm1-gt[0]',
         format: 'json',
         page: 1,
-        perPage: 100,
+        perPage: 22,
       },
-      'it overrides the request limit with the pagination request options'
+      'the request limit overrides the perPage option'
     );
   });
 
--- a/packages/data/tests/unit/adapters/facts/elide-test.ts
+++ b/packages/data/tests/unit/adapters/facts/elide-test.ts
@@ -296,8 +296,8 @@ module('Unit | Adapter | facts/elide', function (hooks) {
 
     assert.deepEqual(
       adapter['getPaginationOptions'](TestRequest, { perPage: 3, page: 2 }),
-      { first: 3, after: 3 },
-      'specifying `perPage` and `page` overrides request limit'
+      { first: 10000, after: 0 },
+      'specifying `perPage` and `page` does not override request limit'
     );
 
     const limitless = { ...TestRequest, limit: null };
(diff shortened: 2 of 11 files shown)


packages/reports/addon/components/report-builder.hbs
```diff
@@ -43,9 +43,20 @@
             />
           </button>
           {{#unless @isFiltersCollapsed}}
-            <span class="report-builder__container-header__message flex align-items-center">
-              Click <DenaliIcon @icon="filter-add" @size="small" class="report-builder__container-header__message--icon m-x-2" /> next to a dimension/metric to add a filter.
-            </span>
+            <DenaliInputGroup @label="Row Limit" class="report-builder__row-limit">
+              {{#if (gt this.request.limit 10000)}}
```
kevinhinterlong, line 47: We might need to change this to a warning about incomplete results because the UI doesn't support paging through fact requests

kevinhinterlong (commented):
It looks like the default 10k limit is manually being passed in, so we'll have to add a check if the request already has a limit defined
https://github.com/yavin-dev/framework/blob/269b7861e6e19fffd435893d138c5c9e8e175518/packages/reports/addon/routes/reports/report/view.ts#L38
https://github.com/yavin-dev/framework/blob/94fd8e01ac41f200626f1f90962b4ee4d0368071/packages/dashboards/addon/services/dashboard-data.ts#L49

packages/reports/addon/components/report-builder.hbs
```diff
@@ -43,9 +43,20 @@
...
+                <DenaliIcon @icon="warning" @size="small" id="report-builder__row-limit-warning" class="report-builder__row-limit-warning--icon"/>
+                <EmberTooltip targetId="report-builder__row-limit-warning" @popperContainer="body">
+                  You will likely have performance issues above 10,000 rows, use at your own risk.
+                </EmberTooltip>
+              {{/if}}
+              <DenaliInput placeholder="Row Limit"
+                @size="small"
+                @wrapperClass="w-20"
+                type="number"
```
jkusa, line 56:
```suggestion
                type="number"
                min=1""
                max="10000"
```

packages/reports/addon/components/report-builder.hbs
jkusa: For the time being, I think we should treat 10k the upper limit

jkusa: At least for the UI

packages/reports/addon/components/report-builder.hbs
cythrawll: did we discuss allowing them to set it above 10k? And just showing a warning?

packages/reports/addon/components/report-builder.hbs
cythrawll: Yeah seems like there is some issues we'd have to fix to support above 10k well.

packages/client/test/tests/unit/plugins/fili/adapters/facts-test.ts
```diff
@@ -1227,9 +1227,9 @@ module('Unit | Plugins | Fili | Adapters | facts', function (hooks) {
         having: 'm1-gt[0]',
         format: 'json',
         page: 1,
-        perPage: 100,
+        perPage: 22,
       },
-      'it overrides the request limit with the pagination request options'
+      'the request limit overrides the perPage option'
```
jkusa, line 1232:
```suggestion
      'request limit override adapter options if limit is less than page size'
```

packages/data/tests/unit/adapters/facts/elide-test.ts
```diff
@@ -290,14 +290,20 @@ module('Unit | Adapter | facts/elide', function (hooks) {
 
     assert.deepEqual(
       adapter['getPaginationOptions'](TestRequest, {}),
-      { first: 10000, after: 0 },
+      { first: 10_000, after: 0 },
```
jkusa, line 293: 🎉

jkusa (approved):
Small things, otherwise 👍

